PR #62, opened 2026-04-08
feat(api): satellite support — OAuth credentials + model compat

otarbes:
## Summary

Support pour les satellites fullstack (workspace-fs) qui ont besoin d'accéder aux secrets chiffrés d'appstrate.

- **`GET /api/connections/:connectionId/credentials`** — Retourne l'access_token OAuth déchiffré pour les cloud drives (Google Drive, etc.). Auto-refresh si le token est expiré. Filtrage par orgId pour empêcher l'accès cross-org.
- **`GET /api/models/:id/config` — champ `compat`** — Retourne les overrides de compatibilité Pi SDK (Mistral/Together rejettent `store`, `max_completion_tokens`). Les satellites n'ont plus à maintenir leur propre détection de provider.
- **`API_KEY_ALLOWED_SCOPES`** — Ajout de `connections:read` et `connections:connect` pour que les API keys satellites puissent lister et initier des connexions OAuth.

## Contexte

workspace-fs est un satellite fullstack qui consomme l'API appstrate via API key (`ask_*`). Il orchestre des containers Docker pour le chat LLM et gère le stockage fichiers (S3 + cloud drives). Ces endpoints sont nécessaires pour :

1. Résoudre les tokens Google Drive au runtime (chaque requête Drive a besoin d'un access_token frais)
2. Configurer correctement les containers chat avec les flags de compatibilité provider

### `connections.ts` — Endpoint credentials
- Lookup par `connectionId` dans `userProviderConnections`
- Vérifie `conn.orgId === c.get("orgId")` (sécurité cross-org)
- Force `forceRefreshCredentials()` si `expiresAt` est passé ou absent (buffer 5 min)
- Retourne `{ access_token }` uniquement
- Même pattern que `GET /api/models/:id/config`

### `models.ts` — Champ compat
- `buildModelCompat(baseUrl)` détecte les providers stricts (mistral.ai, together.ai)
- Retourne `{ supportsStore, maxTokensField, supportsDeveloperRole }` conditionnellement

### `permissions.ts` — Scopes API key
- `connections:read` — lire les credentials
- `connections:connect` — initier les flows OAuth

## Test plan

- [x] Token refresh automatique vérifié (token expiré de +24h → nouveau token valide)
- [x] Upload/download/rename/delete fichiers Google Drive via workspace-fs
- [x] Filtre org vérifié (connectionId d'une autre org → 404)
- [x] Compat retourné pour Mistral, absent pour OpenRouter
- [x] TypeScript, ESLint, Prettier, OpenAPI validation passent

### DIFF
--- a/apps/api/src/lib/permissions.ts
+++ b/apps/api/src/lib/permissions.ts
@@ -220,6 +220,9 @@ const ROLE_PERMISSIONS: Record<OrgRole, ReadonlySet<Permission>> = {
  * Session-only operations (org management, billing, personal profiles, etc.) are excluded.
  */
 export const API_KEY_ALLOWED_SCOPES: ReadonlySet<Permission> = new Set<Permission>([
+  // Connections (read credentials + connect for satellite apps)
+  "connections:read",
+  "connections:connect",
   // Agents
   "agents:read",
   "agents:write",
--- a/apps/api/src/routes/connections.ts
+++ b/apps/api/src/routes/connections.ts
@@ -325,7 +325,7 @@ export function createConnectionsRouter() {
   // Used by satellite apps (workspace-fs) to resolve cloud drive tokens.
   // Forces a token refresh if the stored token is expired.
   router.get("/:connectionId/credentials", requirePermission("connections", "read"), async (c) => {
-    const connectionId = c.req.param("connectionId");
+    const connectionId = c.req.param("connectionId")!;
 
     const [conn] = await db
       .select()
--- a/apps/api/src/routes/connections.ts
+++ b/apps/api/src/routes/connections.ts
@@ -6,7 +6,14 @@ import type { AppEnv } from "../types/index.ts";
 import { getItemId } from "./packages.ts";
 import { logger } from "../lib/logger.ts";
 import { escapeHtml } from "../lib/html.ts";
-import { ApiError, forbidden, invalidRequest, internalError, parseBody } from "../lib/errors.ts";
+import {
+  ApiError,
+  forbidden,
+  invalidRequest,
+  internalError,
+  notFound,
+  parseBody,
+} from "../lib/errors.ts";
 import { requirePermission } from "../middleware/require-permission.ts";
 import {
   listActorConnections,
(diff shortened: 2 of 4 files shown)


pierrecabriere:
## Review — Architectural concern

The PR is clean and functional, but the approach bypasses the existing end-user model that portal/ already uses. Before merging, we should decide if this is the right pattern for workspace-fs.

### The problem

workspace-fs is a **multi-user product** (office workers, each with their own Drive, their own files). But this PR gives it **org-level** access via API key:

- `GET /connections/:id/credentials` → returns the decrypted OAuth access_token for **any connection in the org**
- `GET /models/:id/config` → returns the LLM API key **in plaintext**

workspace-fs then has to decide which user is allowed to access which token. User-level authorization is pushed into the satellite instead of staying in Appstrate where it belongs.

### Comparison with portal/

portal/ does exactly what workspace should be doing:

| | End-user pattern (portal) | API key pattern (this PR) |
|---|---|---|
| **Scoping** | Per end-user | Entire org |
| **Who authorizes?** | Appstrate | The satellite |
| **Secrets exposed** | Only the authenticated user's | All users in the org |
| **Compromised satellite** | 1 user affected | All org users affected |

### What workspace-fs should do instead

Follow the same pattern as portal:

1. **Own Better Auth instance** in workspace-fs — each workspace user has an account
2. **Each workspace user = an Appstrate end-user** — 1:1 mapping
3. **OAuth connections scoped to the end-user** — not the org
4. **API calls with end-user context** (`Appstrate-User` header) — Appstrate filters by user, not by org
5. **Tokens returned only for the authenticated end-user**

### The infrastructure already exists

The end-user OAuth connection system is **fully built and production-tested**:

- `connectionProfiles` has `endUserId` FK — each end-user gets a default profile on creation
- `Appstrate-User` header on API key requests impersonates an end-user (with audit trail)
- `getActor(c)` resolves member OR end-user transparently in all connection routes
- `actorFilter()` / `actorInsert()` scope all queries to the correct actor
- **portal/ already uses this**: `initiateEndUserConnection()`, `getEndUserConnections()`, `disconnectEndUserConnection()` — all scoped per end-user via the `Appstrate-User` header

There is no gap to fill — workspace-fs just needs to follow the same pattern.

### Note on `/models/:id/config`

LLM models are org-level resources, so the API key approach makes more sense here. But exposing `apiKey` in plaintext is still risky — the sidecar (`runtime-pi/sidecar`) already exists to isolate LLM credentials. workspace-fs could proxy its LLM requests through a similar mechanism rather than fetching the raw key.

### TL;DR

workspace-fs is the same use case as portal (multi-user product consuming Appstrate), not an infrastructure service. The end-user connection system already exists, is production-tested by portal, and handles OAuth scoping, token refresh, and actor isolation out of the box. workspace-fs should use `Appstrate-User` header + existing connection endpoints rather than opening new org-level endpoints that expose all users' secrets.

PR #68, opened 2026-04-09
fix: use native Mistral provider instead of OpenAI-compatible

otarbes:
## Summary

- Mistral model presets now use `mistral-conversations` API instead of `openai-completions`
- Fixes agent crash after tool calls with Codestral/Mistral models
- Root cause: Mistral returns `content: null` in OpenAI-compatible format after tool calls, and the Pi SDK crashes on `message.content.filter(...)` (`pi-agent-core`, `agent-loop.js:111`)
- The Pi SDK's native Mistral provider (`mistral-conversations`) handles this correctly

## Details

The Pi SDK maintainer has explicitly declined to add a null guard (issues #1367, #1899 on `badlogic/pi-mono`), considering it a provider-side issue. Using the native provider is the correct fix.

**Note**: Existing Mistral models already configured in the DB will keep `openai-completions`. They need to be recreated via presets or have their `api` field updated manually.

## Test plan

- [ ] Create a new Mistral model from presets (should use `mistral-conversations`)
- [ ] Start a chat session with Codestral
- [ ] Trigger a tool call (e.g. ask agent to use a tool)
- [ ] Verify agent responds correctly after tool execution (no crash)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
-### Fixed
-
-- Mistral/Codestral models now use `mistral-conversations` API instead of `openai-completions`. Fixes agent crash (`message.content` undefined) after tool calls — the Pi SDK's native Mistral provider handles tool call responses correctly. (`apps/web/src/lib/model-presets.ts`)
-
 ### Added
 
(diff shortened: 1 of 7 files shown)


pierrecabriere:
@otarbes il faudrait revert ton premier commit (6ed564e1f1ce701ee0ce7b40c2778f18a681587e) il ne devrait pas faire partie de cette pr

PR #73, opened 2026-04-09
refactor: extract webhooks into a module + formalize the module system

pierrecabriere:
## Summary

Formalizes the `AppstrateModule` contract and extracts webhooks into a built-in module. Scheduling and provider-management were also extracted during iteration but moved back into core once it became clear the coupling with the run pipeline (FK, filtering, enrichment, realtime, model resolution on the hot path) made module isolation cost more than it delivered. Webhooks keeps a genuinely clean boundary — a single `onRunStatusChange` listener, no reach-backs, isolated BullMQ delivery worker — so it stays a module.

### Module system

- **Auto-discovery**: dropping a directory under `apps/api/src/modules/<name>/` with a default export is enough — no registration table, no hardcoded list. `verify-openapi` does its own filesystem scan to enumerate built-ins.
- **Module contract** (`@appstrate/core/module`, published on npm so external modules don't depend on the API package): manifest, `init(ctx)`, `createRouter()`, `publicPaths`, `features`, `hooks`, `events`, `openApiPaths`, `openApiComponentSchemas`, `openApiSchemas`, `emailOverrides`, `shutdown()`.
- **Hooks vs events**: hooks (`beforeRun`, `afterRun`, `beforeSignup`) are first-match-wins; events (`onRunStatusChange`, `onOrgCreate`, `onOrgDelete`) are broadcast-to-all. The platform dispatches by name, never by module id.
- **`afterRun` hook**: symmetric with `beforeRun`, returns a metadata patch persisted on the run record (used by Cloud for credit accounting). Fires on every terminal status.
- **Module-owned database schemas**: each module owns its tables (`webhooks`, `webhook_deliveries` for the webhooks module), with its own Drizzle migration directory, journal, and tracking table (`__drizzle_migrations_<moduleId>`). Applied via `ctx.applyMigrations()` in `init()`.
- **Init context**: `databaseUrl`, `redisUrl`, `appUrl`, `isEmbeddedDb`, `applyMigrations`, `getSendMail`, `getOrgAdminEmails`.
- **Topological sort**: modules loaded in dependency order via `manifest.dependencies`. Circular dependencies throw at boot.
- **Permissions**: RBAC stays in core (`apps/api/src/lib/permissions.ts`) — module manifests do not declare permissions. Module routes use the same typed `requirePermission` helper as core.
- **Disabling a module**: remove it from `APPSTRATE_MODULES` and nothing from that module is imported, initialized, or mounted. Core is unaffected — it knows nothing about the module's existence. Default is empty (OSS mode).

### Core boot changes

- **Auto-migrations at boot**: core migrations run automatically on both PostgreSQL and PGlite — no manual `db:migrate` step. Wrapped in a session-scoped `pg_advisory_lock` held on a reserved postgres-js connection (see pool-safety fix below) so concurrent replicas don't race on `__drizzle_migrations`.
- **NOTICE suppression** during migrations (no more \"already exists, skipping\" spam).
- **`openapi:diff` in CI**: `bun run check` now runs the breaking-change detector against the assembled spec (including module contributions), fixing a bug where it was reading the stale core-only spec.
- **Run enrichment fields** documented in the OpenAPI `Run` schema.

### Pool-safe advisory locks

The old pattern ran \`pg_advisory_lock\`, \`migrate()\`, and \`pg_advisory_unlock\` as three separate calls against the shared Drizzle pool, so they could each land on a different backend connection. The unlock would silently target the wrong session, leaving the original lock held forever and blocking any subsequent concurrent replica.

Fix: new `reservePgConnection()` helper in `@appstrate/db/client` reserves one postgres-js connection for the lock/unlock pair; the migrator keeps running on the shared `db` since holding the lock on any one backend is enough to block concurrent callers. Applied to both `applyCoreMigrations` (boot) and `applyPostgresMigrations` (module migrations helper).

### Frontend

- Module UI is feature-gated: sidebar entries, routes, and tabs for module-owned features default to `false` and are enabled only when the corresponding module is loaded. Disabled modules have zero UI footprint.

### Disabling modules

```bash
APPSTRATE_MODULES=                      # bare minimum, webhooks module off (OSS default)
APPSTRATE_MODULES=webhooks              # webhooks module on
APPSTRATE_MODULES=@appstrate/cloud      # cloud module (external npm)
APPSTRATE_MODULES=webhooks,@appstrate/cloud
```

## Test plan
- [x] `bun run check` passes (typecheck + eslint + prettier + verify-openapi + openapi:diff + system-packages)
- [x] `bun test` — 1902 pass, 0 fail
- [x] Cloud: `bun run check` + `bun run test:unit` — green
- [x] Integration test for `applyModuleMigrations` covers idempotence and cross-module isolation (both previously unreachable because of the pool-leak bug)
- [x] Module-loader unit test covers empty-set behaviour (no module routes, features, public paths, or hooks contributed)
- [ ] Manual: verify `APPSTRATE_MODULES=\"\"` boots without the `/api/webhooks` routes


pierrecabriere:
### Code review

Found 2 issues:

1. `APPSTRATE_MODULES` default changed from `""` to `"scheduling,webhooks,provider-management"`, contradicting CLAUDE.md which states "Disabling a module = zero footprint... Default is empty (OSS mode)". The CLAUDE.md text itself was not updated in this PR, so the documented OSS default and the runtime default now disagree.

https://github.com/appstrate/appstrate/blob/83bbd9bdec495da376c085cb153f8ab9b0aeef91/packages/env/src/index.ts#L37-L41

2. `webhooksModule.init()` does not await `initWebhookWorker()`, while `schedulingModule.init()` correctly awaits `initScheduleWorker()`. `initWebhookWorker` is `async` and can throw during BullMQ/queue setup. Per the module contract in `@appstrate/core/module`, "any error [in init] is treated as a fatal init failure" — leaving this unawaited turns a fatal boot error into a silent unhandled promise rejection.

https://github.com/appstrate/appstrate/blob/83bbd9bdec495da376c085cb153f8ab9b0aeef91/apps/api/src/modules/webhooks/index.ts#L26-L30

Compare with the scheduling module:

https://github.com/appstrate/appstrate/blob/83bbd9bdec495da376c085cb153f8ab9b0aeef91/apps/api/src/modules/scheduling/index.ts#L22-L26

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

pierrecabriere:
### Code review

Found 1 issue:

1. Drizzle-kit snapshot state is corrupted. `packages/db/drizzle/0004_provider_management_into_core.sql` and its journal entry exist, but `meta/0004_snapshot.json` is missing. In addition, the last snapshot on disk (`meta/0003_snapshot.json`) still contains the `webhooks` / `webhook_deliveries` tables even though the webhooks Drizzle schema was moved out of `packages/db/src/schema/` into the module. The next time anyone runs `bun run db:generate`, drizzle-kit will diff the current core schema (no webhooks, has `org_provider_keys`/`org_models`) against `0003_snapshot.json` and produce a core migration containing `DROP TABLE "webhooks" CASCADE` + `DROP TABLE "webhook_deliveries" CASCADE` (plus a duplicate of the 0004 provider-management changes). I reproduced this locally — running `bun run db:generate` on this branch generates exactly that destructive migration. If a developer doesn't notice and commits it, the next deploy wipes all webhook data. Per `appstrate/CLAUDE.md`: _"DB migration (core): Edit \`packages/db/src/schema.ts\` → \`bun run db:generate\`"_, so this footgun is on the documented path.

https://github.com/appstrate/appstrate/blob/8ef2c11d6af57bcad22026be998881ca14870478/packages/db/drizzle/meta/_journal.json#L32-L41

Two related concerns that didn't meet the confidence bar to block, but worth addressing before merge to main:

- **`run.completed` → `run.success` rename has no data migration.** `origin/main` currently stores `"run.completed"` in `webhooks.events` (see `apps/api/src/services/webhooks.ts` on main), and the new enum in [`apps/api/src/modules/webhooks/service.ts#L26-L34`](https://github.com/appstrate/appstrate/blob/8ef2c11d6af57bcad22026be998881ca14870478/apps/api/src/modules/webhooks/service.ts#L26-L34) only accepts `run.success`. The dispatch filter at [`service.ts#L458-L462`](https://github.com/appstrate/appstrate/blob/8ef2c11d6af57bcad22026be998881ca14870478/apps/api/src/modules/webhooks/service.ts#L458-L462) uses `wh.events?.includes(eventType)` with `eventType = "run.success"`, so any existing subscription with `["run.completed", ...]` will silently stop firing when this lands on main. Consider an `UPDATE webhooks SET events = array_replace(events, 'run.completed', 'run.success')` in the module's initial migration (or a core migration prior to the module taking over the table).

- **`pg_advisory_lock` used across unpinned pool connections.** In [`apps/api/src/lib/modules/migrate.ts#L72-L89`](https://github.com/appstrate/appstrate/blob/8ef2c11d6af57bcad22026be998881ca14870478/apps/api/src/lib/modules/migrate.ts#L72-L89), the lock/migrate/unlock sequence is three separate `db.execute()` calls on a drizzle+postgres.js instance configured with `max: 20` pooling. `pg_advisory_lock` is session-level — the lock is bound to the specific connection that ran the `SELECT`. There is no `sql.reserve()` or transaction wrapper pinning these three statements to one connection, so unlock may run on a different pooled connection than the lock (silently returns `false`), leaving the original connection holding the lock until it's recycled. For a single-replica boot this usually "just works" by luck of LIFO reuse, but for the multi-replica scenario this code is explicitly designed to protect against (per the comment at line 71), it's unsound. Pin with `await db.transaction(...)` or `sql.reserve()`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #258, opened 2026-04-25
feat(cli): stamp install source at build time (#249 phase 1)

pierrecabriere:
## Summary

Phase 1 of #249. Stamps every CLI binary with the distribution channel that produced it (`"curl"` for the signed release binary, `"bun"` for the npm package, `"unknown"` for source/dev builds).

## Wiring

- `apps/cli/src/lib/install-source.ts` — exports `INSTALL_SOURCE`, `INSTALL_SOURCES`, `upgradeHint(source)`. Resolves via a `typeof` guard on the build-time identifier `__APPSTRATE_INSTALL_SOURCE__`, falling back to `"unknown"` when the bundle was not stamped.
- `apps/cli/scripts/build.ts` — wraps `bun build` with the `--define` so the npm `prepack` chain stays clean. Reads `APPSTRATE_INSTALL_SOURCE` env var with `"unknown"` default.
- `apps/cli/package.json` — `build` now calls `bun scripts/build.ts`.
- `.github/workflows/release.yml` — adds `--define '__APPSTRATE_INSTALL_SOURCE__="curl"'` to the compile step.
- `.github/workflows/publish-cli.yml` — sets `APPSTRATE_INSTALL_SOURCE=bun` before `npm pack`.

## Why now

Unblocks phases 2 (`self-update`) and 3 (`doctor`) of #249 — both need to know which channel produced the running binary. No runtime behavior change in this PR; the constant is exported but no command consumes it yet.

## Test plan
- [x] `bun test apps/cli/test/install-source.test.ts` (7 tests, end-to-end build+run probes)
- [x] `bun test` in `apps/cli` — 717 pass
- [x] `bun run check` (turbo + verify-openapi) — all green
- [x] Manual smoke: `APPSTRATE_INSTALL_SOURCE=bun bun scripts/build.ts` → `dist/cli.js` reports `bun`

### DIFF
--- a/.github/workflows/publish-cli.yml
+++ b/.github/workflows/publish-cli.yml
@@ -96,6 +96,11 @@ jobs:
         # inlined assets (the #175 regression), we fail BEFORE polluting
         # the npm registry — unlike a post-publish smoke test, this
         # failure mode is recoverable without `npm deprecate`.
+        #
+        # APPSTRATE_INSTALL_SOURCE=bun stamps the bundled CLI as the
+        # npm-channel artefact (see `src/lib/install-source.ts`, issue #249).
+        env:
+          APPSTRATE_INSTALL_SOURCE: bun
         run: |
           npm pack
           # `npm pack` names the tarball `<package>-<version>.tgz` deterministically;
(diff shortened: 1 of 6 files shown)


pierrecabriere:
Superseded by #270 (consolidated single PR for issue #249).

Issue #149, opened 2026-04-17
Binary response body corrupted by UTF-8 encoding in /proxy


mike1572:
### Description

Binary file downloads through the sidecar /proxy endpoint are corrupted. Bytes in the range 0x80-0xFF that don't form valid UTF-8 sequences are replaced with U+FFFD (replacement character, EF BF BD), making files like PDFs, images, and archives unreadable.

In runtime-pi/sidecar/app.ts (~line 437), the response body is decoded as text:

const responseText = await targetRes.text();
// ...
return c.body(text, targetRes.status, responseHeaders);
targetRes.text() decodes the upstream response as UTF-8, which irreversibly replaces non-UTF-8 bytes with U+FFFD.

PR #50 fixed the same issue on the request body side (uploads) by switching from c.req.text() to c.req.arrayBuffer(). The response body path was not updated.

### Steps to Reproduce

```typescript
Create an agent with @appstrate/google-drive provider
Download a PDF via the sidecar proxy:

curl -s "$SIDECAR_URL/proxy" \
  -H "X-Provider: @appstrate/google-drive" \
  -H "X-Target: https://www.googleapis.com/drive/v3/files/FILE_ID?alt=media" \
  -H "Authorization: Bearer {{access_token}}" \
  -o /workspace/file.pdf
The downloaded file is larger than the original (e.g., 74 KB → 121 KB) and corrupted
All PDF parsers (pdf-parse, pdfminer, pypdf, pymupdf) fail because zlib-compressed streams contain replacement characters
```

### Expected Behavior

Downloading a 74 KB PDF via the sidecar proxy returns an identical 74 KB file
The file is readable by standard PDF parsers (pdf-parse, pdfminer, pypdf, pymupdf)

### Actual Behavior

The downloaded file is 121 KB (bytes expanded by UTF-8 encoding)
Bytes >0x7F are replaced with U+FFFD (EF BF BD)
All PDF parsers fail — zlib-compressed streams are irreversibly corrupted
No header or config option exists to bypass this behavior

### Operating System

macOS 26

### Bun version

1.5x

### Docker version

28.5x

pierrecabriere:
Duplicate #151 (@otarbes) has a ready-to-apply patch and two regression tests (`runtime-pi/sidecar/test/app.test.ts`) that fail before / pass after. Closed that one; consolidating here.

Issue #151, opened 2026-04-17
Sidecar /proxy corrupts binary response bodies (PDF, images, XLSX)


otarbes:
## Summary

The sidecar `/proxy` endpoint decodes upstream responses via `targetRes.text()`
(`runtime-pi/sidecar/app.ts:447`), which silently replaces invalid UTF-8 bytes
with `U+FFFD` (encoded as `0xEF 0xBF 0xBD` — 3 bytes per invalid byte).

Any agent downloading a binary file (PDF from Drive, image from Slack, XLSX
from SharePoint, …) through the sidecar receives corrupted bytes. The file
cannot be re-uploaded or parsed downstream.

Symmetric counterpart to the bug fixed on the **request** side in #50 /
`6be5ce81` — same root cause (`text()` on arbitrary bytes), same corruption
pattern, opposite direction.

## Reproduction

5 lines demonstrate the issue in isolation:

```ts
const pdfLike = new Uint8Array([0x25, 0x50, 0x44, 0x46, 0xff, 0xfe, 0x80]);
const asText = new TextDecoder().decode(pdfLike);
const reEncoded = new TextEncoder().encode(asText);
console.log(reEncoded);
// → [0x25, 0x50, 0x44, 0x46, 0xEF, 0xBF, 0xBD, 0xEF, 0xBF, 0xBD, 0xEF, 0xBF, 0xBD]
// 7 bytes in → 13 bytes out. Original bytes lost.
```

With a 60 KB binary payload through `/proxy`, the current code returns
**~100 KB of corrupted bytes** even before truncation kicks in (truncation
is applied on char count, not byte count).

## Proposed fix

Read upstream as `ArrayBuffer` and truncate by byte length — mirrors the
`/llm/*` proxy in the same file (`app.ts:233`), which already streams bytes
through without decoding.

```diff
- const responseText = await targetRes.text();
- const truncated = responseText.length > maxSize;
- const text = truncated ? responseText.slice(0, maxSize) : responseText;
+ const responseBytes = await targetRes.arrayBuffer();
+ const truncated = responseBytes.byteLength > maxSize;
+ const body = truncated ? responseBytes.slice(0, maxSize) : responseBytes;
  // ...
- return c.body(text, targetRes.status, responseHeaders);
+ return c.body(body, targetRes.status, responseHeaders);
```

## Tests

Added to `runtime-pi/sidecar/test/app.test.ts`, inside the existing
`describe("ALL /proxy — binary body integrity")` block:

[...]

PR #260, opened 2026-04-25
fix(afps-runtime): restore binary upload/download parity in provider tools

pierrecabriere:
## Summary

The migration from `curl $SIDECAR_URL/proxy` to typed `<provider>_call` tools (PR #227) re-introduced the regression that closed issues #149 and #151 had fixed in the sidecar: bytes were decoded as UTF-8 text in `serializeFetchResponse`, corrupting binary payloads (PDFs, images, ZIPs from Google Drive etc.). On the request side, `{ fromFile }` body references and `responseMode.toFile` were declared in the JSON schema but ignored. The sidecar HTTP layer is fine — the fix lives entirely in the client-side resolvers.

This is **PR-1** of a 4-PR series:
- **PR-1 (this PR)**: byte-safe `serializeFetchResponse`, workspace-rooted `{ fromFile }` and `responseMode.toFile`, discriminated-union response body.
- **PR-2** (next): sidecar-side limits / streaming readiness.
- **PR-3**: system prompt + web run-viewer updates for the new body shape (binary preview).
- **PR-4**: streaming.

## What changed

- Replaced `body: { inline; inlineEncoding } | { file }` with a discriminated union `{ kind: \"text\" | \"inline\" | \"file\"; ... }`. `inline` now always carries base64 + mimeType + size; `file` carries a workspace-relative path + sha256 + size.
- `serializeFetchResponse` reads `arrayBuffer()` (NEVER `text()`) and decides the variant via a strict text-mime whitelist (`text/*`, `application/json`, `application/xml`, `application/javascript`, `+json` / `+xml` suffixes, `; charset=...`). Anything else round-trips as `inline` base64 or spills to `responses/<toolCallId>.bin` past the inline cap. Default cap 64 KB, hard cap 1 MB — both mirror the sidecar.
- `responseMode.toFile` is now honoured (writes the requested workspace path and returns a `file` body).
- `resolveBodyStream` accepts a `workspace` opt and resolves `{ fromFile }` via a new `resolveSafePath` helper that refuses absolute paths, parent-traversal, and symlinks whose realpath escapes the workspace. Pre-checks size against `MAX_REQUEST_BODY_SIZE` (5 MB) so the runtime surfaces a typed `RESOLVER_BODY_TOO_LARGE` error instead of letting the sidecar return 413.
- `ProviderCallFn` signature changes from `(req) => …` to `(req, ctx) => …` where `ctx` carries `{ workspace, toolCallId, signal }`. Workspace flows from `ToolContext` through the resolver rather than constructor injection, so `runtime-pi/entrypoint.ts` and `delete process.env.SIDECAR_URL` are untouched.
- Sidecar + RemoteAppstrate resolvers forward `X-Max-Response-Size` (capped at 1 MB) when the agent asks for a larger inline payload OR routes the response to a file — the sidecar's default 50 KB cap no longer truncates before the resolver spills.
- New error codes: `RESOLVER_BODY_TOO_LARGE`, `RESOLVER_PATH_OUTSIDE_WORKSPACE`, `RESOLVER_PATH_INVALID`.

## Out of scope (handled in subsequent PRs)

- Sidecar HTTP layer is correct and unchanged (issues #149 / #151 already fixed it server-side).
- `runtime-pi/entrypoint.ts` Phase C wiring — workspace flows via ctx, no construction-time change needed.
- System prompt (`platform-prompt.ts`) — PR-3.
- Web run viewer — PR-3 (only the type-level union is updated here; no UX changes).
- Streaming — PR-4.

## Test plan

- [x] `bun run check` clean (typecheck + verify-openapi + prettier across all packages)
- [x] `bun test packages/afps-runtime` — 435 pass (17 new in `test/resolvers/binary-roundtrip.test.ts`)
- [x] `bun test packages/runner-pi` — 86 pass
- [x] `bun test runtime-pi` — 209 pass
- [x] `bun test apps/cli` — 710 pass
- [x] PNG / PDF round-trip preserves sha256 on upload AND download
- [x] Octet-stream with ASCII body comes back as `inline` base64 (strict whitelist)
- [x] Path traversal / absolute / symlink-escape rejected with typed error
- [x] >5 MB `{ fromFile }` rejected with `RESOLVER_BODY_TOO_LARGE`
- [x] `responseMode.toFile` writes to the workspace; `>maxInlineBytes` auto-spills


pierrecabriere:
Superseded by #272 — the 4 staged PRs are consolidated into a single review unit.

PR #265, opened 2026-04-25
feat(installer): dual-install pre-check in bootstrap.sh (#249 phase 4)

pierrecabriere:
## Summary

Phase 4 of #249. Before downloading anything, `scripts/bootstrap.sh` now scans `\$PATH` for an existing `appstrate` binary at a different location than `\$DEST`. The flow is:

| State | Action |
|---|---|
| No existing `appstrate` on PATH | Continue (current behavior). |
| Existing at same resolved path as `\$DEST` (re-install over self) | Continue silently. |
| Existing at a different path, interactive TTY | Prompt `[y/N]`, abort by default. |
| Existing at a different path, CI or non-TTY | Abort with hint to set `APPSTRATE_FORCE_DUAL=1`. |
| `APPSTRATE_FORCE_DUAL=1` | Bypass + emit a banner. |

## Implementation notes

- The pre-check runs BEFORE `curl` for the binary, so a user who chose the wrong install method doesn't burn bandwidth + minisign cycles before aborting.
- `resolve_path` canonicalises **directory** symlinks but tolerates a non-existent leaf (the leaf is `\$DEST`, which doesn't exist on a fresh install). Handles macOS Homebrew `/usr/local/bin → /opt/homebrew/bin` correctly.
- BSD `realpath` (macOS default) and GNU `readlink -f` both fail on missing paths — splitting `dirname` + `basename` and only resolving the dir avoids this without a portable shell hack like `realpath -m` (Linux-only).
- The interactive prompt reads from `/dev/tty` when stdin is the curl pipe — same trick rustup uses.

## Tests

Three new scenarios in `.github/workflows/test-install.yml` (job `bootstrap-verify-dry-run`):
- 8a.1: CI=true + dual-install detected → must abort with the dual-install banner + CI guard.
- 8a.2: CI=true + `APPSTRATE_FORCE_DUAL=1` → must succeed and emit the override banner.
- 8a.3: existing == DEST (re-install over self) → must succeed without nagging.

Validated locally with a stripped harness against all three scenarios + the BSD-realpath edge case on macOS.

## Test plan
- [x] `shellcheck scripts/bootstrap.sh` — clean
- [x] `shfmt -d -i 2 -ci scripts/bootstrap.sh` — clean
- [x] Local harness: all three branches behave correctly
- [x] `bun run check` — repo-wide quality gate passes
- [x] CI `Test Installer` workflow exercises the new scenarios on every push

### DIFF
--- a/.github/workflows/test-install.yml
+++ b/.github/workflows/test-install.yml
@@ -463,6 +463,82 @@ jobs:
           echo "::endgroup::"
           echo "✓ empty-manifest path OK"
 
+          # ── Scenario 8a: dual-install pre-check (issue #249) ─────────
+          # When `appstrate` already exists on PATH at a different path
+          # than $DEST, bootstrap MUST refuse non-interactively (CI=true).
+          # APPSTRATE_FORCE_DUAL=1 must override.
+          echo "::group::dual-install pre-check"
+          mkdir -p "$GITHUB_WORKSPACE/bin-existing"
+          # Plant a fake existing appstrate in a directory that will appear
+          # on PATH BEFORE the install destination, so command -v finds it.
+          cat > "$GITHUB_WORKSPACE/bin-existing/appstrate" <<'EXISTING'
+          #!/usr/bin/env bash
+          echo "EXISTING_APPSTRATE 0.9.0"
+          exit 0
+          EXISTING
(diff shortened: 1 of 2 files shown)


pierrecabriere:
Superseded by #270 (consolidated single PR for issue #249).

PR #287, opened 2026-04-26
feat(sidecar): env-configurable body caps + structured 413 errors

pierrecabriere:
## Summary

The sidecar enforces two body-size caps that gate every `provider_call`:

- **JSON-RPC envelope**: `MAX_MCP_REQUEST_BODY_SIZE = 256 KB` (HTTP-layer guard)
- **Per-request body** (post base64 decode): `MAX_REQUEST_BODY_SIZE = 5 MB`

Both are hardcoded. Worse, the envelope cap is by far the tighter of the two — the user-facing limit on `provider_call` with `body.fromBytes` is **~190 KB raw**, not the 5 MB the inner cap suggests, because base64 inflation (~1.37×) plus JSON-RPC overhead must fit the envelope.

This PR makes both caps env-configurable and raises the defaults so out-of-the-box uploads up to 10 MB work without operator action. It also replaces prose 413 / `isError` messages with structured payloads so agents can act on the failure programmatically.

### Caps
- `SIDECAR_MAX_REQUEST_BODY_BYTES` (default **10 MB**, was 5 MB) — per-request body after decode.
- `SIDECAR_MAX_MCP_ENVELOPE_BYTES` (default **16 MB**, was 256 KB) — JSON-RPC envelope at HTTP edge. Sized to fit base64-inflated 10 MB body plus JSON-RPC overhead.
- Loud-failing env parsing (`readPositiveByteEnv` in `helpers.ts`): rejects malformed values, non-positive values, and anything above the 100 MB absolute ceiling. Sidecar boot fails fast on misconfiguration.
- `MAX_REQUEST_BODY_SIZE` decoupled from `MAX_SUBSTITUTE_BODY_SIZE` (which keeps its 5 MB value, separate concern).

### Structured 413 / oversize errors
Both the HTTP-edge 413 and the inner `isError: true` tool result now carry structured fields:
```ts
{
  reason: "PAYLOAD_TOO_LARGE",
  scope: "mcp_envelope" | "request_body",
  limit: number,
  actual: number,
  envVar: "SIDECAR_MAX_MCP_ENVELOPE_BYTES" | "SIDECAR_MAX_REQUEST_BODY_BYTES",
  hint: "..."
}
```
Agents can branch on these fields instead of regexing the prose message.

### Runtime mirror
`packages/afps-runtime/src/resolvers/provider-tool.ts` mirrors `MAX_REQUEST_BODY_SIZE` so over-sized uploads still fail fast client-side with a typed error before any bytes hit the wire. The runtime reads the same env var and falls back silently on malformed input (it runs in many contexts, not just the sidecar container — throwing would be unhelpful).

### Tests
- Existing cap tests now read caps from helpers (instead of hard-coding 256 KB) and additionally assert the structured `error.data` shape.
- New `readPositiveByteEnv` unit tests covering: unset, empty, valid integer, non-numeric, zero/negative, non-integer, above ceiling, custom ceiling.
- New `provider_call` `body.fromBytes` over-cap test asserting the structured `structuredContent.error` shape.

## Memory profile

Per-call peak when uploading a 10 MB binary:
- Envelope buffered for SDK `req.json()`: ~13.4 MB (base64-inflated body + JSON-RPC overhead).
- Decoded request body materialised before `executeProviderCall`: 10 MB.
- Total per concurrent call: ~23 MB.

The MCP transport is single-request stateless per-`/mcp` invocation, so concurrent calls compound only as fast as the agent's tool-call loop allows. Acceptable for a sidecar container (typically 256–512 MB memory limit).

## Out of scope

- Chunked / resumable uploads for files larger than the cap. The structured error's `hint` field points operators at the right knobs and explains why both caps must be raised together. Larger uploads need a different mechanism (`provider_upload` primitive with protocol adapters), tracked separately.
- Streaming the JSON-RPC envelope decode (peak memory could drop further but requires a non-trivial transport refactor).

## Test plan

- [x] `bun run check` — typecheck + lint + format + verify:openapi + breaking-change detection — all pass.
- [x] `cd runtime-pi && bun test` — 285 tests pass.
- [x] `cd packages/afps-runtime && bun test` — 425 tests pass.
- [x] `cd runtime-pi/sidecar && bun test test/helpers.test.ts test/mcp.test.ts` — 103 tests pass, including the new env-parsing and structured-error coverage.

### DIFF
--- a/apps/api/src/services/orchestrator/docker-orchestrator.ts
+++ b/apps/api/src/services/orchestrator/docker-orchestrator.ts
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { getEnv } from "@appstrate/env";
+import { pickOperatorSidecarEnv } from "@appstrate/runner-pi";
 import type { ContainerOrchestrator } from "./interface.ts";
 import type {
   WorkloadHandle,
@@ -116,7 +117,11 @@ export class DockerOrchestrator implements ContainerOrchestrator {
     }
 
     // 2. Fallback: fresh creation (~500-1500ms)
-    const sidecarEnv: Record<string, string> = { PORT: "8080" };
+    // Operator-tunable sidecar caps are read from the API host's
+    // process.env and forwarded into the spawned container so
+    // overrides apply to fresh sidecars (pooled sidecars pick up the
+    // value at pool creation time — see sidecar-pool.ts).
+    const sidecarEnv: Record<string, string> = { PORT: "8080", ...pickOperatorSidecarEnv() };
     if (resolvedConfig.runToken) {
       sidecarEnv.RUN_TOKEN = resolvedConfig.runToken;
       sidecarEnv.PLATFORM_API_URL = resolvedConfig.platformApiUrl;
--- a/apps/api/src/services/sidecar-pool.ts
+++ b/apps/api/src/services/sidecar-pool.ts
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { getEnv } from "@appstrate/env";
+import { pickOperatorSidecarEnv } from "@appstrate/runner-pi";
 import { logger } from "../lib/logger.ts";
 import {
   createContainer,
@@ -198,9 +199,13 @@ export async function startSidecarAndHealthCheck(containerId: string): Promise<n
 /** Create a single pooled sidecar container. */
 async function createPooledSidecar(): Promise<PooledSidecar> {
   const configSecret = crypto.randomUUID();
+  // Operator-tunable sidecar caps are forwarded from the API host so
+  // pooled sidecars enforce the same limits as freshly-spawned ones.
+  // The values are frozen at module load inside the sidecar, so an env
+  // change on the API host requires recycling the pool to take effect.
   const containerId = await createContainer(
     crypto.randomUUID().slice(0, 8),
-    { PORT: "8080", CONFIG_SECRET: configSecret },
+    { PORT: "8080", CONFIG_SECRET: configSecret, ...pickOperatorSidecarEnv() },
     {
       image: getSidecarImage(),
       adapterName: "sidecar-pool",
--- a/packages/afps-runtime/src/resolvers/provider-tool.ts
+++ b/packages/afps-runtime/src/resolvers/provider-tool.ts
@@ -31,11 +31,25 @@ export const defaultInlineLimit = 256 * 1024;
 export const ABSOLUTE_MAX_RESPONSE_SIZE = 1_000_000;
 
 /**
- * Hard upper bound on `{ fromFile }` request bodies. Mirrors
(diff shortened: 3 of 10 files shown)


pierrecabriere:
## Production-readiness review — fix-up commit pushed

Re-read the diff after opening the PR and found three gaps that would have made the env override silently inert in production:

### 1. Env propagation gap (the big one)

The Docker orchestrator builds the spawned sidecar's env from an explicit whitelist (`PORT`, `RUN_TOKEN`, `PLATFORM_API_URL`, `PROXY_URL`, `PI_*`). New env vars don't flow through unless added there. Same problem in `sidecar-pool.ts` for pooled sidecars and in `buildRuntimePiEnv` for the agent container's runtime mirror.

In docker mode (the production deployment) my "env-configurable" was effectively dead code — defaults always applied regardless of operator override.

**Fix:** added `pickOperatorSidecarEnv()` + `SIDECAR_OPERATOR_ENV_KEYS` to `@appstrate/runner-pi` as the single source of truth. Wired into all three orchestration sites. Empty values are dropped (an empty `SIDECAR_MAX_REQUEST_BODY_BYTES` would crash sidecar boot via `readPositiveByteEnv`).

The process orchestrator was already fine — it inherits `process.env` wholesale via `cleanProcessEnv()`.

### 2. Sub-1 MB cap rendered as "0 MB"

`Math.floor(MAX_REQUEST_BODY_SIZE / (1024 * 1024))` — if an operator sets the cap below 1 MB, the Zod description sent to the LLM said "uploads up to 0 MB". Edge case but visible footgun.

**Fix:** replaced with a `formatByteCap` helper that picks bytes/KB/MB based on which unit divides cleanly. 10 MB → "10 MB"; 512 KB → "512 KB"; 1500 → "1500 bytes".

### 3. No coverage for the env-forwarding plumbing

**Fix:** added tests for `buildRuntimePiEnv` forwarding the request-body cap (and explicitly NOT forwarding the envelope cap, which is sidecar-internal), `pickOperatorSidecarEnv` filtering with the `keys` argument, and empty-value dropping.

### Caveats now documented

- **Pooled sidecars freeze the cap at module load.** An operator env change on the API host requires recycling the pool to take effect on existing pooled containers (fresh containers pick up the new value immediately). Documented in the `sidecar-pool.ts` comment.
- **The runtime mirror only receives `SIDECAR_MAX_REQUEST_BODY_BYTES`** (not the envelope cap) — the agent runtime never builds JSON-RPC envelopes itself, so forwarding the envelope cap would be misleading. Documented in `buildRuntimePiEnv`.

### Validation

- `bun run check` — typecheck + lint + format + verify:openapi + detect:breaking + system-packages — passes.
- runner-pi: 90/90 (added 13 new tests for forwarding + helper), runtime-pi: 285/285, afps-runtime: 425/425.
- apps/api unit: 670/691 — the 21 failures are baseline (Redis-dependent rate-limit + idempotency middleware), unchanged from `origin/main`.

PR #322, opened 2026-04-27
feat(llm-proxy): wire mistral-conversations through /api/llm-proxy/*

pierrecabriere:
## Summary

- Adds Mistral as the third protocol family the LLM proxy routes (alongside `openai-completions` and `anthropic-messages`).
- Unblocks `appstrate run` against orgs whose default model preset is Mistral — currently fails fast with `Preset uses protocol "mistral-conversations", which /api/llm-proxy/* does not route yet`.
- New `mistralConversationsAdapter` in `apps/api/src/services/llm-proxy/mistral.ts`, registered as one route entry in `routes/llm-proxy.ts`. CLI's `PROXY_SUPPORTED_APIS` extended + `buildProxyBaseUrl()` mapping. OpenAPI path documented.

## Why this is a small change

`mistral-conversations` is a misnomer inherited from pi-ai's registry. Under the hood pi-ai dispatches Mistral models through the official `@mistralai/mistralai` SDK's `chat.stream(...)`, which targets `POST /v1/chat/completions` — NOT the Beta `/v1/conversations` agentic API. Confirmed by:

- `node_modules/@mistralai/mistralai/funcs/chatStream.js:67` — `pathToFunc("/v1/chat/completions#stream")`
- pi-ai's `dist/providers/mistral.js` calling `mistral.chat.stream(payload, …)` with body `{ model, messages, temperature, maxTokens, tools, stream, … }`
- pi-ai's `dist/models.generated.js` — every Mistral model entry has `baseUrl: "https://api.mistral.ai"`

The wire format is OpenAI-compatible: snake_case body, `Authorization: Bearer`, identical `{prompt_tokens, completion_tokens, total_tokens}` usage shape (`@mistralai/mistralai/models/components/usageinfo.js`). The adapter is a thin clone of `openai.ts` with `HEADERS_TO_FORWARD = new Set()` (Mistral has no `openai-organization` / `openai-beta` equivalent — the SDK's `x-affinity` sticky-session header is intentionally dropped).

## Files changed

| File | Change |
| --- | --- |
| `apps/api/src/services/llm-proxy/mistral.ts` | New adapter (~115 lines) |
| `apps/api/src/routes/llm-proxy.ts` | Register `/mistral-conversations/v1/chat/completions` → `/v1/chat/completions` |
| `apps/api/src/openapi/paths/llm-proxy.ts` | New path entry mirroring the OpenAI doc |
| `apps/cli/src/lib/models.ts` | `PROXY_SUPPORTED_APIS` += `mistral-conversations` |
| `apps/cli/src/commands/run/model.ts` | `buildProxyBaseUrl` returns `${instance}/api/llm-proxy/mistral-conversations/v1` |
| `scripts/verify-openapi.ts` | `expectedEndpoints` += new path |
| `apps/api/test/unit/llm-proxy-adapters.test.ts` | +6 unit tests on the new adapter |
| `apps/api/test/integration/routes/llm-proxy.test.ts` | +2 integration tests (forward+usage row, preset/protocol mismatch 400) |
| `apps/cli/test/run-model.test.ts` | +1 test pinning the proxy base URL for mistral presets |

## Test plan

- [x] `bun run check` — turbo check (TS + ESLint + Prettier + verify-openapi)
- [x] `bun test apps/api/test/unit/llm-proxy-adapters.test.ts` — 20 pass / 0 fail
- [x] `bun test apps/cli/test/run-model.test.ts` — 21 pass / 0 fail
- [ ] `bun test apps/api/test/integration/routes/llm-proxy.test.ts` (needs Docker/PG — runs in CI)
- [ ] Smoke: pin a Mistral preset as org default, run `appstrate run @scope/agent` against a deployed instance — confirm SSE stream and `llm_usage` row with `api='mistral-conversations'`.

## Caveats

- For streaming, Mistral emits `usage` on the terminal SSE frame when the caller opts in (same convention as OpenAI's `stream_options.include_usage`). When the SDK doesn't request it, `parseSseUsage` returns `null` and the metering row is skipped — same fallback as the OpenAI adapter.
- No upstream-key migration: existing Mistral entries in `org_models` / `org_system_provider_keys` already carry the right `baseUrl` and `api` from pi-ai's model registry, so this PR is pure additive routing.

### DIFF
--- a/apps/api/src/openapi/paths/llm-proxy.ts
+++ b/apps/api/src/openapi/paths/llm-proxy.ts
@@ -4,10 +4,11 @@
  * LLM proxy endpoints — server-side model injection for remote-backed
  * AFPS runs (docs/specs/REMOTE_CLI_EXECUTION_SPEC.md §Phase 3).
  *
- * Two protocol families ship today; each gets its own concrete endpoint
+ * Three protocol families ship today; each gets its own concrete endpoint
  * so callers hit the upstream shape they already know (OpenAI Chat
- * Completions, Anthropic Messages). Additional families land as new
- * path entries — the route surface stays concrete per the spec.
+ * Completions, Anthropic Messages, Mistral Chat Completions). Additional
+ * families land as new path entries — the route surface stays concrete
+ * per the spec.
  */
 
 const baseParameters = [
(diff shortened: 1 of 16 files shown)


pierrecabriere:
## Anthropic OAuth fix (commit 39afef0b)

**Problem**: After the OAuth header layer fix (commit 62170095) cleared the 401, the CLI hit an opaque `429 rate_limit_error: "Error"` against Anthropic OAuth tokens — while the dashboard worked. Diagnosed root cause: Anthropic gates `sk-ant-oat-…` tokens to Claude-Code identity at the **body** level (system prompt + tool-name renaming), not just headers. pi-ai injects this body shape locally only when `apiKey.includes("sk-ant-oat")` fires. The runner-pi container path works because the sidecar gives pi-ai an OAuth-shaped placeholder; the CLI's preset path was passing `x-platform-bearer-injected-via-headers`, so pi-ai took the API-key branch and shipped unshaped bodies.

**Fix**: Mirror the runner-pi pattern.

- `GET /api/models` now exposes `keyKind: "oauth" | "api-key" | null` (Anthropic-only). System models read it inline from `ModelDefinition.apiKey`; DB models batch-decrypt by `providerKeyId` to dedupe.
- CLI's `resolvePresetModel` uses `sk-ant-oat-placeholder` when `keyKind === "oauth"` so pi-ai's OAuth branch fires and reshapes the body before it leaves the process.
- `model.headers["Authorization"]` still wins over pi-ai's OAuth bearer because the Anthropic SDK applies `defaultHeaders` AFTER `authHeaders` in `buildHeaders` (later wins), so the proxy still authenticates with the Appstrate token.

**No new headers, no proxy-side body rewrite, no auth pipeline changes.** The proxy adapter's existing OAuth detection (`apps/api/src/services/llm-proxy/anthropic.ts`) already swaps the placeholder bearer for the real upstream OAuth token and merges the caller's `anthropic-beta` with the OAuth-required betas.

Tests: 5 new (3 server-side keyKind detection + 2 CLI placeholder selection). Full test suite green.

PR #352, opened 2026-04-29
fix(auth): bootstrap-token redeem post-review hardening

### LINKED ISSUE #344: Installer: secure-by-default signup mode + interactive prompts via /dev/tty for curl|bash
## Problem

`curl -fsSL https://get.appstrate.dev | bash` currently produces a **publicly-open Appstrate instance by default**. No prompt, no warning, no nudge. The closed-mode feature exists (`AUTH_DISABLE_SIGNUP` + `AUTH_BOOTSTRAP_OWNER_EMAIL`) but is reachable only by:

1. Pre-setting `APPSTRATE_BOOTSTRAP_OWNER_EMAIL=…` before `curl|bash` (requires reading docs first), OR
2. Editing the generated `.env` post-install (the 3-line footer in `.env` is invisible to operators who never `cat .env`).

### Current code path

`apps/cli/src/commands/install.ts:227` — `resolveBootstrapEmail()`:

```ts
if (fromEnv) { /* env var wins */ }
if (opts.mode === "upgrade") return {};
if (opts.nonInteractive) return {};        // ← curl|bash falls here
if (opts.tier === 0) return {};
// otherwise: interactive clack prompt
```

The pipe `curl … | bash`:
- closes stdin (no TTY) → `nonInteractive = true`
- triggers `--yes` automatically (cf. installer output: `--yes: Tier 3 selected automatically`)

→ Returns `{}` → no `AUTH_DISABLE_SIGNUP`, no `AUTH_BOOTSTRAP_OWNER_EMAIL` → **open mode**.

This is the worst-of-three-worlds:
- Not interactive (operator can't choose)
- Not closed-by-default (instance is vulnerable until docs are read)
- No bootstrap-admin signal (no indication that action is required)

### Pattern A — Rustup: `/dev/tty` reopening

`rustup-init.sh` solves this exact problem since 2015. It stays **interactive even when piped from `curl`** by reconnecting `/dev/tty` directly:

> *« This script was piped into `sh` though and doesn't have stdin to pass to its children. Instead we're going to explicitly connect /dev/tty to the installer's stdin. »*

Logic:
- `[ ! -t 0 ]` detects piped stdin
- If stdout is still a TTY (`[ -t 1 ]`) → reopen `/dev/tty` for the installer's stdin
- If neither stdin nor stdout is a TTY (real CI / Dockerfile / cloud-init) → fail fast with:
  > *« Unable to run interactively. Run with `-y` to accept defaults, `--help` for additional options »*

`curl https://sh.rustup.rs | sh` therefore stays interactive in the user's terminal. `-y` must be passed explicitly to skip prompts. Silent default-acceptance is refused.

### Pattern B — Plausible CE: secure-by-default, no interactivity at all

Plausible refuses to boot without explicit config:
- `ADMIN_USER_EMAIL`, `ADMIN_USER_NAME`, `ADMIN_USER_PWD` required
- `DISABLE_REGISTRATION=true` is the default

→ No "open mode" reachable by accident. The admin is bootstrapped from env vars (same primitive Appstrate already has) but it's **mandatory**, not optional.

### Pattern C — Keycloak / OpenStack: temporary admin to revoke

Keycloak creates a temporary admin on first boot, the operator MUST create the permanent one and explicitly remove the temp. OpenStack is the same with `keystone-manage bootstrap` + mandatory removal of `AdminTokenAuthMiddleware`. Higher operator cost, stricter security.

### What the three patterns agree on

**Security must be reached by default, not by reading docs post-install.** None of these tools ship a quietly-open instance after a one-line installer.

## Proposal

A 3-layer fix combining Rustup's interactivity primitive with Plausible's secure-default semantics.

### Layer 1 — Rustup-style interactive prompts via `/dev/tty`

In the bootstrap shell script (`get.appstrate.dev`) and in the CLI itself:

```sh
# In the curl-served wrapper:
if [ ! -t 0 ] && [ -t 1 ] && [ -e /dev/tty ]; then
  # Piped from curl, but user has a controlling terminal — reattach.
  exec </dev/tty
fi
```

In `apps/cli/src/commands/install.ts`, the `nonInteractive` detection should treat the post-reattach state as interactive. The clack prompts in `resolveBootstrapEmail` then fire normally for the dominant `curl|bash` case (95% of installs are devs at a terminal).

Acceptance:
- `curl -fsSL https://get.appstrate.dev | bash` shows the bootstrap-email prompt in the terminal.
- `curl … | bash --yes` skips the prompt (explicit override preserved).
- `APPSTRATE_BOOTSTRAP_OWNER_EMAIL=… curl … | bash` skips the prompt (current behavior preserved).
- CI / Dockerfile `RUN` / cloud-init (no `/dev/tty`) → see Layer 2.

### Layer 2 — Secure default in genuinely non-interactive environments

When no TTY is reachable AND no override is set, the installer must NOT silently open the instance. Two sub-options, pick one:

**2a. Fail-fast (Rustup-strict)** — refuse to install:
```
Error: Cannot install non-interactively without an admin email.
  Pass APPSTRATE_BOOTSTRAP_OWNER_EMAIL=you@example.com (recommended), or
  Pass APPSTRATE_ALLOW_PUBLIC_SIGNUP=1 to acknowledge an open instance.
```

**2b. Closed-by-default (Plausible-strict)** — boot with `AUTH_DISABLE_SIGNUP=true` and no bootstrap admin, then surface "Set up your admin" in the post-install banner + first-page UX. Operator must claim the instance via a one-time token (printed to stdout at install time, regenerable via `appstrate bootstrap-token`).

Recommendation: **2b** for the curl|bash path (the installer always succeeds, no operator footgun), **2a** as an opt-in via `--strict-bootstrap` flag for users who want the rustup behavior. The token-claim flow is more work than 2a but eliminates the "I just ran curl|bash on my server, what now" cliff.

### Layer 3 — Loud post-install banner when an instance ends up open

For any path that lands on open mode (whether by Layer 1 user choice or Layer 2 Plausible-style closed default), the final installer output must include a 5-line ASCII banner that's hard to miss:

```
┌─────────────────────────────────────────────────────────────────┐
│  ⚠ Public signup is ENABLED on this instance.                   │
│  To lock it down:                                               │
│    cd ~/appstrate && $EDITOR .env                               │
│    Set AUTH_DISABLE_SIGNUP=true and AUTH_BOOTSTRAP_OWNER_EMAIL  │
│  Docs: https://appstrate.dev/docs/self-hosting/auth-modes       │
└─────────────────────────────────────────────────────────────────┘
```

The footer in the generated `.env` should stay (operators who do read it benefit), but it cannot be the only signal — most never read `.env` after generation.

## Why this is the right shape

- **Layer 1** matches what the dominant install path (`curl | bash` from a dev's laptop) actually wants: a prompt. Today we silently skip it because we mistake "stdin piped" for "user is a robot".
- **Layer 2** removes the silent-open footgun in genuinely scripted environments. Either we refuse, or we ship closed and require explicit claim — both are safer than the status quo.
- **Layer 3** is a cheap belt-and-suspenders for any path where the choice still lands on open. Banner + `.env` footer + post-install URL = three independent surfaces, only one needs to land.

The 3-layer combo is exactly the convergence of rustup (interactivity primitive) + Plausible (secure-default semantics) + Keycloak (operator must claim before the instance is usable). All three are battle-tested in the same product class.

## Acceptance criteria

- [ ] `get.appstrate.dev` bootstrap script reattaches `/dev/tty` when stdin is piped but a controlling terminal is reachable.
- [ ] `apps/cli/src/commands/install.ts:resolveBootstrapEmail` treats post-reattach state as interactive (the `nonInteractive` flag flips off when `process.stdin.isTTY` is true after reattach).
- [ ] In genuinely non-interactive environments without env override, the installer either fail-fasts or boots closed (decision pending — see proposal Layer 2). Default behavior MUST NOT be silent open mode.
- [ ] If the install ends up open (any reason), final installer output includes the banner from Layer 3.
- [ ] `--yes` / `APPSTRATE_YES=1` preserves current "skip all prompts" semantics — explicit opt-in, not the default for curl|bash.
- [ ] Tests cover: (a) piped stdin + TTY stdout → prompt fires, (b) piped stdin + no TTY + no env override → fail-fast or closed-default, (c) `--yes` → defaults applied, (d) `APPSTRATE_BOOTSTRAP_OWNER_EMAIL=…` → closed mode configured, no prompt.
- [ ] Docs updated: `examples/self-hosting/AUTH_MODES.md`, the docs page on `https://appstrate.dev/docs/self-hosting/`, and the installer README.

## Open questions

1. **Layer 2: 2a vs 2b?** Fail-fast is simpler and matches rustup. Closed-default-with-claim-token is friendlier for the "I just want to try it" cloud-init path. Lean 2b but worth a discussion.
2. **Token claim flow (if 2b)** — printed to stdout at install (loses on cloud-init), regenerable via `appstrate bootstrap-token` (requires SSH access)? Both?
3. **Should `--yes` keep its current "open mode" semantics, or also default-closed?** Stripe-style "be safe by default even with --yes" argues yes; UX-friction argues no. Probably keep `--yes` = current (explicit opt-in to silent open) and document loudly.

## References

- [rustup-init.sh — `/dev/tty` reattach pattern](https://github.com/rust-lang/rustup/blob/main/rustup-init.sh)
- [Plausible CE — `DISABLE_REGISTRATION` + admin env vars by default](https://plausible.io/docs/self-hosting)
- [Keycloak — bootstrap admin & recovery flow](https://www.keycloak.org/server/bootstrap-admin-recovery)
- [OpenStack Keystone — `keystone-manage bootstrap` + middleware removal](https://docs.openstack.org/keystone/pike/admin/identity-bootstrap.html)
- [DEV — How to build a trustworthy `curl | bash` workflow](https://dev.to/operous/how-to-build-a-trustworthy-curl-pipe-bash-workflow-4bb)
- [Cloud Posse — the curl-bash pattern](https://docs.cloudposse.com/resources/glossary/curl-bash/)
- Existing related code: `apps/cli/src/commands/install.ts:227` (`resolveBootstrapEmail`), `apps/cli/src/lib/install/secrets.ts:69` (`BootstrapOverrides`), `apps/cli/src/lib/install/secrets.ts:115` (closed-mode env emission), `apps/cli/src/lib/install/secrets.ts:180` (`.env` footer)

pierrecabriere:
## Summary

Follow-up to #347 — four post-review fixes that didn't make it into the merged PR (commit landed after the merge button was clicked).

- **Force `realm: "platform"` when `withBootstrapTokenRedemption` is active** (`packages/db/src/auth.ts`). The redeem route forwards `c.req.raw.headers` to BA's `signUpEmail`. Without this guard, a stray `oidc_pending_client` cookie on the redeem request would route the bootstrap owner into an `end_user:<applicationId>` realm — wrong audience for an instance-owner row, and unrecoverable once committed.

- **Boot-time reconciliation of the in-memory consumed flag** (new `reconcileBootstrapTokenAtBoot` in `apps/api/src/lib/bootstrap-token.ts`, wired from `boot.ts`). When `AUTH_BOOTSTRAP_TOKEN` is still configured but at least one org already exists, mark consumed at startup so the SPA stops sending returning visitors to `/claim` only to hit a 410 on submit.

- **Document the `TRUST_PROXY` requirement for the redeem rate-limit** (`examples/self-hosting/AUTH_MODES.md`). The per-IP limiter degrades to a global counter behind a reverse proxy without `TRUST_PROXY` configured. State-machine table updated to reflect the new boot-time reconcile (Pending? = no\* once an org exists).

- **Remove dead `Headers.getSetCookie` fallback** (`apps/api/src/routes/auth-bootstrap.ts`). Guaranteed on Bun, Node 18.14+, and modern Undici — the runtimes Appstrate supports.

## Test plan

- [x] `bunx tsc` clean on `apps/api` + `packages/db`
- [x] `bun test apps/api/test/unit/bootstrap-token.test.ts` — 15/15 pass
- [x] `bun test apps/api/test/integration/auth-bootstrap-redeem.test.ts` — 10/10 pass
- [x] `bun test apps/api/test/integration/auth-bootstrap-org.test.ts apps/api/test/integration/middleware/` — 52/52 pass
- [ ] CI must run the full suite

## Refs

- Follows #347 (merged 2026-04-29). Fixes #344-related security audit findings raised post-merge.


pierrecabriere:
Superseded — fixes will be folded back into the reopened #347.

PR #370, opened 2026-05-08
fix(runtime): gate Checkpoint/Pinned-Slots/Memory footers on tool availability (#368)

### LINKED ISSUE #368: Prompt sections reference platform tools (`pin`, `note`, `recall_memory`) without checking they're in the agent's dependencies
## Context

Follow-up from #362 review.

The platform prompt builder (`packages/afps-runtime/src/bundle/platform-prompt.ts`) has three sections that hard-code references to AFPS platform tools (`pin`, `note`, `recall_memory`):

| Section | Line | Gating today |
| --- | --- | --- |
| `## Checkpoint` | `platform-prompt.ts:316` | `context.checkpoint != null` |
| `## Pinned Slots` | `platform-prompt.ts:339` | `context.pinnedSlots` non-empty |
| `## Memory` | `platform-prompt.ts:367` | **always emitted** (per ADR-012, agent always needs to know `recall_memory` exists) |

All three footers (lines 327, 356–357, 379–386) instruct the agent to call `pin(...)` / `note(...)` / `recall_memory(...)`.

But these tools are **not** ambient. They live in `packages/afps-runtime/src/resolvers/platform-tools.ts:84-127` and ship as AFPS 1.3+ packages (`@afps/state`, `@afps/memory`, …). Agents must declare them in `dependencies.tools[]` to get them at runtime.

## Problem

The sections are gated on **data presence**, not on **tool availability**. They diverge in two scenarios:

1. **Tool dependency removed across versions.** Agent v1 ships `@afps/state` → users accumulate rows in `package_persistence`. Agent v2 drops `@afps/state`. Result: prompt v2 still renders `## Pinned Slots` from the v1 rows and tells the LLM to call `pin(...)`, but the tool isn't registered. The LLM then hallucinates calls or reports an unrelated error.
2. **Memory section is unconditional.** Even an agent that ships zero AFPS tools (no `@afps/memory`, no `@afps/state`) still gets a `## Memory` section telling it to call `note`, `recall_memory`, and `pin`. This is the worst case because no data needs to exist for the broken instruction to be emitted.

In the happy path (deps stable, fresh agent), this doesn't trigger — no rows means no rendering — so #362 didn't cause regressions when merged. But the contract leak is still there.

## Proposed fix

Gate each section's footer (and, for `## Memory`, the section itself when no archive tool is present) on the corresponding tool being in `availableTools`:

```ts
const hasPin     = opts.availableTools?.some((t) => t.name === "pin")           ?? false;
const hasNote    = opts.availableTools?.some((t) => t.name === "note")          ?? false;
const hasRecall  = opts.availableTools?.some((t) => t.name === "recall_memory") ?? false;
```

Section-by-section behaviour:

- **`## Checkpoint`** — render the data block unconditionally (the snapshot is informative regardless), but emit the "call `pin({ key: \"checkpoint\" })`" footer only when `hasPin`. Without `pin`, append a one-liner explaining the data is read-only carry-over.
- **`## Pinned Slots`** — same pattern: render the slot blocks (data is data), gate the update-instructions footer on `hasPin`.
- **`## Memory`** — keep the per-memory list rendering unchanged. Gate each footer sentence on its tool: skip the `note(...)` instruction if `!hasNote`, the `recall_memory(...)` instruction if `!hasRecall`, the `pin(...)` instruction if `!hasPin`. If all three are absent, skip the footer entirely. Consider skipping the whole `## Memory` section (including the "No memories pinned" fallback) when none of `hasNote || hasRecall || hasPin` holds, since the section's only purpose is to teach the agent about archive APIs it doesn't have.

## Open design questions

1. **Stale data, missing tool — how loud?** When `## Pinned Slots` has rows but `pin` is gone, do we (a) silently render data without an update footer, (b) render data plus a one-liner like "These slots are read-only in this build — `pin` is not a registered tool", or (c) skip the section entirely? (c) hides legitimately useful state from the LLM; (b) is most explicit but adds prompt noise; (a) is simplest. I lean (b).
2. **`## Checkpoint` semantics with no `pin`.** Same question. Read-only carry-over still has value (resume context), so render-with-warning seems right.
3. **Detection by name vs. by id.** `availableTools` items have both `id` and `name` (`platform-prompt.ts:83` → `PlatformPromptTool`). The reserved names `pin` / `note` / `recall_memory` come from the AFPS spec, not from package ids — so checking `t.name === "pin"` is correct, but a custom tool could shadow the name. Worth deciding whether to also enforce a known package-id allowlist. Probably overkill — the AFPS spec reserves these names.
4. **`recall_memory` is provided by the sidecar's MCP layer** (`runtime-pi/sidecar/mcp.ts:500`), not by an `@afps/*` tool package. It's surfaced into `availableTools` via `runtime-pi/extensions/mcp-direct.ts` when the run wires MCP. Verify the same `availableTools.some(t => t.name === "recall_memory")` check works there — if MCP tools take a different code path, we may need to plumb a separate signal.

## Tests to add

In `packages/afps-runtime/test/bundle/platform-prompt.test.ts`:

- `## Checkpoint` rendered without footer when `pin` missing from `availableTools`
- `## Pinned Slots` rendered without footer when `pin` missing
- `## Memory` footer sentences gated individually on `hasNote` / `hasRecall` / `hasPin`
- `## Memory` section omitted entirely when none of the three tools are present
- All current tests still pass when the relevant tools ARE present (regression check on #362's added tests)

## Out of scope

Audit of other prompt sections (`## Output Format`, `## Connected Providers`) for the same drift. They reference `output` and `provider_call` — same architectural issue, same fix pattern, but separate from the memory/state surface. Track in a follow-up if needed.

## References

- PR #362 — original review thread
- `packages/afps-runtime/src/bundle/platform-prompt.ts:316-387` — affected sections
- `packages/afps-runtime/src/resolvers/platform-tools.ts:84-127` — `pin` / `note` tool definitions
- `runtime-pi/sidecar/mcp.ts:500` — `recall_memory` MCP tool
- ADR-012 / ADR-013 — memory-tier rationale (motivates always-emit `## Memory` today)

pierrecabriere:
## Summary

Fixes #368. The platform prompt's `## Checkpoint`, `## Pinned Slots`, and `## Memory` sections referenced `pin` / `note` / `recall_memory` unconditionally, even when those tools weren't wired for the run. Each section's footer is now gated on the relevant tool being present in `availableTools`; the `## Memory` section is omitted entirely when none of `pin` / `note` / `recall_memory` are wired.

## Failure modes addressed

1. **Stale data after a dep removal.** Agent v1 ships `@afps/state`, accumulates pinned-slot rows; agent v2 drops the dep. Before this fix, the v2 prompt still rendered `## Pinned Slots` from v1 rows and instructed the LLM to call `pin(...)` — which isn't registered. Now the data block still renders (it's informative), but the update footer is replaced by a one-liner explaining the slots are read-only carry-over.
2. **Memory section unconditional.** Agents shipping zero AFPS tool packages still got a `## Memory` section telling them to call `note` / `recall_memory` / `pin`. Now the section is omitted entirely when none are wired.

## Implementation

- `renderPlatformPrompt` computes `hasPin` / `hasNote` / `hasRecall` from `availableTools[].name` and gates each footer accordingly. Per-sentence gating in `## Memory` so partially-wired surfaces (e.g. only `note` available) get the right subset.
- `## Checkpoint` and `## Pinned Slots`: data block always renders; footer swaps to a "read-only carry-over" one-liner when `pin` is missing (issue's design Q1/Q2 — option (b)).
- `## Memory`: whole section omitted if none of `note` / `recall_memory` / `pin` are wired.

Two collateral fixes were needed for the gating to work end-to-end:

- **`toolFromPackage` precedence** (`platform-prompt-inputs.ts`): now reads `manifest.tool.name` (the LLM-facing tool name per AFPS 1.5+ tool packages) before falling back to `manifest.name`. Without this, the gate `t.name === "pin"` never matched real bundles — `@appstrate/pin`'s manifest has `name: "@appstrate/pin"` and `tool: { name: "pin" }`. The package display name is incidental; the LLM-facing identifier is what matters.
- **Platform-injected runtime tools** (`apps/api/services/adapters/prompt-builder.ts`): `recall_memory` and `run_history` are wired by `runtime-pi/mcp/direct.ts` for every platform run but aren't bundle packages, so `buildPlatformPromptInputs` can't derive them. The platform prompt-builder now appends them to `availableTools` after bundle derivation, addressing the issue's open question Q4. CLI runs (which don't wire these tools) correctly omit them.

## Tests

- `platform-prompt.test.ts`: 11 new tests under `reserved-tool gating (#368)`:
  - Checkpoint footer gated on `pin` (rendered with read-only warning when absent, full instruction when present)
  - Pinned Slots footer same gating
  - Memory section omitted when none of the three tools are wired (incl. with stale memory data — issue scenario #1)
  - Each footer sentence gated independently for the only-`note`, only-`recall_memory`, only-`pin` cases, and full footer when all three are wired
  - Pinned-memory list still renders (issue invariant: data block unchanged)
  - Custom-tool name shadowing activates the gate (documented Q3 side-effect)
- `platform-prompt-inputs.test.ts`: 2 new tests for the `manifest.tool.name` precedence and `manifest.name` fallback.
- `platform-prompt-parity.test.ts`: fixture extended with three reserved-tool packages so the cross-path parity assertions keep exercising every section.
- `prompt-builder.test.ts` (apps/api): updated the "no tools" assertion to reflect that platform-injected `run_history` + `recall_memory` always surface.

## Test plan

- [x] `cd packages/afps-runtime && bun test` — 533 pass, 0 fail
- [x] `cd packages/afps-runtime && bun run check` — typecheck pass
- [x] `cd apps/api && bun test test/unit/prompt-builder.test.ts` — 44 pass, 0 fail
- [x] `cd apps/api && bun run typecheck` — pass
- [x] `cd packages/runner-pi && bun test` — 131 pass, 0 fail (regression check)
- [x] `bun run detect:breaking` — 0 breaking changes (3 unrelated non-breaking changes already on main)
- [x] Prettier + ESLint on all changed files

## Out of scope

Per the issue: auditing other prompt sections (`## Output Format` referencing `output`, `## Connected Providers` referencing `provider_call`) for the same drift pattern. Track in a follow-up if needed.


pierrecabriere:
Second commit pushed (`refactor(runtime): tools own their prose via TOOL.md, platform owns data shells`) addressing the architectural concern raised during review: the prior approach hardcoded `pin` / `note` / `recall_memory` names in `platform-prompt.ts`, which would force editing that file every time a new state-bearing tool is added.

**Now**:
- Platform owns SECTIONS (headers + data dumps from DB).
- Tools own USAGE PROSE (each ships a `TOOL.md`, or a `TOOL.md`-equivalent fragment in `RUNTIME_INJECTED_TOOLS` for tools wired outside the bundle).
- Adding `@afps/journal` or any new state tool = ship the package; zero changes to the platform.

**Net diff for the refactor commit**: -65 lines on top of the prior gating. `platform-prompt.ts` lost 92 lines (-135 +43); the new `runner-pi/src/runtime-injected-tools.ts` adds 117 lines as the single source of truth shared by the runtime container (`runtime-pi/mcp/direct.ts`) and the platform prompt builder (`apps/api/services/adapters/prompt-builder.ts`).

**Test impact**: 978 tests pass across the four affected suites (afps-runtime 526, runner-pi 131, runtime-pi 277, apps/api prompt-builder 44). The 11 gating-specific tests added in the prior commit were replaced with data-driven rendering tests.

pierrecabriere:
Third commit pushed (`refactor(runtime): collapse runtime-injected tools to one descriptor list`) — pushes the modularity to the limit.

**The remaining hardcode in commit 2 was**: `runtime-pi/mcp/direct.ts` had two near-identical `make<Tool>Extension` factories, each duplicating the call-forwarding boilerplate. Adding a new runtime-wired tool would have meant editing three places.

**Now**: one descriptor in `RUNTIME_INJECTED_TOOLS` carries `id`, `name`, `description`, `parameters` (JSON Schema), `doc` (TOOL.md prose). A single generic `makeMcpForwardExtension(tool, opts)` consumes the descriptor — registers Pi tool, emits `${name}.called`/`.completed` events, forwards to MCP. The `buildMcpDirectFactories` registration loop iterates the descriptor list; the sidecar-advertisement guard does the same.

**Adding a new runtime-injected tool now**: append one entry to `runtime-injected-tools.ts`. The registration, the prompt's `### Tools` listing, the prompt's TOOL.md-equivalent doc, the sidecar guard, and the event taxonomy all wire up by iteration.

**Net diff for this commit**: -71 lines in `direct.ts` (duplication gone), +18 lines on the descriptor (parameters fields), +98 lines of contract tests.

**Tests**: 984 pass (+6 new contract tests in `packages/runner-pi/test/runtime-injected-tools.test.ts` locking the descriptor invariants — fields populated, doc heading convention, JSON Schema purity, name uniqueness).

pierrecabriere:
Fourth commit pushed (`refactor(runner-pi): per-tool directory layout for runtime-injected tools`) — completes the SOC pass.

**Question raised**: in commit 3 the descriptor and the doc still cohabited in a single `runtime-injected-tools.ts`. SOC isn't really respected — adding a new tool means editing a shared file that mixes multiple tools' concerns.

**Fix**: each runtime-injected tool is now its own self-contained directory mirroring the bundle-tool layout:

```
packages/runner-pi/src/runtime-tools/
├── index.ts                  ← aggregates RUNTIME_INJECTED_TOOLS
├── types.ts                  ← RuntimeInjectedTool interface
├── run-history/
│   ├── tool.ts               ← descriptor + parameters
│   └── TOOL.md               ← LLM-facing prose
└── recall-memory/
    ├── tool.ts
    └── TOOL.md
```

`tool.ts` loads its co-located `TOOL.md` synchronously via `readFileSync(new URL("./TOOL.md", import.meta.url))`. (Tried Bun's `import doc from "./TOOL.md" with { type: "text" }` first, but TS can't resolve it through cross-package typechecking — `apps/api` transitively typechecks runner-pi sources and ambient `*.md` declarations don't propagate across project boundaries. `readFileSync` is plain TS, works everywhere, same effective behavior since both run at module-import time.)

**Adding a new runtime-injected tool now**: one new directory + one line in `index.ts`. Identical mental model to bundle tools (`scripts/system-packages/tool-<name>-<version>/`). A future move of `recall_memory` from "runtime-injected" to "bundle-shipped" is a directory copy — the prose and parameters travel together.

**Net**: 9 files changed, +253/-159. Two new contract tests lock the layout: directory count matches `RUNTIME_INJECTED_TOOLS.length`, and `tool.doc === readFileSync(co-located TOOL.md)` so a future regression where someone inlines the prose back into `tool.ts` fails loudly.

**Tests**: 986 pass (afps-runtime 526, runner-pi 139 — +2 layout tests, runtime-pi 277, apps/api prompt-builder 44).

PR #395, opened 2026-05-10
feat: add llm_usage.billed_at for ledger-first billing (Layer 4)

pierrecabriere:
## Summary

Adds a `billed_at` timestamp column on `llm_usage` so the cloud module can switch from snapshot-billing (`runs.cost` SUM at finalize time) to ledger-first billing (per-row claim with `billed_at IS NULL` as the work queue).

This is **Layer 4** of the billing-correctness plan:
- **Layer 1** (#394) routed every terminal-state path through `finalizeRun`, closing the cancel-bypass leak.
- **Layer 3** ([appstrate/cloud#14](https://github.com/appstrate/cloud/pull/14)) added a periodic sweeper that catches any `runs.cost > 0 AND no cloud_usage_records` gap.
- **Layer 4** (this PR + [appstrate/cloud#15](https://github.com/appstrate/cloud/pull/15)) eliminates the temporal coupling between the cost write and the billing debit. A late-arriving `llm_usage` row (sidecar buffer flush after `finalizeRun` already ran, runner reconnect) automatically enters the cloud sweeper's work queue because `billed_at IS NULL` is the canonical signal — no platform-side coordination required.

## Changes

- **`packages/db/src/schema/runs.ts`** — new `billedAt` column on `llm_usage`. Doc explains: cloud-only, single writer, single reader, kept on the row to avoid the JOIN tax of a side-table.
- **`packages/db/drizzle/0020_llm_usage_billed_at.sql`** — `ALTER TABLE … ADD COLUMN billed_at TIMESTAMPTZ` + `CREATE INDEX idx_llm_usage_unbilled_run_id ON llm_usage (run_id) WHERE billed_at IS NULL`. The partial index makes cloud's sweeper SELECT a cheap range scan; once a row is billed it leaves the index automatically.
- **`packages/db/drizzle/meta/_journal.json`** — registers the migration.

## Why on `llm_usage` (platform-owned) rather than a side-table cloud could own?

I considered `cloud_billed_llm_usages(llm_usage_id, billed_at)` to give cloud full ownership of its data. Rejected because:
- Doubles the partial-index footprint (we'd need both `idx_llm_usage_unbilled_run_id` AND a FK index on the side-table).
- Forces a JOIN on every billing tick (cloud would have to LEFT JOIN to detect "rows we haven't seen yet").
- A single nullable column the platform never reads is the smallest possible coupling. The platform can't even tell whether the cloud module is loaded.

## Test plan

- [x] `bun run check` — turbo check + verify-openapi clean
- [x] Migration applies cleanly on PostgreSQL test DB (verified — column + partial index present after `bun test`)
- [x] No platform-side test coverage required (the column is informational from the platform's perspective; the contract lives in cloud#15's tests)

## Out of scope

- Billing logic — lives in `appstrate/cloud#15` (the ledger-first `recordUsage` rewrite + backfill migration).
- Renaming any existing column or breaking the `runs.cost` snapshot semantic — the snapshot is still computed and stored, it just stops being the source of truth for billing.

## Companion PR

[appstrate/cloud#15](https://github.com/appstrate/cloud/pull/15) — the cloud-side ledger-first `recordUsage` + backfill migration. Merge order: this PR first (column must exist before cloud reads it), then cloud#15.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -447,11 +447,17 @@ export function requireAdmin() {
 
 ### Orphaned run recovery
 
-On platform startup, any runs left in `pending` or `running` state from a previous crash are marked as `failed`. This prevents stale run tokens from remaining valid:
+On platform startup, any runs left in `pending` or `running` state from a previous crash are listed and finalized through `synthesiseFinalize` → `finalizeRun`. Routing through the canonical terminal-state pipeline (rather than a direct `UPDATE`) ensures the `afterRun` hook fires for every orphan, so billing and observability modules see the exact same lifecycle whether a run terminated cleanly, was cancelled, timed out, or was killed by a server crash. This also prevents stale run tokens from remaining valid:
 
 ```typescript
-// index.ts — startup
-await markOrphanRunsFailed();
+// boot.ts — startup
+const orphanIds = await listOrphanRunIds();
+for (const runId of orphanIds) {
+  await synthesiseFinalize(runId, {
+    status: "failed",
+    error: { message: "Server restarted while run was in progress. Please retry." },
+  });
+}
 ```
 
 **Standard:** This multi-layer authentication model implements the access control architecture described in **NIST SP 800-207** (Zero Trust Architecture): per-request verification, no implicit trust, and session-scoped credentials with automatic expiration.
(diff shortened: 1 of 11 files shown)


pierrecabriere:
Fermé : violation de la frontière open-core. Ajouter `billed_at` à `llm_usage` dans le schéma core fait que le repo OSS connaît un détail d'implémentation du module cloud (la colonne n'a aucune sémantique en OSS, l'index partiel `WHERE billed_at IS NULL` non plus).

La règle (CLAUDE.md) : *"All billing tables are Cloud-owned and created at boot via independent Drizzle migrations on the same PostgreSQL database."* Cloud doit posséder son propre état de facturation, pas muter une table core.

Réécrit côté cloud via une table side-car `cloud_billed_llm_usage` (PK `llm_usage_id BIGINT`) — voir le re-spin de cloud#15. Aucun changement OSS requis.

PR #420, opened 2026-05-13
fix(queue): give BullMQ Worker its own duplicated Redis connection

pierrecabriere:
## Summary

`BullMQQueue` was wiring its Worker against the shared `getRedisConnection()` publisher (`bullmq-queue.ts:92`). `worker.close()` disconnected that shared socket while a BRPOP was in flight; ioredis flushed the pending command with `new Error("Connection is closed")` and the rejection escaped BullMQ's run loop as an unhandled rejection — `bun:test` marks the file as failed.

First test to hit this is `apps/api/test/integration/services/model-providers-pairing-cleanup-worker.test.ts` (introduced by #397). The test itself passes; the file fails on `# Unhandled error between tests`. Integration tests passed on run [25771906264](https://github.com/appstrate/appstrate/actions/runs/25771906264) and failed on run [25772493808](https://github.com/appstrate/appstrate/actions/runs/25772493808) with identical test code → flaky by shutdown timing.

## Changes

- Worker gets `getRedisConnection().duplicate()` (BullMQ-owned, isolated).
- Swallow the specific `Connection is closed.` error on that duplicate.
- `shutdown()` explicitly `quit()`s the duplicate after `worker.close()`.

The Queue keeps the shared publisher (non-blocking ops, no blocking client). The scheduler queue and refresh-worker queue use the same `BullMQQueue` wrapper and benefit from the same fix — they don't currently hit this path because they have no integration test exercising full init→shutdown, but the leak was latent.

## Test plan

- [ ] CI Integration job goes green on this branch.
- [ ] Re-run CI a couple of times to confirm the flake is gone.
- [ ] Local: `cd apps/api && bun test test/integration/services/model-providers-pairing-cleanup-worker.test.ts` runs clean.

### DIFF
--- a/apps/api/src/infra/queue/bullmq-queue.ts
+++ b/apps/api/src/infra/queue/bullmq-queue.ts
@@ -6,6 +6,7 @@
 
 import { Queue, Worker, UnrecoverableError } from "bullmq";
 import type { ConnectionOptions } from "bullmq";
+import type Redis from "ioredis";
 import { getRedisConnection } from "../../lib/redis.ts";
 import { logger } from "../../lib/logger.ts";
 import type {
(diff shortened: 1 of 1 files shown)


pierrecabriere:
Closing — see review in conversation. Approach was a band-aid (string-match on ioredis error message, hybrid shared/duplicate connection pattern, no local repro). Reopening with a proper fix: pass ConnectionOptions config so BullMQ owns the full connection lifecycle, plus a deterministic regression test.

Issue #437, opened 2026-05-13
epic: full Portkey integration — drop custom LLM gateway code, adopt open catalog


pierrecabriere:
> **Epic — Full Portkey integration.** Supersedes #436 (decision spike, validated 2026-05-13).
>
> **Status (2026-05-13)**: Implementation **shipped** in PR #439. All phases except 4b/4c landed. Merging the PR closes this epic.

# Outcome

Phase 1 → 3 + 4a shipped: Portkey is mandatory and inline-config'd for every API-key LLM call ; pricing catalog migré vers BerriAI/litellm (~400 modèles indexés, capabilities + cost) ; picker UX ouvert (featured + all-models split pour les providers couverts par le catalogue). Phase 4b (fallback chains) deferred ; Phase 4c (OTel) dropped — bundle 1.15.2 OSS n'embarque ni surface OTel/Prometheus ni calcul de coût en \$. Voir PR #439 pour le rationale du compromis.

# Why

The #427 incident exposed three gaps in our LLM call path:
1. **No org-level TPM rate limit** — per-run fan-out cap (#429) protects one run, not 10 concurrent runs in the same org
2. **No `Retry-After` honoring + jittered backoff** — SDK Pi's internal retry is opaque, hits TPM windows back-to-back, exhausts silently
3. **No circuit breaker** — context-growth runaway (the actual #427 trigger) is not detected

The roadmap drafted to fix this (#431/#432/#433/#434) was building a SOTA rate-limit/retry/circuit-breaker stack from scratch with `rate-limiter-flexible` + `cockatiel` + OpenTelemetry — about 2-3 weeks of careful work, mostly reinventing the wheel.

In March 2026, [Portkey AI Gateway](https://github.com/Portkey-AI/gateway) went fully open-source (MIT). It is purpose-built for the LLM gateway role and ships:
- Token-aware rate limiting per virtual key / inline config
- Retry with `Retry-After` honoring + exponential backoff + jitter
- Circuit breakers with provider failover
- 1 600+ provider catalog (though only ~70 are actually compiled into the OSS gateway bundle — verified post-spike)

A 2-day spike on 2026-05-13 confirmed feasibility under Bun + Tier 0 + inline config.

# Phases — final status

## Phase 1 — Stateless proxy integration — ✅ SHIPPED

[...]

PR #438, opened 2026-05-13
feat(modules): portkey gateway phase 1 (#437)

pierrecabriere:
## Summary

Opt-in `portkey` built-in module that spawns the [open-source Portkey AI Gateway](https://github.com/Portkey-AI/gateway) as a Bun sub-process and re-points the sidecar `/llm/*` proxy at it via inline `x-portkey-config`. Closes phase 1 of #437.

- Subscription-OAuth providers (Codex, Claude Pro) bypass Portkey — they keep using the sidecar's existing OAuth wireFormat path.
- Pi SDK retry disabled when Portkey is wired (`MODEL_RETRY_ENABLED=false`) to prevent retry-amplification on 429.
- Not in default `MODULES` — enable with `MODULES=portkey,oidc,webhooks,core-providers,@appstrate/module-codex`.
- Zero footprint when absent: `getPortkeyRouter()` returns null and the legacy direct-upstream path is unchanged.

## Smoke-test findings consumed in this PR

- **SSE pass-through**: 5 concurrent streams, 50 chunks each, max gap 102 ms — reproduced as an integration test.
- **Telemetry**: zero SDKs in the bundle. Open question on `PORTKEY_DISABLE_TELEMETRY` closed.
- **Sub-process drain**: Portkey doesn't drain natively (~1 s SIGTERM-to-exit). The existing `apps/api/src/lib/shutdown.ts` ordering already drains in-flight runs BEFORE `shutdownModules()` — load-bearing.
- **PORT env var**: Portkey reads `--port=<n>` CLI flag, not `PORT` env var. Lifecycle passes both.

## Test plan

- [x] `bun run check` — typecheck + prettier + eslint + verify-openapi (5 cached, 14 fresh, all green)
- [x] `bun test` — full monorepo: **5520 pass / 0 fail / 12607 expects across 413 files**
- [x] Unit: `apps/api/src/modules/portkey/test/unit/config.test.ts` (7 tests)
- [x] Integration: real Portkey spawn → ready signal → upstream sees decrypted credential → SSE chunks pass through unbuffered → idempotent re-start
- [x] Integration: `runPlatformContainer` consults router; falls through to direct upstream when none installed; sets `MODEL_RETRY_ENABLED=false` only when Portkey is active
- [x] Sidecar unit: `x-portkey-config` attached when set, absent on legacy path

## Out of scope (deferred to phase 2+)

- Pricing catalog adoption (`Portkey-AI/models`)
- Open model catalog UX (1 600+ providers)
- `apps/api/src/services/llm-proxy/*` server-side path (used by `@appstrate/github-action`) — currently bypasses Portkey
- In-process Hono mount, semantic cache, multi-provider fallback chains

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -26,6 +26,7 @@
     "@appstrate/shared-types": "workspace:*",
     "@better-auth/oauth-provider": "^1.6.9",
     "@hono/swagger-ui": "^0.6.1",
+    "@portkey-ai/gateway": "^1.15.2",
     "ajv": "^8.20.0",
     "ajv-formats": "^3.0.1",
     "better-auth": "^1.6.9",
--- a/apps/api/src/modules/portkey/README.md
+++ b/apps/api/src/modules/portkey/README.md
@@ -1,0 +1,51 @@
+# `portkey` — built-in module
+
+Phase 1 integration of the open-source [Portkey AI Gateway](https://github.com/Portkey-AI/gateway) as Appstrate's LLM gateway. See issue [#437](https://github.com/appstrate/appstrate/issues/437) for the full epic.
+
+## What it does
+
+When `MODULES=portkey,…` is set, the module:
+
+1. Spawns `@portkey-ai/gateway/build/start-server.js` as a Bun sub-process at `init()`, bound to `127.0.0.1:${PORTKEY_PORT}` (default `8787`).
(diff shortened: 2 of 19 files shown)


pierrecabriere:
Renamed source branch to `feat/portkey-migration-437` — this branch will hold the full Portkey epic (phases 1 → 4) and be merged as a single unit. Replaced by the new PR on the migration branch.

PR #447, opened 2026-05-14
fix(sidecar): honor multipart body shape in provider_call

### LINKED ISSUE #402: bug: `ctx.providerCall` fails when body is a structured `{multipart}` payload
## Symptom

Custom tools that call `ctx.providerCall({ body: { multipart: [...] } })` (the shape declared in the spec) either error out or end up sending an incorrectly serialized body upstream — the sidecar does not recognize the `{multipart}` form and passes it through as-is instead of building a proper `FormData` / multipart wire payload.

## Proposed solution

Commit `e802631d` on `otarbes/appstrate-oss:bugs-evos-oli`: in the sidecar path of `ctx.providerCall`, detect `body.multipart`, construct a `FormData` (or equivalent streaming multipart), set the correct `Content-Type: multipart/form-data; boundary=…`.

## Tests

Add a test case covering a binary upload via multipart to a mock provider.

pierrecabriere:
## Summary

- Add `multipart` body variant to the `provider_call` MCP tool schema (oneOf: string | { fromBytes } | { multipart: [...] }).
- Build `FormData` at the sidecar boundary; let Bun's `fetch()` set the boundary token in `Content-Type: multipart/form-data; boundary=…`.
- Strip caller-supplied `Content-Type: multipart/...` headers so the boundary in the header matches the bytes on the wire.
- Enforce `SIDECAR_MAX_REQUEST_BODY_BYTES` across the **sum** of decoded file-part bytes (mirrors the structured 413 contract used by the `{ fromBytes }` path: `{ code: "PAYLOAD_TOO_LARGE", scope, limit, actual, envVar }`).
- Substitute `{{vars}}` in string-valued field parts only when `substituteBody: true`; binary parts always pass through verbatim.
- Body builder is a closure (`build(activeCreds)`), so the 401-retry path can re-render the FormData with the refreshed token — identical to the buffered text path.

## Out of scope / follow-ups

- The AFPS spec's TS types do not yet declare the `multipart` shape on `ProviderCallBody`. Declared locally in the sidecar handler (runtime contract); a follow-up will lift it into `@afps-spec/schema` so runtime resolvers can construct it with full IDE help.

## Test plan

- [x] Multipart round-trip: upstream sees `multipart/form-data; boundary=…` Content-Type, all field parts and file parts decode via `Request#formData()`.
- [x] Binary file part: raw bytes (PNG header) match after upload.
- [x] Variable substitution scoped to field strings when `substituteBody: true`; binary bytes containing literal `{{access_token}}` text pass through untouched.
- [x] No substitution by default (placeholder reaches upstream as literal text).
- [x] Caller-supplied `Content-Type: multipart/form-data; boundary=AAAAAA` is stripped — `fetch()` picks a fresh boundary.
- [x] 413 + structured `PAYLOAD_TOO_LARGE` when sum of decoded part bytes exceeds `SIDECAR_MAX_REQUEST_BODY_BYTES`.
- [x] Invalid base64 in a file part surfaces the same error message shape as the `{ fromBytes }` path.
- [x] `tools/list` advertises the new variant (3-way `oneOf` on `body`).
- [x] End-to-end JSON-RPC dispatch through the SDK stateless transport.

Closes #402

### DIFF
--- a/runtime-pi/sidecar/credential-proxy.ts
+++ b/runtime-pi/sidecar/credential-proxy.ts
@@ -59,7 +59,17 @@ export type ProviderRequestBody =
   | { kind: "none" }
   | { kind: "buffered"; bytes: ArrayBuffer; text?: string }
   | { kind: "streaming"; stream: ReadableStream }
-  | { kind: "formData"; build: (activeCreds: Record<string, string>) => FormData };
+  | {
+      kind: "formData";
+      build: (activeCreds: Record<string, string>) => FormData;
+      /**
+       * Field-part templates that will undergo `{{var}}` substitution
+       * when `substituteBody: true`. Used by the pre-flight check to
+       * fail-closed on unresolved placeholders — mirrors the buffered
+       * text path. Empty/omitted means no substitution will happen.
+       */
+      fieldTemplates?: string[];
+    };
 
 export interface ProviderCallArgs {
   providerId: string;
@@ -196,7 +206,9 @@ export async function executeProviderCall(
     }
   }
 
-  // 6. Pre-check body placeholder resolution (buffered + substitute path only).
+  // 6. Pre-check body placeholder resolution (buffered text + multipart
+  //    field-parts under substituteBody). Streaming + binary buffered
+  //    bodies are pass-through.
   if (substituteBody && body.kind === "buffered" && body.text !== undefined) {
     const testBody = substituteVars(body.text, creds.credentials);
     const unresolvedInBody = findUnresolvedPlaceholders(testBody);
@@ -205,6 +217,21 @@ export async function executeProviderCall(
         ok: false,
         status: 400,
         error: `Unresolved placeholders in body: {{${unresolvedInBody.join()}}}`,
+      };
+    }
+  }
+  if (substituteBody && body.kind === "formData" && body.fieldTemplates?.length) {
+    const unresolved = new Set<string>();
+    for (const template of body.fieldTemplates) {
+      for (const v of findUnresolvedPlaceholders(substituteVars(template, creds.credentials))) {
+        unresolved.add(v);
+      }
+    }
+    if (unresolved.size) {
+      return {
+        ok: false,
+        status: 400,
+        error: `Unresolved placeholders in body: {{${[...unresolved].join()}}}`,
       };
     }
   }
--- a/runtime-pi/sidecar/credential-proxy.ts
+++ b/runtime-pi/sidecar/credential-proxy.ts
@@ -45,13 +45,21 @@ import { logger } from "./logger.ts";
 
 /**
  * Body modes the proxy core accepts. The HTTP handler can produce
- * any of the three; the MCP handler only ever produces "buffered"
- * (since JSON-RPC carries body as a string).
+ * "none" / "buffered" / "streaming"; the MCP handler produces
+ * "buffered" for text + binary uploads, and "formData" for the
+ * `{ multipart: [...] }` body shape.
+ *
+ * The `formData` variant carries a builder closure rather than a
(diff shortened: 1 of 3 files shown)


pierrecabriere:
## Fresh-eyes review

**Looks good**
- The `build(activeCreds)` closure is the right shape — the 401-retry path in `credential-proxy.ts:300` re-enters `doUpstreamRequest` → `buildBody(refreshed.credentials)` → `body.build(...)`, so the refreshed token reaches the next attempt. Mirrors the buffered text path correctly.
- Content-Type strip at `credential-proxy.ts:250-259` is case-insensitive and scoped to `multipart/*`, which means a caller setting `Content-Type: application/json` on a non-multipart shape is unaffected. Regression risk is low.
- Cap is enforced on the **sum** of decoded bytes pre-FormData assembly (`mcp.ts:230-258`), and the structured `PAYLOAD_TOO_LARGE` shape matches the `{fromBytes}` path exactly. The 16 MB MCP envelope cap (mcp.ts:1485-1491) lands earlier in the pipeline and bounds the base64 string size before this validator runs, so the OOM-via-100MB-base64 vector is covered.

**Concerns**

1. **(hard inconsistency)** The buffered text path fail-closes on unresolved `{{placeholders}}` after substitution (`credential-proxy.ts:200-209`), but the multipart field-part path does not. A field `value: "Bearer {{access_tokn}}"` with `substituteBody: true` and a typo will silently ship the literal `{{access_tokn}}` to the upstream third party. *Fix:* run `findUnresolvedPlaceholders` on each substituted field value in the build closure (or once at validation time against the available cred keys) and return a 400 with the same `Unresolved placeholders in body` shape.

2. **(missing cap)** No `maxItems` on `body.multipart` and no part-count cap in `validateMultipartParts`. An LLM-supplied array of 100k single-byte parts fits the envelope cap and the body-size cap (sum-of-decoded-bytes is tiny) but allocates 100k `Blob`s + `FormData` entries. *Fix:* add `maxItems: 256` (or similar) to the JSON Schema and an early `if (parts.length > MAX_MULTIPART_PARTS) return …` guard.

3. **(missing cap)** No length cap on `name`, `filename`, or `contentType`. Each is caller-supplied and ends up in part headers; nothing prevents a 10 MB filename string. The decoded-bytes cap doesn't count these. *Fix:* enforce `maxLength` on `name`/`filename`/`contentType` in the schema and in the validator (e.g. 256/1024/256).

4. **(weak runtime guard, speculative)** `validateMultipartParts(args.body.multipart)` trusts `multipart` to be an array — the MCP SDK does not validate `tools/call` args against `inputSchema`. A non-array `multipart` (`null`, a string, an object) either no-ops or emits a confusing per-char error. *Fix:* `if (!Array.isArray(parts)) return { ok: false, result: … }` at the top of the validator.

5. **(test gap)** `multipart.test.ts:684-723` asserts the stale boundary is stripped, but does not assert that a non-multipart `Content-Type` (e.g. `application/json` accidentally set by the caller alongside a multipart body) is preserved through the strip filter. *Fix:* one extra assertion verifying the regex only matches `multipart/*`.

**Nice-to-have**
- The build closure slices a fresh `ArrayBuffer` per attempt (`mcp.ts:707-710`) — fine for the common case, but on a 401-retry of a 9 MB upload this duplicates memory. Holding the `Uint8Array` and constructing the `Blob` once outside the closure (then only re-running field substitution per attempt) would avoid the realloc. Minor.
- Comment on `mcp.ts:711` could note explicitly that `file.filename` and `file.name` are **never** substituted (the test at line 589 covers field-bytes-no-substitution but not filename-no-substitution — would close a small audit gap).

`Closes #402` is in the PR body, so merge will close the issue.

PR #449, opened 2026-05-14
feat(sidecar): x-tlsClientByUrl for TLS-fingerprinting upstreams

### LINKED ISSUE #403: feat: provider extension `x-tlsClientByUrl` to route specific URLs through `curl` for TLS-exotic upstreams
## Context

Some providers (notably those behind Cloudflare, Akamai Bot Manager, or with TLS fingerprinting such as JA3/JA4) reject requests coming from Node-native / undici clients because the TLS ClientHello doesn't match a real browser. A well-known workaround is to route through `curl` (which has its own TLS stack and is accepted).

## Problem

There's no declarative mechanism today to say "for this specific URL of provider X, go through curl; for everything else, keep using the undici client".

## Proposed solution

Commit `f74e27b4` on `otarbes/appstrate-oss:bugs-evos-oli`: new provider extension `x-tlsClientByUrl` in the manifest — a URL-pattern → client mapping (`undici` by default, `curl` as opt-in). Implementation adds `runtime-pi/sidecar/curl-runner.ts` which wraps `curl` with the internal client API.

## Real-world use cases

Providers used by agents `petitspas-to-drive` (Amisgest) and `fizz-to-drive`.

pierrecabriere:
## Summary
- Provider manifest extension `x-tlsClientByUrl: [{ pattern, client }]`
- New `runtime-pi/sidecar/curl-runner.ts` — `curlFetch(url, init): Promise<Response>` via argv-form `Bun.spawn`
- Sidecar credential-proxy dispatches per URL pattern; default remains Bun `fetch()`
- Exit-code → HTTP-status mapping (28→504, 6/7/35→502, …)

## Why
Real upstreams (Amisgest, Fizz; broader: any provider behind Cloudflare with JA3/JA4 checks) reject Node-stack TLS handshakes. Plain curl already clears the bar for many; `curl-impersonate` is the follow-up upgrade path for harder fingerprinting (Docker base swap, no API change).

## Security posture
- Opt-in per manifest pattern; no global escape hatch
- Argv-form spawn; no shell interpolation; HTTPS-only; bounded timeout
- Same SSRF + authorizedUris allowlist as the default fetch path
- CR/LF in caller headers refused (header-splitting defence in depth)
- Stderr captured separately, logged at debug; never returned in response body

## Test plan
- [x] curl-runner unit: GET/POST/redirect/timeout/dns-fail/SSL-fail/stderr-isolation (18 tests)
- [x] credential-proxy: matching URL routes to curl, non-matching uses default fetch, no-tlsClientByUrl falls through, explicit `undici` is a no-op (4 new tests on top of existing 6)
- [x] Manifest validation: invalid client + empty pattern rejected; valid extension accepted; resolver extracts the field (5 new core tests)
- [x] curl installed in sidecar Dockerfile

## Follow-up
- Swap the sidecar Docker base or add `curl-impersonate` for full JA4 coverage (no API change at this layer)

Closes #403


pierrecabriere:
Fresh-eyes production-readiness review of the curl-runner path.

## Looks good

- Argv-form `Bun.spawn(["curl", ...])` with `--` URL terminator (curl-runner.ts:167), `--proto =https`/`--proto-redir =https`, `--max-time` mirroring `OUTBOUND_TIMEOUT_MS`, CRLF rejection on header names + values (curl-runner.ts:150). All the obvious subprocess-injection vectors are closed.
- SSRF + `authorizedUris` are enforced in `executeProviderCall` (credential-proxy.ts:165-182) **before** the dispatch branch, so the curl path inherits parity for free. Credential header is materialized into `resolvedHeaders` first and then passed verbatim to `curlFetchFn` (credential-proxy.ts:267-282) — credentials cannot be silently lost on the curl branch, and the test at line 265-270 asserts this.
- Stderr cap (64 KB), stderr never reaches the response body (test line 261), and `-f` is intentionally absent so upstream 4xx/5xx round-trip with status preserved.

## Concerns

1. **No memory cap on curl stdout (security / availability).** `readAll` in curl-runner.ts:378-397 drains stdout into a single `Uint8Array` with no ceiling. An attacker-controlled or misbehaving upstream returning a 10 GB response will OOM the sidecar — the fetch path is bounded by `MAX_STREAMED_BODY_SIZE` (100 MB) in `helpers.ts`. **Fix**: mirror that cap in `readAll`; on overflow, abort the child and return a synthetic 502. This is the highest-severity finding because it's reachable from any provider that has opted into curl and serves an oversized response (or whose response is attacker-influenced).

2. **Method string is not validated against CRLF (security, request-line injection).** curl-runner.ts:108 takes `init.method ?? "GET"` straight from the caller (`ProviderCallArgs.method`, which is LLM-controlled via the MCP `provider_call` handler) and feeds it to `-X` verbatim. `curl` writes `-X` into the request line without sanitising; an LLM-supplied method containing `\r\n` could splice extra request lines/headers into the wire. The header path rejects CRLF explicitly — apply the same `/[\r\n]/` guard (or a stricter `^[A-Z]+$` allowlist of HTTP methods) at line 108. Likely-non-exploitable today since curl may itself reject, but the defence-in-depth posture in the docblock makes the omission stand out.

3. **`-L` (redirect-follow) default diverges from the fetch path (API divergence).** Bun's default for fetch is `redirect: "follow"`; the curl path defaults to **no redirects** unless the caller explicitly passes `redirect: "follow"`. `executeProviderCall` (credential-proxy.ts:275-281) does not propagate any `redirect` option into `curlInit`, so every provider routed through curl silently loses redirect-following. The docblock at curl-runner.ts:32-35 acknowledges this explicit choice but the consumer side doesn't reflect it. **Fix**: either propagate the caller's redirect intent explicitly in `executeProviderCall`, or document the divergence as expected and ensure callers know they need `-L` for Cloudflare-fronted upstreams that 302 to a worker subdomain.

4. **AbortSignal not honoured on the curl path (robustness gap).** `executeProviderCall` passes `signal: AbortSignal.timeout(OUTBOUND_TIMEOUT_MS)` on the fetch path (credential-proxy.ts:250) but the curl branch only sets `timeoutMs` — `init.signal` is never plumbed to `curl-runner`. A cancelled run can leave a curl subprocess running for up to 30s, defeating the cancel pub/sub contract documented in CLAUDE.md. **Fix**: in `curl-runner`, attach `signal.addEventListener("abort", () => child.kill())`.

5. **`extractTimeoutFromSignal` reads a private Bun field (`_timeout`) (fragility).** curl-runner.ts:362-368 reaches into `signal._timeout` — undocumented, may break across Bun versions. Today `executeProviderCall` always sets `timeoutMs` explicitly, so the fallback is dead code in practice. Either delete it or guard with a comment that calls out the Bun-version coupling.

6. **HTTP/2 status-line parser is tight (robustness).** The regex `^HTTP\/[\d.]+\s+(\d{3})(?:\s+(.*))?$` (line 299) expects either no reason phrase or one preceded by a space. Real curl `--include` output on HTTP/2 is `HTTP/2 200\r\n` (no trailing whitespace, no reason phrase) — that matches. But the test suite covers only HTTP/1.1 status lines; add an HTTP/2 fixture so a future regex tweak doesn't regress.

7. **Pattern semantics inherit `**` from `matchesAuthorizedUriSpec` — an operator pattern of `**` matches everything (defeats opt-in).** That's by design according to the comment, but a manifest-validator warning at import time (when `pattern === "**"`) would make the foot-gun visible. Not blocking — flag as future hardening.

## Nice-to-have

- The body-write block (curl-runner.ts:180-188) does not `await` the sink — Bun's `FileSink.write` is sync for buffered writes today, but the `end?.()` optional-chain hints uncertainty. Worth a one-line comment or an explicit `await sink.flush?.()` to make the contract explicit.
- `synthResponse` (line 342) includes the host in the body — minor info disclosure, but the docblock already says "never includes the upstream URL beyond the hostname" so this is consistent. Just confirming intent.

pierrecabriere:
Fermée pour relancer la décision en amont. Implémentation propre, mais la **direction stratégique** (faire vivre la TLS-evasion dans le sidecar) mérite d'être tranchée avant de committer ~1400 lignes + curl en lockstep CVE.

Les 3 questions à répondre — proxy tenant suffisant ? service d'impersonation externe ? plan pour le palier curl-impersonate ? — sont consolidées dans #458.

Le travail n'est pas perdu : si la décision en sortie de #458 valide la direction, la branche `feat/tls-client-by-url-403` est intacte et conflict-free (merge fait avec main il y a quelques minutes).

PR #448, opened 2026-05-14
refactor(sidecar): p-limit + unified env helpers

pierrecabriere:
## Summary
- Replace custom `Semaphore` (94 LoC + 148-LoC test) with `p-limit`
- Unify 3 `readPositive*Env` helpers (bytes / tokens / concurrency) into one parameterized `readPositiveIntEnv` in `helpers.ts`
- Net: ~250 LoC removed; 2 source files deleted

## Why
- `p-limit` is the standard-library answer for Node concurrency limiting (30M+ DL/wk, FIFO, wrap-the-op API → no release pairing footgun)
- Three copies of the same env-parsing algorithm collapse to one

## Behavior
Zero change. FIFO order, default fallbacks, throw-on-bad-input — all identical.

## Test plan
- [x] `bun run check`
- [x] `bun test runtime-pi/sidecar/test` (355 tests pass — mcp concurrency-cap still asserted end-to-end)
- [x] Parametrized helpers.test.ts covers all valid/invalid/ceiling cases for `readPositiveIntEnv`

Closes #444

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #450, opened 2026-05-14
feat(sidecar): p-queue migration + per-provider caps + graceful drain

### LINKED ISSUE #430: fix(sidecar): race condition in Semaphore.acquire() can exceed maxConcurrent cap
> **Scope**
> - **Layer**: sidecar `provider_call` concurrency limiter
> - **Portkey relation**: **Independent of #436.** This is a bug in our own concurrency primitive — fix regardless of the Portkey decision.
> - **Priority**: P0 (correctness bug).

Follow-up from #429 (closes #427).

## Problem

`runtime-pi/sidecar/semaphore.ts` increments `active` **after** the `await` on the waiter side, while `release()` decrements **before** resolving the waiter. Between those two microtasks, a synchronous `acquire()` from another caller can sneak under the cap.

### Reproducer

With `maxConcurrent=2`:

1. `r1, r2` held, `p3` parked, `active=2`.
2. `r1()` releases: `active=1`, queues waiter resume in microtask.
3. **Before** the microtask runs, another `acquire()` lands synchronously: sees `active=1 < 2`, sets `active=2`, returns a permit immediately.
4. Microtask runs, `p3` resumes, `active=3` → **cap exceeded**.

The E2E test in `mcp.test.ts` does not catch this because all 4 callers register at T0 via `Promise.all`. Under real MCP traffic where calls land at arbitrary times, the race can fire.

## Proposed fix

Two viable options:

### Option A — Patch the custom class

Transfer the permit to the waiter without touching the counter — the canonical pattern used by `p-limit`:

```ts
makeRelease() {
  return () => {
    if (released) return;
    released = true;
    const next = this.waiters.shift();
    if (next) next();            // transfer: don't decrement
    else this.active--;          // no waiter: free the permit
  };
}

async acquire() {
  if (this.active < this.maxConcurrent) { this.active++; return makeRelease(); }
  await new Promise<void>((r) => this.waiters.push(r));
  return makeRelease();           // active already accounted for by prior release
}
```

### Option B — Adopt [`p-queue`](https://github.com/sindresorhus/p-queue)

(MIT, ~10 KB, includes the fixed semaphore pattern + `pause()`/`drain()`/`onIdle()` needed for graceful shutdown in #435.)

Trade-off: gain a dep but fold #435's drain requirement into the same library swap. Likely preferable now that #435 is on the roadmap.

## Acceptance criteria

- [ ] New unit test in `runtime-pi/sidecar/test/semaphore.test.ts` that reproduces the race (release → sync acquire interleaved before waiter resume) and asserts `peak ≤ maxConcurrent`.
- [ ] Decide between Option A or Option B.
- [ ] Fix lands; the new race test passes.
- [ ] All 358 existing sidecar tests still pass.

## References

- PR #429 introduced the semaphore.
- [p-limit canonical pattern](https://github.com/sindresorhus/p-limit/blob/main/index.js).
- #435 (sidecar polish — drain/pause depend on this).

### LINKED ISSUE #435: polish: per-provider concurrency caps, graceful drain, queue-depth alerts
> **Scope**
> - **Layer**: sidecar `provider_call` MCP layer — concurrency caps, lifecycle, hygiene.
> - **Portkey relation**: **Independent of #436.** Portkey doesn't touch the sidecar's MCP layer or user-API `provider_call`. Do regardless of #436's decision.
> - **Priority**: P5 (polish — do after #430 lands).

Follow-up from #429 (closes #427). Lower-priority polish items consolidated.

### 1. Per-provider concurrency caps

Today `SIDECAR_PROVIDER_CALL_CONCURRENCY` is a single global cap. Different providers tolerate vastly different concurrency (Gmail handles >>10, ClickUp throttles fast). Replace the single semaphore with a `Map<providerId, Semaphore>` to take advantage of generous providers.

Schema for config:

```json
{
  "concurrency": {
    "default": 3,
    "@appstrate/gmail": 8,
    "@appstrate/clickup": 2
  }
}
```

### 2. Graceful drain on sidecar shutdown

Sidecars are ephemeral per run today, but if we ever reuse them long-lived (agent pools), `SIGTERM` should:
- Refuse new `provider_call` invocations (`503 Service Unavailable, X-Drain-Reason: shutdown`).
- Wait for in-flight permits to drain (with a max timeout, e.g. 30s).
- Exit cleanly.

Today the semaphore has no `drain()` / `pause()` — depends on #430 (fix race first, then add drain support, or migrate to `p-queue` which provides both at once).

### 3. Queue-depth alert

Emit an `appstrate.error` log event when the semaphore queue exceeds a threshold (e.g. >10 waiters) for >30s. Strong signal that the agent's fan-out logic is broken — surfaces the bug before the run times out.

### 4. Clean up artifact from #429

`runtime-pi/sidecar/mcp.ts:381` has a superfluous `{ }` block inside `providerCallInner`, kept to preserve diff indentation in #429. Remove it now that the change has shipped.

### 5. Race-condition test (post #430)

Once #430 lands, add an integration test that reproduces the race in MCP traffic terms (not just unit-level): rapid-fire `provider_call` invocations interleaved with permit releases, asserting cap never exceeded.

## Acceptance criteria

- [ ] Per-provider concurrency `Map<providerId, Semaphore>` (or `Map<providerId, pQueue>` if migrated) wired in `createApp`.
- [ ] Config schema published in sidecar README.
- [ ] `drain()` / `pause()` methods on the semaphore (or via lib swap to `p-queue`).
- [ ] SIGTERM handler in sidecar HTTP server that drains.
- [ ] Queue-depth log event with threshold + dwell time.
- [ ] `{ }` block removed from `mcp.ts:381`.
- [ ] MCP-level race-condition test added.

## Depends on

- **#430** (race fix must land first to safely add drain semantics).

## References

- #427, #429, #430.

### LINKED ISSUE #444: refactor(sidecar): replace custom Semaphore + dedupe `readPositive*Env` helpers
## Context

Architectural audit of PR #429 (rate-limit incident fixes for #427). The functional fixes are correct, but two pieces of custom code re-implement well-trodden patterns:

1. A 94-LoC `Semaphore` class (`runtime-pi/sidecar/semaphore.ts`) + 148-LoC test suite duplicates what `p-limit` (or a 5-line inline limiter) does in 30 LoC.
2. Three `readPositive*Env(name, default)` helpers spread across `helpers.ts`, `token-budget.ts`, `semaphore.ts` — same algorithm, three copies, ~70 LoC dupliqués.

Neither is broken; both are well-tested. But they add custom surface where the standard library answer is shorter and more battle-tested.

### Current state

`runtime-pi/sidecar/semaphore.ts:20-69` ships a hand-rolled FIFO semaphore with idempotent release. Used at exactly one site: `mcp.ts:366-371` wrapping `provider_call`:

```ts
const release = providerCallSemaphore ? await providerCallSemaphore.acquire() : null;
try {
  return await providerCallInner(rawArgs);
} finally {
  release?.();
}
```

### Proposed: `p-limit`

```ts
import pLimit from "p-limit";
const limit = pLimit(readPositiveIntEnv("SIDECAR_PROVIDER_CALL_CONCURRENCY", 3));
// usage:
const result = await limit(() => providerCallInner(rawArgs));
```

`p-limit` enveloppe l'opération — pas d'acquire/release paire à orchestrer dans un `finally`, donc plus difficile à mal utiliser. 30M+ DL/sem, zero deps, FIFO garanti, idempotence par fermeture de Promise.

### Proposed (alternative, zero-dep): inline limiter

```ts
function createLimiter(max: number) {
  let active = 0;
  const q: Array<() => void> = [];
  const next = () => { active--; q.shift()?.(); };
  return async <T>(fn: () => Promise<T>): Promise<T> => {
    if (active >= max) await new Promise<void>(r => q.push(r));
    active++;
    try { return await fn(); }
    finally { next(); }
  };
}
```

Same wrap-the-operation API, no dep, ~10 LoC.

### Test impact

148 LoC dans `test/semaphore.test.ts`. Decompose:
- FIFO order, capacity, parking, idempotent release (~120 LoC) → couvert à 100 % par les tests upstream de `p-limit` (ou par l'inline équivalent — la trivialité du code rend la majorité des tests redondante)
- Env-var parsing (~28 LoC) → garder, mais après Part 2 ces tests deviennent une cellule de la suite générique

### Files touched

- `runtime-pi/sidecar/semaphore.ts` — supprimé (ou raboté à 10 LoC pour `createLimiter`)
- `runtime-pi/sidecar/mcp.ts` — adapter le call site (5 LoC modifiées)
- `runtime-pi/sidecar/app.ts` — adapter le wiring (3 LoC)
- `runtime-pi/sidecar/test/semaphore.test.ts` — supprimé (ou réduit à 28 LoC)
- `runtime-pi/sidecar/test/mcp.test.ts` — adapter l'expectation de wiring (already exists, pinning concurrency end-to-end)

**Net : ~200 LoC supprimées.**

### Current state

Three near-identical helpers:

| File | Helper | Differences |
|---|---|---|
| `runtime-pi/sidecar/helpers.ts:43` | `readPositiveByteEnv(name, default, ceiling)` | has a ceiling |
| `runtime-pi/sidecar/token-budget.ts:193` | `readPositiveTokenEnv(name, default)` | error message says "tokens" |
| `runtime-pi/sidecar/semaphore.ts:86` | `readPositiveConcurrencyEnv(name, default)` | generic error message |

Tous lisent `process.env[name]`, retournent le default si `undefined`/`""`, parsent en `Number`, valident `Number.isInteger && > 0`, throwent un message paramétré. ~70 LoC dupliquées au total.

### Proposed

Un seul helper paramétré (resté dans `helpers.ts`) :

```ts
export function readPositiveIntEnv(
  name: string,
  defaultValue: number,
  opts: { unit?: string; ceiling?: number } = {},
): number {
  const raw = process.env[name];
  if (raw === undefined || raw === "") return defaultValue;
  const parsed = Number(raw);
  const unit = opts.unit ? ` (${opts.unit})` : "";
  if (!Number.isFinite(parsed) || !Number.isInteger(parsed) || parsed <= 0) {
    throw new Error(`${name} must be a positive integer${unit}, got ${JSON.stringify(raw)}.`);
  }
  if (opts.ceiling !== undefined && parsed > opts.ceiling) {
    throw new Error(
      `${name}=${parsed} exceeds the absolute ceiling of ${opts.ceiling}${unit ? ` ${opts.unit}` : ""}. ` +
        `Caps above this require code changes.`,
    );
  }
  return parsed;
}
```

Callsites:

```ts
readPositiveIntEnv("SIDECAR_MAX_REQUEST_BODY_BYTES", DEFAULT, { unit: "bytes", ceiling: ABSOLUTE_BODY_CEILING });
readPositiveIntEnv("SIDECAR_INLINE_TOOL_OUTPUT_TOKENS", DEFAULT, { unit: "tokens" });
readPositiveIntEnv("SIDECAR_PROVIDER_CALL_CONCURRENCY", DEFAULT);
```

### Test impact

Les 3 paquets de tests (un par helper) → un seul paquet paramétré sur le helper unifié, ~30 LoC au lieu de ~75.

**Net : ~40-50 LoC supprimées.**

## Total impact

- ~250 LoC supprimées sur le sidecar
- 2 fichiers source supprimés (`semaphore.ts`, `test/semaphore.test.ts`) ou réduits à des squelettes
- 0 changement comportemental — pure simplification d'API + déduplication

## Out of scope

- Le backstop adapter-error dans `finalizeRun` (`run-event-ingestion.ts:247-262`) — 16 LoC bien ciblés, pas de redondance.
- Le changement de constante `DEFAULT_RUN_OUTPUT_BUDGET_TOKENS` (200K → 100K) — déjà mergé, valide.
- Les nouveaux tests d'intégration finalize — utiles, à garder.

pierrecabriere:
## Summary

Three follow-ups from #429 (rate-limit fixes for #427), combined into one PR because they all touch `runtime-pi/sidecar/{semaphore,mcp,app,server,helpers,token-budget}.ts` and share the same test files.

- **Closes #430** (P0 race in `Semaphore.acquire()` that could exceed `maxConcurrent`). Migrated to [`p-queue`](https://github.com/sindresorhus/p-queue) (MIT, ~10 KB, zero sidecar-build runtime deps save for `eventemitter3` + `p-timeout`), which transfers permits synchronously between callers — closing the race window by construction. p-queue also gives us the `pause()`, `onIdle()`, `size`, `pending` primitives that #435's drain + queue-depth-alert need, so the same library swap closes all three issues.
- **Closes #444** (replace custom `Semaphore` + dedupe `readPositive*Env` helpers). 94-LoC `Semaphore` + 148-LoC test suite deleted. Three near-identical helpers (`readPositiveByteEnv`, `readPositiveTokenEnv`, `readPositiveConcurrencyEnv`) collapsed into one parameterized `readPositiveIntEnv(name, default, { unit?, ceiling? })` in `helpers.ts`. Old helpers removed.
- **Closes #435** (per-provider caps, graceful drain, queue-depth alert, dead `{ }` cleanup, MCP-level race test):
  - `SIDECAR_PROVIDER_CALL_CONCURRENCY` now accepts either a plain integer (legacy) or a JSON object with `default` plus per-`providerId` overrides — generous providers run higher than rate-sensitive ones.
  - SIGTERM/SIGINT handler in `server.ts` calls `drainRegistry`: `/health` flips to `503` with `X-Drain-Reason: shutdown`; new `provider_call` invocations return a structured `DRAINING` MCP tool error; in-flight calls drain up to 30 s, then process exits 0 (clean) or 1 (timeout).
  - Queue-depth watcher polls every 5 s; when any provider's parked-waiter count stays above 10 for 30 s it emits one structured `appstrate.error` log line with `{ providerId, depth, dwellMs }`. Re-armed only after depth drops below threshold.
  - Dead `{ }` block in `mcp.ts:381` (kept across #429 to preserve diff indentation) removed.
  - MCP-level race regression test in `mcp.test.ts` (50 simultaneous `tools/call` invocations against a slow mock upstream, peak ≤ cap).

## Why `p-queue` over `p-limit` or inline?

The user's spec called this out. `p-queue` was selected because:

1. Race-free permit transfer by construction (canonical pattern, same as `p-limit`).
2. Ships `pause()` / `onIdle()` / `size` / `pending` — the exact primitives #435's drain + queue-depth alert need. No additional code on top.
3. MIT, ~10 KB, zero deps inside the sidecar bundle apart from `eventemitter3` + `p-timeout`.

Pinned to exact version `9.2.0` in `runtime-pi/sidecar/package.json` (existing entries use `^`, but the user requested exact pinning).

## Diff stats

- 18 files changed, ~1656 insertions, ~504 deletions
- Net source LoC roughly neutral: deleted 94 LoC `Semaphore` + 60 LoC of duplicated env helpers; added ~290 LoC of `LimiterRegistry` with per-provider lanes, drain, queue-depth watcher.
- Test LoC +~ 500 net (limiter.test.ts, server-drain.test.ts, 4 new test groups in mcp.test.ts, 1 new test in app.test.ts).

## Test plan

- [x] `cd runtime-pi/sidecar && bun test` — **373 pass / 0 fail / 881 expect()** (up from 358).
- [x] `bun run check` — typecheck + lint + format + verify:openapi + detect:breaking + build:system-packages:check all green; only pre-existing 1 lint warning in `packages/ui/src/schema-form/file-widget.tsx`.
- [x] New regression tests:
  - `limiter.test.ts` — 18 cases covering parser edge-cases, cap correctness under 200 jittered runs, per-provider isolation, drain semantics, queue-depth alert with injected clock.
  - `server-drain.test.ts` — 4 cases pinning `drainRegistry` exit-code contract under both clean drain and timeout.
  - `mcp.test.ts` — 4 new cases (50-call MCP race for #430, JSON env format roundtrip, queue-depth alert through real MCP traffic, structured `DRAINING` tool error).
  - `app.test.ts` — 1 new case (`/health` flips to 503 + `X-Drain-Reason: shutdown` after `pause()`).
- [x] README documentation updated with new "Provider concurrency" section.


pierrecabriere:
## Fresh-eye review (independent)

### Verdict
⚠️ Ship after addressing the SIGTERM/HTTP-server gap; everything else is polish.

### What works well
- `LimiterRegistry.run()` is genuinely race-free by construction (`runtime-pi/sidecar/limiter.ts:262-266`). The whole `acquire → increment → release` window is folded inside `p-queue.add(fn)`, so the #430 release-then-resolve interleaving is unreachable through this API. The author's "test reproduces the contract, not the bug" caveat is honest and correct.
- Snapshot deliberately re-labels p-queue's footgun naming (`queue.size` → `pending`, `queue.pending` → `active`) at `limiter.ts:241-243`. Tests pin both directions (`test/limiter.test.ts:1858-1866`).
- Watcher state (`armedSince` / `alerted`) is per-providerId, cleared on below-threshold (`limiter.ts:326-329`). Re-arm semantics correctly tested at `test/limiter.test.ts:1785-1838`. Timer uses `unref()` so test imports don't hang the process (`limiter.ts:299-303`).
- Env parser is strict and fails loud at boot for every edge case enumerated: `""`, `"5"`, plain ints, JSON objects, `-1`, `1.5`, `"not json"`, `null`, `[1,2,3]`. Covered by `test/limiter.test.ts:1520-1554`. Legacy plain-int form is preserved.
- Dead `{ }` block from #429 actually removed (`mcp.ts` `providerCallInner` now starts directly at `const args = rawArgs as {…}`). Not relocated.
- Dwell/threshold/poll/clock are all injectable (`LimiterRegistryOptions` + `alertSink` + `now`); tests advance time with `now = 50/150/200` rather than `Bun.sleep` (`test/limiter.test.ts:1737-1782`).
- No `as any`, no `@ts-ignore` in `limiter.ts`. The one cast (`as T` at line 266) is well-commented.

### Required before merge (must-fix)
- **SIGTERM does not drain in-flight `/llm/*` or non-`provider_call` MCP traffic.** `server.ts:104-118` only drains the `providerCallLimiter`, then calls `process.exit(code)`. The Hono server is never stopped or quiesced. Per the platform contract (CLAUDE.md "sidecar protocol"), `/llm/*` is the load-bearing chat-completion passthrough — a SIGTERM while a streaming completion is in flight will sever the stream at the 30-second deadline (or sooner if no `provider_call` is queued, since `onIdle()` resolves immediately when the registry is empty). `/health` flipping to 503 is necessary but not sufficient. At minimum: track in-flight `/llm/*` and `/mcp` requests so `drainRegistry` waits on both before exiting, or document this explicitly as a non-goal in the README and #435.

### Suggestions (non-blocking polish)
- `mcp.ts:370-372` keys the queue by the **unvalidated** raw `rawArgs.providerId` string, ahead of the `PROVIDER_ID_RE` check inside `providerCallInner`. A misbehaving agent passing `"gmial"`, `"@foo/bar baz"`, etc. silently creates a fresh PQueue per typo. Lifetime is bounded by run duration, but consider validating `providerId` before `limiter.run()` (or sharing one `_invalid` lane).
- `parseConcurrencyConfig` accepts negative-value `perProvider` entries with the same generic message as `default` (`limiter.ts:118-125`); good. But `{"@appstrate/foo":-1}` isn't tested explicitly — only `{"default":-1}` is. Add one explicit case.
- `p-queue@9.2.0` exact-pin departs from the repo's `^` convention. The PR description says "user requested" — confirm or document the rationale in `package.json` itself.
- `DEFAULT_QUEUE_DEPTH_POLL_MS = 5000` with `DEFAULT_QUEUE_DEPTH_DWELL_MS = 30_000` gives at most 7 samples per dwell — adequate but tight. JSDoc claims `pollMs ≤ dwellMs/2`; not enforced.
- `LimiterRegistry.onIdle()` snapshots `[...this.queues.values()]` once (`limiter.ts:284`) — if a new providerId arrives between `pause()` and the snapshot, its queue is skipped. In practice `pause()` happens first so no new `run()` calls can land — safe — but worth a one-line comment to pin the invariant.

### Acceptance-criteria audit
- **#430**: ⚠️ Race fixed by construction; honest gap on "MCP-level race test" is partly covered by `test/mcp.test.ts:1902-1958` (50 simultaneous calls, peak ≤ cap, asserts peak == cap to prove saturation). Acceptable.
- **#444**: ✅ `Semaphore` + test suite deleted; three `readPositive*Env` collapsed into one `readPositiveIntEnv` with `{unit, ceiling}`. Token-budget callers migrated. Net LoC reduction realised.
- **#435 sub-items**:
  - per-provider caps: ✅ JSON + plain-int form, README documented.
  - drain: ⚠️ `provider_call` drain is clean; `/llm/*` + non-`provider_call` MCP traffic is NOT drained — see must-fix.
  - queue-depth alert: ✅ threshold + dwell + re-arm, structured `appstrate.error` log, watcher cleaned up via `dispose()`.
  - `{ }` cleanup: ✅ removed.
  - MCP race test: ✅ added in `test/mcp.test.ts`.

### Honest gaps acknowledged
The author's "the test can't reproduce the bug *via the new API* because the new API is race-free by construction" is valid and structurally unavoidable — `p-queue.add(fn)` does not expose `acquire/release`, so the #429 interleaving is unreachable. The 50-call MCP test adds end-to-end pressure. The author missed: the SIGTERM contract narrowed silently — pre-PR there was no drain at all (so no regression), but the PR description claims a graceful drain "in-flight calls drain up to 30s" that only holds for `provider_call`, not for `/llm/*` chat completions.

pierrecabriere:
Supplanté par #448 (mergé) qui résout proprement le scope strict #444 (suppression du `Semaphore` custom, dedup des `readPositive*Env`) **et** ferme #430 implicitement — `p-limit` utilise le canonical transfer pattern, donc la race P0 est résolue par construction.

Cette PR visait plus large (#430 + #444 + **#435** en un swap de lib via `p-queue`), mais #435 est marqué P5 (polish) par son auteur et dépend d'un cas d'usage qui n'existe pas encore (« if we ever reuse them long-lived (agent pools) »). Maintenir cette PR ouverte exigerait d'annuler tout #448 pour gagner uniquement le scope #435 — non rentable à ce stade (YAGNI).

#435 reste ouverte et sera traitée dans une PR dédiée le moment venu, avec une décision fraîche : rester sur `p-limit` + drain custom, ou migrer vers `p-queue`.

PR #451, opened 2026-05-14
feat(runner-pi): enable Pi auto-compaction driven by model catalog

### LINKED ISSUE #445: Pi auto-compaction: enable + drive limits from the model catalog
## Problem

Long-running agents hit a hard 400 from upstream LLM providers once their conversation grows past the model context window:

```
400 {"type":"error","error":{"type":"invalid_request_error",
"message":"prompt is too long: 226498 tokens > 200000 maximum"},
"request_id":"req_011Cb31MGg8WSPfH3SKiD7bv"}
```

Pi SDK ships an **auto-compaction** mechanism that summarizes older messages before the window saturates — but it is **disabled** in our runner:

```ts
// packages/runner-pi/src/pi-runner.ts:246-247
settingsManager: SettingsManager.inMemory({
  compaction: { enabled: false },
  retry: { … },
}),
```

It has been off since the AFPS-runtime extraction (#227) with no rationale recorded. Result: every run that exceeds `model.contextWindow` crashes with a 400 instead of compacting.

## Root cause

Two distinct issues stacked:

1. **Compaction is hard-disabled.** Even if a model has a known window, the runner does nothing about it.
2. **The runner has no per-model knob to feed compaction.** Pi's defaults are `reserveTokens: 16384` / `keepRecentTokens: 20000`. Those defaults are wrong on both extremes:
   - **Claude Sonnet 4.5 in thinking mode** (`maxTokens: 64000`): a 16k reserve is smaller than the output budget — the very first call post-compaction can underflow and return a 400 again.
   - **GPT-4.1 (1M window)** / **Gemini 2.5 (2M window)**: 20k of recent tokens is way too conservative — we'd compact on a run that has 800k+ of headroom left.

## The catalog already has what we need

`apps/api/src/data/pricing/<providerId>.json` (vendored from LiteLLM, refreshed weekly by `scripts/refresh-pricing-catalog.ts`) exposes the exact per-model values:

```json
"claude-3-7-sonnet-20250219": {
  "contextWindow": 200000,
  "maxTokens": 64000,
  "capabilities": ["text", "image", "reasoning"],
  "cost": { … }
}

"ft:gpt-4.1-2025-04-14": {
  "contextWindow": 1047576,
  "maxTokens": 32768,
  …
}
```

These are **provider-published** numbers — refresh themselves on weekly CI bumps. No Appstrate heuristic involved.

## The missing wire

`ResolvedModel.contextWindow` and `ResolvedModel.maxTokens` are populated from `org_models` rows / `SYSTEM_PROVIDER_KEYS` env entries — but the catalog fallback is **only** wired for `cost`:

```ts
// apps/api/src/services/org-models.ts:324
function resolveCost(providerId, modelId, explicit) {
  return explicit ?? lookupCatalogModel(providerId, modelId)?.cost ?? null;
}
```

When an admin (or the OAuth provider auto-seeder for codex / claude-code / openai) doesn't fill the `contextWindow` / `maxTokens` columns, those fields land as `null` in `ResolvedModel`, and the runner has no source of truth to feed Pi.

### 1. Generalize the catalog cascade

Collapse `resolveCost` into a single helper covering all three capability fields (extends audit findings #2 / #3 from the portkey-migration PR to the capability surface):

\`\`\`ts
// apps/api/src/services/org-models.ts
function resolveCapabilities(
  providerId: string,
  modelId: string,
  explicit: {
    contextWindow: number | null;
    maxTokens: number | null;
    cost: ModelCost | null;
  },
): {
  contextWindow: number | null;
  maxTokens: number | null;
  cost: ModelCost | null;
} {
  const catalog = lookupCatalogModel(providerId, modelId);
  return {
    contextWindow: explicit.contextWindow ?? catalog?.contextWindow ?? null,
    maxTokens: explicit.maxTokens ?? catalog?.maxTokens ?? null,
    cost: explicit.cost ?? catalog?.cost ?? null,
  };
}
\`\`\`

Use it in `buildSystemResolvedModel` and `buildDbResolvedModel` so the cascade applies uniformly to env-driven and DB-driven models.

### 2. Enable compaction in PiRunner, parameterized by the resolved model

\`\`\`ts
// packages/runner-pi/src/pi-runner.ts
compaction:
  process.env.MODEL_COMPACTION_ENABLED === "false"
    ? { enabled: false }
    : {
        enabled: true,
        reserveTokens: model.maxTokens ?? 16384,
        keepRecentTokens: Math.max(
          20000,
          Math.floor((model.contextWindow ?? 200_000) * 0.1),
        ),
      },
\`\`\`

Rationale for each mapping:

| Knob | Mapping | Why |
|---|---|---|
| \`reserveTokens\` | \`model.maxTokens\` (fallback 16384) | This **is** the budget reserved for the response — it MUST be ≥ \`max_tokens\`, otherwise the first call post-compaction underflows and returns the same 400. Critical for Claude thinking mode (\`maxTokens: 64000\`). |
| \`keepRecentTokens\` | \`max(20000, 10% × contextWindow)\` | Preserves the ratio across model sizes: 20k on Claude 200k, 100k on GPT-4.1 1M, 200k on Gemini 2M. Avoids both under-compaction (small windows) and over-compaction (large windows). |

The env-var opt-out (\`MODEL_COMPACTION_ENABLED=false\`) mirrors the \`MODEL_RETRY_ENABLED\` pattern already in the same file for operators stacking external middleware.

## Acceptance criteria

- [ ] \`resolveCapabilities\` (or equivalent) wired in \`buildSystemResolvedModel\` and \`buildDbResolvedModel\`; existing \`resolveCost\` removed
- [ ] PiRunner enables Pi auto-compaction by default, parameterized from \`model.maxTokens\` + \`model.contextWindow\`
- [ ] \`MODEL_COMPACTION_ENABLED=false\` opts out (cohérent avec \`MODEL_RETRY_ENABLED\`)
- [ ] Unit test in \`apps/api/test/unit/\` covering the cascade for each of the 3 fields (explicit > catalog > null)
- [ ] Unit test in \`packages/runner-pi/test/\` covering the compaction settings derivation (null model → defaults, populated model → catalog-aligned)
- [ ] Manual sanity: long run on Claude Sonnet 4.5 + GPT-4.1 1M no longer hits \`prompt is too long\`

## Out of scope

- Re-architecting Pi's compaction prompt or split-turn logic (upstream concern)
- Sidecar token-budget for tool outputs (already handled by #390 — separate axis)
- Surfacing compaction events on the dashboard (can ship later; bridge already forwards \`CompactionEntry\`)

## References

- Pi SDK compaction docs: \`node_modules/@mariozechner/pi-coding-agent/docs/compaction.md\`
- Hard-disable location: \`packages/runner-pi/src/pi-runner.ts:246-247\`
- Cost-only fallback location: \`apps/api/src/services/org-models.ts:324-330\`
- Catalog source: \`apps/api/src/data/pricing/*.json\` (refresher: \`scripts/refresh-pricing-catalog.ts\`)

pierrecabriere:
## Summary

Closes #445.

- Generalize the catalog cascade in `apps/api/src/services/org-models.ts`: new `resolveCapabilities(providerId, modelId, explicit)` helper covers `contextWindow` + `maxTokens` + `cost` uniformly with the `explicit > LiteLLM-vendored catalog > null` precedence. Wired into both `buildSystemResolvedModel` (env-driven `SYSTEM_PROVIDER_KEYS`) and `buildDbResolvedModel` (DB-driven `org_models`) so they cannot drift. Honors `catalogProviderId` for OAuth wrappers (codex → openai, claude-code → anthropic).
- Enable Pi auto-compaction in `packages/runner-pi/src/pi-runner.ts`, parameterized from the resolved model via a new pure `derivePiCompactionSettings(model, env)`:
  - `reserveTokens = model.maxTokens ?? 16384` — must be ≥ the response budget or the first call post-compaction underflows and the upstream `prompt is too long` 400 reappears. Critical for Claude Sonnet thinking mode (`maxTokens: 64000`).
  - `keepRecentTokens = max(20000, floor(0.1 × contextWindow))` — preserves the ratio: 20k on 200k window, 100k on 1M, 200k on 2M. The 20k floor protects small-window models from over-compaction.
- `MODEL_COMPACTION_ENABLED=false` opt-out for operators stacking external middleware (mirrors `MODEL_RETRY_ENABLED`).

## Test plan

- [x] `bun test packages/runner-pi/test/pi-runner-compaction.test.ts` — 13 tests covering `reserveTokens` derivation, `keepRecentTokens` ratio + floor, env opt-out, full result shape
- [x] `bun test packages/runner-pi/test` — 154 tests pass, no regression
- [x] `bun test apps/api/test/unit/services/org-models-resolve-capabilities.test.ts` — 13 tests covering per-field cascade (explicit / catalog / miss), `catalogProviderId` alias path, field independence
- [x] `bun test apps/api/test/unit` — 804 tests pass
- [x] `bun test apps/api/test/integration/services/org-models-pricing-catalog.test.ts` — 4 tests pass (existing coverage of the cascade through `loadModel` end-to-end)
- [x] `bun run check` — typecheck + lint + format + verify:openapi + detect:breaking all green
- [ ] Manual sanity post-merge: long run on Claude Sonnet 4.5 + GPT-4.1 1M no longer hits `prompt is too long`

### DIFF
--- a/apps/api/src/services/org-models.ts
+++ b/apps/api/src/services/org-models.ts
@@ -397,9 +397,46 @@ function resolveCatalogDefaults(providerId: string, modelId: string): CatalogDef
   };
 }
 
+/**
+ * Focused cascade for the three numeric/cost capability fields a runner
+ * needs to size compaction, budget responses, and price LLM calls. Keeps
+ * the env-driven (`SYSTEM_PROVIDER_KEYS`) and DB-driven (`org_models`)
+ * builders aligned on the exact same `explicit > catalog > null` chain.
+ *
+ * Honors `catalogProviderId` via {@link resolveCatalogDefaults} so OAuth
+ * wrappers (codex → openai, claude-code → anthropic) hit the right
+ * pricing file. Each field resolves independently — setting one explicit
+ * override does not shadow another field's catalog fallback.
+ */
+export function resolveCapabilities(
+  providerId: string,
+  modelId: string,
+  explicit: {
+    contextWindow: number | null;
+    maxTokens: number | null;
+    cost: ModelCost | null;
+  },
+): {
+  contextWindow: number | null;
+  maxTokens: number | null;
+  cost: ModelCost | null;
+} {
+  const catalog = resolveCatalogDefaults(providerId, modelId);
+  return {
+    contextWindow: explicit.contextWindow ?? catalog.contextWindow ?? null,
+    maxTokens: explicit.maxTokens ?? catalog.maxTokens ?? null,
+    cost: explicit.cost ?? catalog.cost ?? null,
+  };
+}
+
 /** Build a `ResolvedModel` from a system `ModelDefinition` (env-driven). */
 function buildSystemResolvedModel(def: ModelDefinition): ResolvedModel {
   const defaults = resolveCatalogDefaults(def.providerId, def.modelId);
+  const capabilities = resolveCapabilities(def.providerId, def.modelId, {
+    contextWindow: def.contextWindow ?? null,
+    maxTokens: def.maxTokens ?? null,
+    cost: def.cost ?? null,
+  });
   return {
     providerId: def.providerId,
     apiShape: def.apiShape,
@@ -408,10 +445,10 @@ function buildSystemResolvedModel(def: ModelDefinition): ResolvedModel {
     apiKey: def.apiKey,
     label: def.label ?? defaults.label ?? def.modelId,
     input: def.input ?? defaults.input ?? null,
-    contextWindow: def.contextWindow ?? defaults.contextWindow ?? null,
-    maxTokens: def.maxTokens ?? defaults.maxTokens ?? null,
+    contextWindow: capabilities.contextWindow,
+    maxTokens: capabilities.maxTokens,
     reasoning: def.reasoning ?? defaults.reasoning ?? null,
-    cost: def.cost ?? defaults.cost ?? null,
+    cost: capabilities.cost,
     isSystemModel: true,
   };
 }
@@ -422,11 +459,17 @@ function buildSystemResolvedModel(def: ModelDefinition): ResolvedModel {
  * service resolves them from the registry by `providerId` (with the per-row
  * `baseUrlOverride` honored when `baseUrlOverridable: true`). Every catalog-
  * derivable column on `org_models` is an optional override that defers to
- * {@link resolveCatalogDefaults} on null — so a weekly catalog refresh
- * propagates to existing rows.
+ * the catalog on null — so a weekly catalog refresh propagates to existing
+ * rows. Numeric/cost capability fields go through {@link resolveCapabilities}
+ * so the env-driven and DB-driven paths share the exact same cascade.
  */
 function buildDbResolvedModel(row: DbOrgModelRow, creds: DbModelCredentials): ResolvedModel {
   const defaults = resolveCatalogDefaults(creds.providerId, row.modelId);
+  const capabilities = resolveCapabilities(creds.providerId, row.modelId, {
+    contextWindow: row.contextWindow,
+    maxTokens: row.maxTokens,
+    cost: (row.cost as ModelCost | null) ?? null,
+  });
   return {
     providerId: creds.providerId,
     apiShape: creds.apiShape,
@@ -435,10 +478,10 @@ function buildDbResolvedModel(row: DbOrgModelRow, creds: DbModelCredentials): Re
     apiKey: creds.apiKey,
     label: row.label ?? defaults.label ?? row.modelId,
     input: (row.input as string[] | null) ?? defaults.input ?? null,
-    contextWindow: row.contextWindow ?? defaults.contextWindow ?? null,
-    maxTokens: row.maxTokens ?? defaults.maxTokens ?? null,
+    contextWindow: capabilities.contextWindow,
+    maxTokens: capabilities.maxTokens,
     reasoning: row.reasoning ?? defaults.reasoning ?? null,
-    cost: (row.cost as ModelCost | null) ?? defaults.cost ?? null,
+    cost: capabilities.cost,
     isSystemModel: false,
     accountId: creds.accountId,
     credentialId: row.credentialId,
--- a/apps/api/test/unit/services/org-models-resolve-capabilities.test.ts
+++ b/apps/api/test/unit/services/org-models-resolve-capabilities.test.ts
@@ -1,0 +1,213 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Unit tests for `resolveCapabilities` — the focused 3-field cascade
+ * (`explicit > catalog > null`) that feeds both the env-driven
+ * (`SYSTEM_PROVIDER_KEYS`) and DB-driven (`org_models`) builders.
+ * Closes appstrate#445.
+ *
+ * Pure-function tests — no DB, no HTTP. The catalog is the on-disk
+ * LiteLLM-vendored snapshot in `apps/api/src/data/pricing/*.json` so
+ * `gpt-4o` (openai) is a stable fixture, and a hand-registered
+ * test provider validates the `catalogProviderId` alias path.
+ */
+
+import { describe, it, expect, beforeAll, afterAll } from "bun:test";
+import { resolveCapabilities } from "../../../src/services/org-models.ts";
+import {
+  registerModelProvider,
+  resetModelProviders,
+} from "../../../src/services/model-providers/registry.ts";
+import { seedTestModelProviders } from "../../helpers/model-providers.ts";
+
+// `gpt-4o` ships in the vendored snapshot at $2.5 / $10 per M, 128k
+// context, 16k max tokens. Asserting the canonical numbers keeps the
+// test honest against a stale catalog (a regenerate that drops the
+// entry fails loudly here, not in production).
+const GPT_4O = {
+  contextWindow: 128_000,
+  maxTokensMin: 4_096, // Sanity floor — actual is 16_384 on the current snapshot.
+  costInput: 2.5,
+  costOutput: 10,
+};
+
+describe("resolveCapabilities — cascade per field", () => {
+  beforeAll(() => {
+    // The function reads from the model-provider registry via
+    // `resolveCatalogDefaults`. Seed it to the canonical test baseline
+    // (core-providers + synthetic test-oauth) so `openai` resolves.
+    seedTestModelProviders();
+  });
+  afterAll(() => {
+    // Re-seed for the next file in the same `bun test` process.
+    seedTestModelProviders();
+  });
+
+  describe("contextWindow", () => {
+    it("returns explicit when set (ignores catalog)", () => {
+      const res = resolveCapabilities("openai", "gpt-4o", {
+        contextWindow: 999,
+        maxTokens: null,
+        cost: null,
+      });
+      expect(res.contextWindow).toBe(999);
+    });
+
+    it("falls through to catalog when explicit is null", () => {
+      const res = resolveCapabilities("openai", "gpt-4o", {
+        contextWindow: null,
+        maxTokens: null,
+        cost: null,
+      });
+      expect(res.contextWindow).toBe(GPT_4O.contextWindow);
+    });
+
+    it("returns null on catalog miss (unknown model)", () => {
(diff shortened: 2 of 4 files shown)


pierrecabriere:
## Independent review — fresh eyes on #451

### 1. Verdict

⚠️ **Ready with caveats.** No blocking bugs. The core fix is correct: `derivePiCompactionSettings` is a clean pure-function extraction, `resolveCapabilities` correctly subsumes the catalog cascade for the three runner-critical fields, and the tests cover the actual decision matrix that mattered for #445 (Claude thinking @ 64k, 1M/2M windows, the 20k floor). The nits below are mostly defensive-coding gaps and a constant divergence that's worth fixing in a follow-up.

### 2. Blocking issues

None.

### 3. Non-blocking concerns

- **Constant divergence between the runner and the entrypoint env layer.** `packages/runner-pi/src/pi-runner.ts:123` sets `DEFAULT_CONTEXT_WINDOW = 200_000`, while `runtime-pi/env.ts:74` uses `DEFAULT_CONTEXT_WINDOW = 128_000` and `DEFAULT_MAX_TOKENS = 16_384` as the fallback when `MODEL_CONTEXT_WINDOW` / `MODEL_MAX_TOKENS` are missing on the container env. In the **Docker / runtime-pi** path the entrypoint always materialises `model.contextWindow` / `model.maxTokens` as numbers before they reach `derivePiCompactionSettings`, so the new 200k/16384 fallbacks are dead branches at the production call site. They're only ever exercised by the unit tests and the CLI path. That's fine — but it means we now have two different "what's our default context window" answers in the same package (128k for the env layer, 200k for the runner). Worth picking one and centralising. Suggest extracting both into a shared constants file at minimum, ideally aligning them.

- **`reserveTokens >= contextWindow` foot-gun is not guarded.** Pi SDK trigger is `contextTokens > contextWindow - reserveTokens` (verified in `compaction.js:152`). If an operator misconfigures a model so `maxTokens >= contextWindow` (small-window model + thinking-mode budget pasted in), compaction triggers on **every turn**, hammering the LLM with summarization passes. Defensive ceiling like `reserveTokens = min(model.maxTokens ?? 16384, Math.floor(contextWindow * 0.9))` would be cheap insurance. Not a regression — it's a new failure mode introduced by enabling compaction at all. Worth a follow-up ticket.

- **Catalog-null silent downgrade.** Per `apps/api/src/services/org-models.ts:411`: if a row has `contextWindow: null` AND the catalog has no entry (e.g. fine-tuned / unmapped model), `resolveCapabilities` returns `null`. Downstream, `container-env.ts:126` omits `MODEL_CONTEXT_WINDOW`, and the runtime-pi `env.ts:252` defaults to 128k. A user running a real 1M-context model with sparse catalog data silently gets sized as a 128k model. This isn't introduced by this PR — the chain existed before — but enabling compaction by default makes it newly observable. Worth a follow-up: structured logging when `ResolvedModel.contextWindow === null` flows into a runner, ideally a `logger.warn` at the resolve boundary so operators see "model X has no contextWindow — using 128k fallback for compaction sizing".

- **Behavior change not surfaced in `.env.example`.** `MODEL_COMPACTION_ENABLED` is not documented in `.env.example` or in the `CLAUDE.md` env table. The existing `MODEL_RETRY_ENABLED` is also undocumented there, so there's precedent, but compaction-on-by-default is a behavior change for existing self-hosted users — at minimum a release-notes line is warranted. Cost tracking is also impacted: compaction makes its own LLM call (per Pi docs `generateSummary`), so `runs.cost` will trend higher on long runs even when nothing else changes. Operators on tight budgets may be surprised.

- **Test gap: end-to-end wiring not asserted.** Both test files exercise the pure helpers in isolation, which is exactly what they're designed to do, and that's good. But there is **no integration test** asserting that `executeSession` actually wires the helper's output into `SettingsManager.inMemory({ compaction: ... })`. A `ScriptedPiRunner`-style test that asserts a model with `maxTokens: 64_000` results in a `SettingsManager.inMemory` call with `reserveTokens: 64_000` would close the loop. Without it, a future refactor could silently re-disable compaction by inlining a `{ enabled: false }` literal and the unit tests would still pass.

- **Test gap: predicate strictness on `MODEL_COMPACTION_ENABLED`.** `packages/runner-pi/test/pi-runner-compaction.test.ts:81` asserts that `"true"` keeps it on, but doesn't cover `"FALSE"` (uppercase), `" false "` (whitespace), or `"0"`. The strict `=== "false"` matches the `MODEL_RETRY_ENABLED` predicate so this is intentional, but operators copying the value from a logger or YAML may hit this silently. A single `it("only exact-lowercase 'false' opts out")` test pinning the contract would prevent a well-intentioned future PR from "fixing" it.

- **Cosmetic: `buildSystemResolvedModel` (org-models.ts:402) still calls `resolveCatalogDefaults` directly to read `label` / `input` / `reasoning`, then calls `resolveCapabilities` for the numeric trio.** Two catalog lookups per model resolution (the new helper re-calls `resolveCatalogDefaults` internally). Same in `buildDbResolvedModel` at line 467. Not measurably hot on real workloads — both are O(1) hash lookups — but it's the kind of nit that gets flagged in a hot-path audit. Could be cleaned up by either (a) folding all six fields into a single helper, or (b) accepting a pre-computed `defaults` parameter in `resolveCapabilities`. Author called this out in the PR description as a deliberate layering choice; reasonable, but the duplicate lookup deserves a comment in the code explaining why two passes is preferable.

- **The original issue's acceptance criterion "existing `resolveCost` removed" is technically vacuous** — `resolveCost` was already absent from `main` (it's not in the current `apps/api/src/services/org-models.ts`). What existed was `resolveCatalogDefaults`, a six-field helper. The PR adds `resolveCapabilities` on top rather than removing or folding the existing helper. The PR description doesn't call this deviation out explicitly. Not a problem in itself, but the issue's checkbox is misleading — worth noting in the PR description that "no `resolveCost` removal was needed; the existing six-field `resolveCatalogDefaults` continues to serve `label` / `input` / `reasoning` and `resolveCapabilities` layers on top for the numeric trio."

### 4. Things done well

- **Pure-function extraction (`derivePiCompactionSettings`).** Lifts the only piece of logic worth unit-testing out of the unmockable `executeSession` boot path. The `enabled: false | { enabled: true; ...}` discriminated-union return type is clean.
- **Env-opt-out predicate matches `MODEL_RETRY_ENABLED` exactly** (`=== "false"`). Consistency with an existing operator escape hatch is the right call — no new surprises.
- **`catalogProviderId` alias path is explicitly tested** via the synthetic `test-openai-wrapper-445` registration. That's the right way to exercise the OAuth-wrapper case without coupling unit tests to `module-codex` load order.
- **Tests pin the canonical fixtures** (`gpt-4o`: 128k context, $2.5/$10) with a sanity-floor comment explaining that a catalog regenerate dropping the entry must fail loudly. Future-proofing against silent test rot.
- **`reserveTokens = model.maxTokens` mapping is the right call** and the doc-comment in the runner explains why ("MUST be ≥ `max_tokens` or the first call post-compaction underflows") — this is the actual root cause of #445.

### 5. Suggested follow-up tickets

1. **Unify the runner-pi default context-window constants.** `runtime-pi/env.ts` (128k) vs `packages/runner-pi/src/pi-runner.ts` (200k) should agree. Pick one and re-export, or centralise in a shared constants file.
2. **Guard against `reserveTokens >= contextWindow` misconfiguration.** Cap `reserveTokens` at `Math.floor(contextWindow * 0.9)` to prevent the every-turn-compacts foot-gun. Log a warning when the cap fires.
3. **Surface a `logger.warn` at `resolveCapabilities` when the cascade returns `contextWindow: null`.** Operators running unmapped fine-tuned models silently get the 128k fallback in the container env; observability would catch this without changing behavior.
4. **Document `MODEL_COMPACTION_ENABLED` and `MODEL_RETRY_ENABLED` in `.env.example` and `CLAUDE.md` env table.** Both are operator-facing knobs and neither is documented.
5. **Add an integration test for the wiring** — `derivePiCompactionSettings` is asserted, but its return value reaching `SettingsManager.inMemory({ compaction })` is not. A `ScriptedPiRunner`-style assertion against the settings actually installed on the Pi SDK session would lock the contract end-to-end.
6. **Surface compaction events in `runs.cost` attribution and/or the UI.** The PR description notes compaction events are forwarded by the bridge as `CompactionEntry`, but compaction calls the LLM itself and contributes to `runs.cost` without any operator-visible signal that summarization is happening. A small UI indicator or a structured log line on each `CompactionEntry` would close the loop.

— Reviewed without checking out the PR; read from `origin/feat/445-pi-auto-compaction` and cross-referenced the Pi SDK 0.70.6 `compaction.js` and `types.d.ts`.

PR #452, opened 2026-05-14
feat: support password authMode (OAuth2 ROPC) for legacy providers

### LINKED ISSUE #400: feat: support `password` authMode for providers using OAuth2 Resource Owner Password Credentials grant
## Context

Appstrate supports `apiKey`, `oauth1`, `oauth2` authModes today. Several real-world providers (e.g. Amisgest, Fizz) only expose the *Resource Owner Password Credentials* grant (RFC 6749 §4.3): user provides username/password, the sidecar bootstraps an access_token + refresh_token and injects it into outbound requests.

## Problem

There is no mechanism in the AFPS spec / connect / sidecar to declare this mode. Current workaround: do everything inside a custom tool, which duplicates refresh logic across agents and exposes credentials to the user-runtime layer.

## Proposed solution

8 commits ready on `otarbes/appstrate-oss:bugs-evos-oli`:

1. `923fc201` — afps-spec: add `"password"` to `authModeEnum` + `passwordGrant` types
2. `31dbfaaf` — manifest: shape aligned with oauth1/oauth2 patterns
3. `f8943bf2` — connect: propagate `passwordGrant` through credential resolution
4. `576dbafa` — sidecar: new `password-grant.ts` module (bootstrap + refresh + injection)
5. `c5333a1e` — sidecar: wire authMode in `credential-proxy`
6. `021b9493` — Dockerfile: COPY `password-grant.ts`
7. `c6c46000` — refactor: placeholders aligned on `{{name}}` flat convention
8. `eca7334e` — sidecar: capture `Set-Cookie` at every redirect hop (otherwise Amisgest loses session)

## Validation

Validated locally with Amisgest (agent `petitspas-to-drive`) and Fizz providers.

## Rebase notes

Expected conflicts with #379 (connection vs credentials disambiguation), #380 (Profile → UserProfile), #382 (profileId unification). Will rebase on main and split into 2 PRs (spec/connect, then sidecar runtime) to ease review.

pierrecabriere:
## Summary
- New `password` authMode (RFC 6749 §4.3) for providers that only expose ROPC
- `packages/connect/src/password-grant.ts` — bootstrap + refresh + re-bootstrap, modeled on `oauth.ts`
- Provider manifest extension parallel to `oauth2`/`oauth1` (`definition.password.tokenUrl`, optional client + scope + auth-method + content-type knobs)
- Frontend connection form renders for `password` mode (same modal as `custom`, default username + password schema fallback)

## Why
Real-world providers (Amisgest, Fizz, ...) ship only the ROPC grant. Today, agents have to embed refresh logic per-tool, duplicating code and leaking credentials into the runtime layer. With this change, the platform handles the grant once and the sidecar / credential-proxy inject bearer tokens just like oauth2.

## Security posture
ROPC is **deprecated in OAuth 2.1** and discouraged. It is supported here as an explicit, manifest-declared opt-in for upstreams that leave us no choice. Credentials remain encrypted server-side (v1 envelope, same encryption key as oauth2/oauth1) and never reach the agent container — `buildSidecarCredentials` scrubs raw `username`/`password` from the resolved map before injection.

## Refresh / re-bootstrap semantics
`getCredentials()` lazily bootstraps a token on first use (no synchronous upstream call at connect time). `forceRefreshCredentials()` tries `grant_type=refresh_token` first; on `invalid_grant` it falls back to a fresh ROPC exchange using the stored username/password. Transient failures (5xx, network) surface unchanged — no auto-re-bootstrap to avoid upstream account-lockout from repeated wrong-credential attempts during outages.

## Test plan
- [x] Unit: ROPC bootstrap (form-urlencoded + JSON), `client_secret_post` vs `client_secret_basic`, public client, scope, network / non-JSON / `invalid_grant` / `invalid_client` paths (`packages/connect/test/password-grant.test.ts` — 13 tests)
- [x] Integration: provider seed → connection seed → `proxyCall` bootstrap → 401 refresh → refresh-failure re-bootstrap (`apps/api/test/integration/services/password-grant.test.ts` — 3 tests, extends the existing OAuth mock server with per-grant-type overrides)
- [x] `bun run check` — turbo (tsc + eslint + prettier) + `verify:openapi` (5/5 checks pass)
- [x] `bun run test:unit` — full API unit surface (791 tests pass)
- [x] Pre-existing `credential-proxy-refresh` + `connection-manager` integration suites (no regression)
- [ ] Manual: connect to a real ROPC provider (operator-side validation, not in CI)

## Scope cuts
- Provider editor UI exposes `password` in the auth-mode picker and credential-schema flow, but does **not** render a dedicated config section for `passwordTokenUrl` / scope / auth-method — operators with password providers edit the JSON tab directly. A dedicated editor section is a follow-up.
- The `basic` authMode is still routed through the OAuth path in two of the dashboard connection cards (a pre-existing bug). This PR adds `password` as a separate branch alongside `basic` without fixing that path.

## Follow-up
- `@afps-spec/schema` v1 enum addition for `password` (separate PR in the afps-spec repo). The platform extends the upstream `authModeEnum` locally with `z.enum([...afpsAuthModeEnum.options, \"password\"])`; ZIP imports of password providers will fail until the spec releases, but dashboard creation works.

Closes #400

### DIFF
--- a/apps/api/src/openapi/schemas.ts
+++ b/apps/api/src/openapi/schemas.ts
@@ -14,7 +14,7 @@ const ORG_ROLES = [...orgRoleEnum.enumValues];
 
 const providerAuthModeEnum = {
   type: "string",
-  enum: ["oauth2", "oauth1", "api_key", "basic", "custom"],
+  enum: ["oauth2", "oauth1", "api_key", "basic", "custom", "password"],
 } as const;
 
 const providerTokenContentTypeProperty = {
@@ -90,6 +90,15 @@ const providerInputSharedProperties = {
   docsUrl: { type: "string" },
   authorizedUris: { type: "array", items: { type: "string" } },
   allowAllUris: { type: "boolean" },
+  passwordTokenUrl: {
+    type: "string",
+    description:
+      "OAuth2 Resource Owner Password Credentials (RFC 6749 §4.3) token endpoint. Required when authMode === 'password'.",
+  },
+  passwordScope: {
+    type: "string",
+    description: "Optional space-separated scope string sent in the ROPC token request body.",
+  },
 } as const;
 
 /**
--- a/apps/api/src/routes/connections.ts
+++ b/apps/api/src/routes/connections.ts
@@ -174,7 +174,8 @@ export function createConnectionsRouter() {
 
         // Resolve the auth mode from the provider
         const authMode = await getProviderAuthMode(scope, provider);
-        const mode = authMode === "basic" ? "basic" : "custom";
+        const mode: "basic" | "custom" | "password" =
+          authMode === "basic" ? "basic" : authMode === "password" ? "password" : "custom";
 
         const connectionProfileId = data.connectionProfileId ?? (await getDefaultProfileId(actor));
 
--- a/apps/api/src/routes/internal.ts
+++ b/apps/api/src/routes/internal.ts
@@ -26,6 +26,7 @@ import {
   getProviderCredentialId,
   getConnection,
   RefreshError,
+  PasswordGrantError,
 } from "@appstrate/connect";
 import { resolveManifestProviders } from "../lib/manifest-utils.ts";
 import { unauthorized, forbidden, notFound, invalidRequest, internalError } from "../lib/errors.ts";
@@ -387,7 +388,15 @@ export function createInternalRouter() {
       // other OAuth error codes) must not flag — the credential might still
       // be valid and the initial 401 that triggered this refresh may have
       // come from a malformed agent request, not a dead token.
-      if (err instanceof RefreshError && err.kind === "revoked") {
+      //
+      // Same rule for password-grant (ROPC) connections: a `PasswordGrantError`
+      // with kind `"revoked"` means either the stored username/password was
+      // rejected or both the refresh_token AND the re-bootstrap failed. Either
+      // way the connection is unusable and must be flagged.
+      const isRevoked =
+        (err instanceof RefreshError && err.kind === "revoked") ||
+        (err instanceof PasswordGrantError && err.kind === "revoked");
+      if (isRevoked) {
         await db
           .update(userProviderConnections)
           .set({ needsReconnection: true, updatedAt: new Date() })
@@ -404,18 +413,20 @@ export function createInternalRouter() {
           runId,
           providerId,
           connectionProfileId,
-          status: err.status,
+          status: err instanceof RefreshError ? err.status : (err as PasswordGrantError).status,
         });
 
         throw unauthorized(`Token refresh failed for provider '${providerId}': credential revoked`);
       }
 
+      const classifiedErr =
+        err instanceof RefreshError || err instanceof PasswordGrantError ? err : undefined;
       logger.warn("Transient refresh failure, connection left unchanged", {
         runId,
         providerId,
         connectionProfileId,
-        kind: err instanceof RefreshError ? err.kind : "unknown",
-        status: err instanceof RefreshError ? err.status : undefined,
+        kind: classifiedErr?.kind ?? "unknown",
+        status: classifiedErr?.status,
         error: getErrorMessage(err),
       });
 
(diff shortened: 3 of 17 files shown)


pierrecabriere:
Reviewed packages/connect/src/password-grant.ts, the credentials.ts changes, both test files, the route schemas, and the frontend wiring. The grant module is well-shaped (modeled on oauth.ts, RFC §2.3.1 URL-encoding correct in buildTokenHeaders, error classification reused from token-utils, no password in any log statement). The integration test exercises the real bootstrap → 401-refresh → invalid_grant → re-bootstrap cycle through `proxyCall`.

**Looks good**
- RFC 6749 §2.3.1 `client_secret_basic` encoding is correct (`encodeURIComponent` before base64) and unit-tested.
- `buildSidecarCredentials` scrubs raw username/password before the sidecar sees them — agent runtime can never read them.
- Refresh-on-transient deliberately does NOT re-bootstrap (avoids upstream account lockout on a network blip); the comment in credentials.ts:222 documents this clearly.

**Concerns**
1. **[correctness] `PasswordGrantError` is never caught outside the connect package.** `apps/api/src/routes/internal.ts:389` flips `needsReconnection=true` only on `err instanceof RefreshError`. A wrong-password bootstrap (or a dead `refresh_token` that re-bootstrap can't recover) throws `PasswordGrantError(kind: "revoked")` and propagates as a 500 — the connection is never flagged, the dashboard keeps showing "connected", and the next run repeats the call. Fix: add an `err instanceof PasswordGrantError && err.kind === "revoked"` branch alongside the existing `RefreshError` handler (and any other site that catches `RefreshError`).
2. **[robustness] No concurrency lock on bootstrap or re-bootstrap.** `packages/connect/src/token-refresh.ts:17` keeps `inflightRefreshes` per connection so parallel oauth2 refreshes coalesce. `bootstrapPasswordGrantConnection` and `refreshOrBootstrapPasswordGrant` have no equivalent — two parallel `provider_call`s on a fresh password connection POST the same username/password twice and race on the UPDATE. Worse, some IdPs lock the account after N rapid bad-password requests. Fix: reuse the same `Map<connectionId, Promise<…>>` pattern keyed on `connection.id`.
3. **[security] `passwordTokenUrl` is not validated as a URL.** `apps/api/src/routes/providers.ts:194` declares `passwordTokenUrl: z.string().optional()` — no `z.url()`, no `https://` check. Combined with no SSRF guard on the bootstrap fetch, an org admin can point ROPC at `http://169.254.169.254/...` and exfiltrate creds via headers. Fix: `z.url()` at minimum; ideally route the token POST through the existing SSRF guard.
4. **[DX] No user-facing warning that ROPC is deprecated.** The PR body mentions OAuth 2.1 deprecation but neither the dashboard auth-mode picker (`provider-editor-inner.tsx:301`) nor the connect modal carries a warning. Operators picking "Password (OAuth2 ROPC)" deserve a hint that this should be a last resort. Fix: tooltip / helper text on the SelectItem and connect modal.
5. **[DX] AFPS schema mismatch has no in-code warning.** `packages/core/src/validation.ts:134` extends the enum locally; the PR body acknowledges ZIP imports of `authMode: "password"` providers will fail at the AFPS validator layer once the spec catches up. A short TODO referencing the follow-up afps-spec PR on the local extension would prevent surprise.

**Nice-to-have**
- `password-grant.ts:113` skips Basic-auth header construction when `useBasicAuth` is true but `buildTokenHeaders` is still called with `clientSecret ?? ""` — for a public client that explicitly sets `tokenAuthMethod: "client_secret_basic"` (rare but legal), this emits `Basic base64(clientId:)`. Either guard on `clientSecret` presence in `buildTokenHeaders` or document that `client_secret_basic` requires a secret.
- The integration test's `fakeFetch` pattern is clean. Consider a 4th case: bootstrap with `invalid_grant` (wrong stored password) → expect `needsReconnection=true` and a 4xx surfaced upstream. That test would have surfaced concern #1.

pierrecabriere:
Closing in favor of a different direction — see #457 for the rationale.

Short version: ROPC is deprecated in OAuth 2.1 and adding it as a first-class platform `authMode` would institutionalize storing user passwords server-side, which is a material shift in Appstrate's security posture. The need is real (Amisgest, Fizz), but the better path is per-provider external modules (à la `@appstrate/module-claude-code`) + a documented `custom` + manual-bearer fallback for one-offs, so the platform itself never holds the password.

PR #453, opened 2026-05-14
refactor(model-registry): consolidate model-metadata overlap

### LINKED ISSUE #443: refactor(model-registry): consolidate `ModelDefinition` / `CatalogModel` overlap
## Context

Found during PR #439 (Portkey strip) architectural audit. Larger candidate, opened separately because it deserves its own design pass — kept out of the Portkey-strip PR on purpose.

Three model-metadata shapes coexist with significant field overlap (`label`, `contextWindow`, `maxTokens`, `cost`, capability flags):

| Shape | Where | Role |
|---|---|---|
| `CatalogModel` | `apps/api/src/services/pricing-catalog.ts` | Vendored LiteLLM entry (read-only, file-backed) |
| `ModelDefinition` | `apps/api/src/services/model-registry.ts:21` | System-registry entry (env `SYSTEM_PROVIDER_KEYS` + catalog merge) |
| `OrgModelInfo` | `@appstrate/shared-types` | Wire shape the API returns to the React client |

`ModelDefinition` is built by taking `SYSTEM_PROVIDER_KEYS` config and re-projecting catalog metadata on top of it. `OrgModelInfo` is then projected from `ModelDefinition` in `org-models.ts:81-97`. Changing the cost shape (e.g. adding a new pricing dimension) means editing all three definitions and at least two projection functions in lockstep.

## Why this is real

- All three carry the same `ModelCost` instance verbatim — no transformation, no semantic difference.
- All three carry `contextWindow`/`maxTokens` verbatim.
- All three carry a capability surface, but spelled differently: `capabilities: string[]` (catalog) vs `input: string[] | null` + `reasoning: boolean | null` (registry/wire). This is actually a divergence worth normalizing.

## Why it's not trivial

The three layers do serve genuinely different concerns: the catalog is upstream-vendored and read-only, the registry holds operator config + the apiKey/credentialId (which never leaves the server), and the wire shape strips secrets before returning to the client. So we can't just collapse to one type — but we can:

1. Normalize the capability representation (one shape, used everywhere).
2. Define a single \"model metadata\" struct that the three carriers compose, rather than re-declaring its fields.
3. Replace the `mapRow` projection at `org-models.ts:81-97` with a `pickClientFields` helper.

## Out of scope

Renaming user-visible config keys (e.g. \`SYSTEM_PROVIDER_KEYS\` env shape is part of the self-hoster contract).

## Estimated impact

~30-50 LoC removed, mostly in the projection layer. Bigger benefit: changing the cost shape becomes a one-file edit, and the capability surface stops drifting between picker UX and proxy resolution.

pierrecabriere:
## Summary

- New canonical `ModelMetadata` struct in `@appstrate/shared-types` carrying the six shared fields (`label`, `contextWindow`, `maxTokens`, `input`, `reasoning`, `cost`)
- `ModelDefinition` (system-registry) and `OrgModelInfo` (wire shape) compose it via `extends` instead of re-declaring each field independently
- `CatalogModelEntry` retains its `capabilities: string[]` representation (vendored JSON files unchanged); `resolveCatalogDefaults()` continues to project into the canonical `{ input, reasoning }` surface at load time
- Inline `mapRow` lambda in `listOrgModels` extracted to `pickClientFields(row, creds, defaults): OrgModelInfo` — the fallback chain (`row.x ?? defaults.x ?? null`) now lives in a single named helper

## Why

Cost-shape changes previously required editing three type definitions plus two projection sites in lockstep. After this PR, changing the cost shape is a one-field edit on `ModelMetadata`.

## Behavior

Zero behavioral change. All existing model-related tests pass without modification. No DB schema changes; `org_models` columns are unchanged.

## Test plan

- [x] `bun run check` (typecheck + lint + prettier + verify-openapi + system-packages)
- [x] `bun run test:unit` (791 pass, 0 fail)
- [ ] No DB migration; `org_models` schema unchanged

Closes #443


pierrecabriere:
Good structural intent here. The `ModelMetadata` interface is the right abstraction and the `label?: string` / `label: string` override across the extends hierarchy is handled correctly — TypeScript narrows it properly and `OrgModelInfo` keeps its required-string constraint. Sort order is intact, `??` chains are identical to the old inline lambda, and `Closes #443` is present.

**Concerns:**

1. **`pickClientFields` was only extracted for the DB-row path (`mapRow`), not for the system-model path (`mapSystem`).** `org-models.ts` lines 100–120 still contain a fully inlined `OrgModelInfo` literal with the same six metadata fields hand-typed. The stated goal — "cost-shape changes touch exactly one function" — is not met: the `mapSystem` block is a second edit site. Either route `mapSystem` through `pickClientFields` as well (passing a synthetic row-like object), or acknowledge in a comment that the system path is intentionally separate. As written, the PR adds a named helper and leaves a duplicate of it two dozen lines above, which is the opposite of what the description promises.

2. **`pickClientFields` is only tested indirectly.** `GET /api/models` asserts `status 200` and `Array.isArray(body.data)` but never pins a single field value. A silent field drop (e.g. `cost` or `reasoning`) would pass the suite. Given that this refactor's claim is behavioral equivalence, a small unit test that calls `pickClientFields` directly with known inputs and asserts the output would make that claim verifiable without relying on the reader's code review. Alternatively, extend the existing integration test to assert at least `label`, `modelId`, and one metadata field.

3. **+48 net lines for a consolidation refactor warrants a note.** The description says "~+48 LoC" but doesn't explain where they go. The breakdown: `ModelMetadata` + JSDoc ≈ +25 lines, `pickClientFields` definition ≈ +53 lines, minus the inline `mapRow` removal ≈ -19 lines. The extracted helper is longer than the lambda it replaced because it spells out the input row type explicitly — that is fine and arguably improves readability, but clarifying this in the PR body would help reviewers trust the diff faster.

**Nice-to-have:** The `pickClientFields` row-type parameter is a hand-typed structural interface. It would be cleaner to derive it from `typeof reachableRows[number]` (the actual Drizzle inferred type) so schema changes in the DB table are reflected automatically without a second update here.

PR #455, opened 2026-05-14
feat(runtime): parallel agent/sidecar boot + remove sidecar pool

### LINKED ISSUE #406: Parallelize agent boot with sidecar boot via MCP connection retry
## Context

Today the agent container's startup is **serialized** behind the sidecar's health check, even though the two are launched via `Promise.all` in `apps/api/src/services/run-launcher/pi.ts:197-217`.

The asymmetry is hidden in the orchestrator:

- `createSidecar` (`apps/api/src/services/orchestrator/docker-orchestrator.ts:82-170`) **creates AND starts AND waits for `/health`** (via `startSidecarAndHealthCheck` on the fresh path, or `POST /configure` on the pooled path).
- `createWorkload(agent)` (`docker-orchestrator.ts:173-189`) **only creates + injects files** — it does NOT start the container. The agent is started later, sequentially, in `waitForWorkload` (`pi.ts:261`: `await orch.startWorkload(agent)`).

Result: the agent's Bun runtime boot (~400ms–2s for top-level imports + Phase A/B in `runtime-pi/entrypoint.ts`) runs **after** the sidecar is healthy, not in parallel with it. The sidecar's boot time is fully on the critical path.

## Root cause

The MCP client connection has no retry logic. In `packages/mcp-transport/src/client.ts:121-147`, `createMcpHttpClient` calls `await client.connect(transport)` once and propagates any error. The runtime's Phase C (`runtime-pi/entrypoint.ts:309-330`) catches the error and `process.exit(1)`s, so the sidecar **must** be healthy before the agent starts — otherwise the run fails immediately.

## Boot timeline analysis

Between `docker start agent` and the first MCP call, the runtime does ~400ms–2s of work that does NOT touch the sidecar:

- Bun runtime + top-level imports (`@mariozechner/pi-coding-agent`, `@appstrate/afps-runtime`, `@appstrate/runner-pi`) — ~200–800ms
- Env validation + sink bootstrap — negligible
- Phase A: `initGitWorkspace` + `readBundleFromFile` in parallel — ~50–500ms
- Phase B: `prepareBundleForPi` + dynamic `import()` of extension factories — ~100–500ms
- **Phase C** (first touch of the sidecar): `createMcpHttpClient`

This natural boot window is large enough to fully mask the **warm** sidecar boot (~500ms) if we can connect lazily.

### Phase 1 — MCP connection retry with exponential backoff

Wrap `createMcpHttpClient` with retry-on-connect (deadline ~30s, exponential backoff 50ms → 1s). Retry only on connection-level errors (ECONNREFUSED, DNS ENOTFOUND, transport reset). Treat 401 as fatal (already noted at `client.ts:115`).

Strictly positive change — improves resilience to transient sidecar restarts independent of phase 2/3.

### Phase 2 — Start agent in parallel with sidecar boot

Refactor `pi.ts` so that:

- `createSidecar` no longer awaits `/health` (split health check out of creation, or move it to a non-blocking promise).
- `startWorkload(agent)` is called immediately after both containers are created, without waiting for sidecar health.
- The agent's Phase C retry (from phase 1) handles the small race where the sidecar's DNS alias (`docker-orchestrator.ts:162`) or HTTP server isn't ready yet.

Expected gain: ~500ms saved on the warm path. No regression on the cold path (sidecar still takes 20–45s; agent just waits in MCP retry).

### Phase 3 — Reconsider the sidecar pool

After phase 2 ships, measure in production:

- If pool acquisition (~100ms `/configure`) is fully masked by agent boot (~1s), the pool no longer pays for its complexity.
- If bursty load regularly exhausts the pool and we hit fresh creation (~20–45s) frequently, keep it.

If removed: ~290 LoC in `sidecar-pool.ts` + the `/configure` endpoint + `CONFIG_SECRET` auth in `runtime-pi/sidecar/app.ts` + standby network management + replenish loop + fresh fallback + tests.

## Risks / gotchas

1. **DNS alias timing**: agent may DNS-resolve `sidecar` before it joins the run network (`docker-orchestrator.ts:162`). Retry on `ENOTFOUND` covers this.
2. **Error attribution**: today, sidecar boot failure surfaces before agent start. After phase 2, it surfaces as "MCP connect timeout after 30s" at the agent's Phase C. Need clear log correlation (sidecar logs + agent's failed MCP attempts) so the root cause is identifiable.
3. **Pooled path**: `/configure` adds ~50–130ms before the sidecar serves `/mcp`. MCP retry absorbs this naturally — no special handling needed.

## Acceptance criteria

- [ ] `createMcpHttpClient` (or a wrapper used by `runtime-pi/entrypoint.ts:322`) retries on connection errors with exponential backoff and a configurable deadline.
- [ ] `pi.ts` starts the agent container without waiting for sidecar health.
- [ ] Existing tests pass; new test covers "agent boots, sidecar not yet healthy, MCP eventually connects".
- [ ] Documented in `CLAUDE.md` (Sidecar Protocol section): agent boot is no longer serialized behind sidecar health; pool is retained pending phase 3 measurement.
- [ ] Phase 3 (pool decision) tracked as a follow-up after metrics are gathered.

## Out of scope

- Removing the sidecar pool (phase 3 — separate issue once measurements support the decision).
- Pre-warming the agent runtime (architecturally incompatible: agent container holds per-run state — prompt, bundle, env, files — and is not reconfigurable at runtime).

pierrecabriere:
## Summary

Closes #406.

**1. MCP retry handshake — parallelize agent + sidecar boot**
- Wrap `createMcpHttpClient` with AWS full-jitter exponential backoff (50ms→1s, 30s deadline). Retries only on connection-level errors (ECONNREFUSED/ENOTFOUND/ECONNRESET/ETIMEDOUT/transport reset); 401 + other 4xx fatal.
- Drop the synchronous sidecar `/health` await from `DockerOrchestrator.createSidecar`. `pi.ts` runs sidecar + agent creation under `Promise.all`; the MCP retry absorbs the race for DNS-alias readiness and the warm-path ~500ms sidecar boot.
- Sidecar boot failure surfaces through the agent's MCP retry exhaustion message, augmented by a non-blocking watcher that races the sidecar's exit code and appends the platform-side exit code + tail of stderr.

**2. Remove the sidecar pre-warming pool entirely**
- Deleted `sidecar-pool.ts` (~280 LoC), `POST /configure` from the sidecar, `CONFIG_SECRET` auth, the standby network, replenish loop, and `SIDECAR_POOL_SIZE` env var.
- Empirical measurement: bootDurationMs is ~300ms with or without the pool — the agent's own Bun cold start (~1s) masks the warm-image sidecar boot, so pre-warming bought no user-visible latency.
- `DockerOrchestrator.initialize()` pre-pulls the PI + sidecar images at boot via `ensureImage()` — that single amortisation covers the 20-45s cold-pull worry that originally motivated the pool.

**3. Unify platform-URL resolution**
- Split the sidecar boundary types: `SidecarConfig` (sidecar runtime — keeps `platformApiUrl`) vs new `SidecarLaunchSpec` (platform→orchestrator spawn boundary — no URL).
- Both `DockerOrchestrator` and `ProcessOrchestrator` now call their own `resolvePlatformApiUrl()` inside `createSidecar` instead of trusting a plan-provided URL.
- `plan.runApi: { url, token }` → `plan.runToken: string` (only the token half was ever consumed downstream).

## Test plan

- [x] `bun test packages/mcp-transport` — new retry suite (success-first, retry-then-succeed, jitter-bounds, deadline-exceeded, fatal-4xx, signal-abort)
- [x] `bun test apps/api/test/integration/services/run-launcher-parallel-boot.test.ts` — asserts agent createWorkload resolves BEFORE slow-sidecar createSidecar (parallelism contract) and that the agent start still waits on both
- [x] `bun test apps/api/test/integration/services/platform-container-sink.test.ts` — sink contract still holds with the new boundary type
- [x] `bun test` — full 1592-test suite green (789 unit + 803 core/runtime-pi)
- [x] `bun run check` — turbo tsc + eslint + prettier + verify-openapi green


pierrecabriere:
## Verdict

⚠️ **Ready with caveats.** No outright correctness blockers, but two concerns are real and should be addressed (or explicitly accepted) before this lands on prod traffic: a **cold-start regression risk** (30s MCP deadline < pre-PR ~75s health-wait budget) and a **CI gap** (the new integration test was skipped in this PR's CI run).

The retry logic is well-thought-through and the unit-test coverage of the retry surface is genuinely good. The pi.ts parallel-boot contract is captured in tests that will catch a future re-introduction of the synchronous wait. Documentation update is concise and accurate. The remaining items are operational, not structural.

---

## Blocking issues

None — nothing is structurally broken. But please address the cold-start regression risk below as either a code change or a conscious accept-and-document call.

---

## Non-blocking concerns

### 1. Cold-start regression risk — `runtime-pi/entrypoint.ts:336`, `packages/mcp-transport/src/client.ts:240`

The new MCP retry deadline is hardcoded at **30 s**. Before this PR, the cold-start sidecar boot was absorbed by `waitForSidecarHealth` in `sidecar-pool.ts`, which retries **35 times across ~75 s** (the comment there literally says "Cold-start tail (sum ~71 s) — covers 24–45 s container boot").

Now any environment that hits a cold container pull / cold boot taking 30–45 s will fail the agent's first MCP handshake with `MCP connect deadline exceeded after 30000ms`. The original issue (#406) acknowledges 20–45 s cold boots are normal. The chosen 30 s deadline is shorter than the worst case it's supposed to cover.

Options, in increasing order of effort:
- Bump the entrypoint's `deadlineMs` to 60 s or 90 s — matches the pre-PR effective tolerance.
- Add an env var override on the runtime-pi side (`APPSTRATE_MCP_CONNECT_DEADLINE_MS` or similar) so operators on slow registries can tune without a rebuild. The issue's acceptance criteria explicitly say "configurable deadline".
- At minimum, document this in `CLAUDE.md` as a known trade-off so the next operator who hits it doesn't think it's a bug.

The 30 s default tracks the **sidecar's outbound upstream timeout** (per the JSDoc), but those are different concerns and should not be coupled.

### 2. Integration tests skipped in CI — `apps/api/test/integration/services/run-launcher-parallel-boot.test.ts`

The PR adds a new integration test (`#406 parallel-boot — pi.ts vs slow sidecar`) but the **Integration tests** job is reported as `skipping` on this PR's check run. The unit-test suite covers the retry logic well, but the parallel-boot contract test never actually executed in CI. Either:
- The integration test should run on PRs touching `apps/api/src/services/orchestrator/**` or `apps/api/src/services/run-launcher/**`, or
- The author should run it locally and confirm in the PR description.

The PR description's test plan lists it but does not confirm it was run.

### 3. `expectedSidecarExits` leaks one entry per timed-out / aborted run — `apps/api/src/services/orchestrator/docker-orchestrator.ts:178-220`

Sequence:
1. Run hits `timeoutSeconds` or `signal.abort` → `stopWorkload(sidecar)` → `expectedSidecarExits.add(id)`, container receives SIGTERM and exits.
2. Watcher's `waitForExit` resolves → `has(id) === true` → `delete(id)` → returns silently.
3. `pi.ts` finally block calls `removeWorkload(sidecar)` → **`expectedSidecarExits.add(id)` again** → `removeContainer(id)`.
4. Nothing checks the set after this — the entry remains forever.

Result: long-lived API process accumulates one stale string per timed-out/aborted run. Tiny per-run, but **unbounded over the process lifetime**. Trivial to fix — `removeWorkload` should not re-add to the set, or the set should be cleared in `shutdown()`.

Happy path (no stopWorkload, only removeWorkload) is leak-free because the watcher's `waitForExit` always sees the container removal first.

### 4. `watchSidecarExit` watcher is fire-loud-but-do-nothing — `apps/api/src/services/orchestrator/docker-orchestrator.ts:200-241`

The watcher logs but does not cancel the run or hint anything to the agent's MCP retry loop. So a sidecar that dies on boot at t=2s still forces the agent to wait the full 30 s deadline before failing. The user-facing run record will say `MCP connect deadline exceeded after 30000ms (last error: ECONNREFUSED ...)`; the structured root cause (`exitCode: 1, tail: "npm not found"`) lives only in the operator-side logs.

The PR description claims sidecar-boot failure is "augmented with the platform-side container exit code + last log lines so root cause stays identifiable" — that's true for **operators reading server logs**, not for **users reading the dashboard**. The acceptance criterion in the issue ("Need clear log correlation so the root cause is identifiable") is satisfied at the operator level only.

Consider, as a follow-up: when the watcher detects a non-zero exit and a run is still in-flight, fast-path the agent's MCP retry abort so the run fails immediately with the real reason. Out of scope for this PR but worth tracking.

### 5. `streamLogs` tail in watcher has a 500ms hard timeout — `docker-orchestrator.ts:222-231`

```ts
const timer = setTimeout(() => abort.abort(), 500);
```

On a busy Docker daemon, the multiplexed-stream parser may not have emitted any complete lines within 500ms after a sidecar exit. Result: `tail` is empty, the log line just says `Sidecar exited 1` with no context — the very thing the watcher was added to provide. Consider 2–3 s, or fall back to a one-shot `GET /containers/{id}/logs?follow=false&tail=30`.

### 6. Retry attempt-counter / log-line economics — `runtime-pi/entrypoint.ts:341-356`

Every `onRetry` invocation writes a JSON line to stdout. On a healthy warm-path run the typical retry count is 0–2. On a cold-start day, the agent's 30 s deadline at 50ms→1s backoff yields ~30–50 retries. Each retry consumes one slot of the 200-line stdout ring buffer in `pi.ts:waitForWorkload` — push-out of legitimate early-boot logs is possible.

Also: the comment claims "pino-shaped JSON line" but pino includes `time`, `pid`, `hostname`, `name` — this is just `{level, event, url, attempt, delayMs, errorCode, error}`. Either accept the cosmetic divergence or use the actual pino instance.

### 7. `extractErrorCode` walks `.cause` but never `.errors[]` (AggregateError) — `client.ts:323-336`

`AggregateError.errors[]` is increasingly common in modern fetch implementations (DNS lookup failures sometimes surface as an AggregateError of EAGAIN + EAI_AGAIN, for instance). The walk is depth=8 down `.cause` only — won't find a code nested in `.errors[0].cause.code`. Not currently observed but worth a tracking ticket.

### 8. `isFatalConnectError` regex fragility — `client.ts:362-384`

```ts
const statusMatch = /\b(?:HTTP|status(?: code)?)[:\s]*(\d{3})\b/i.exec(msg);
```

Tightly coupled to the MCP SDK's error-message wording. If the upstream SDK changes the message format (e.g. `"Server returned 503"` or `"POST failed with 503"`), the status will be unrecognised and the error will fall through to "unknown shape → fatal" — silent loss of retry for legitimately retryable 503s. The unit test on 401 only validates `calls <= 2`, which would pass even if a single retry slipped through.

Defensive option: pin the SDK version more tightly, or write a contract test against the actual SDK error-throwing path (not just hand-crafted Responses).

### 9. No test for the `watchSidecarExit` path

The watcher is a non-trivial addition with race-condition surface (the `expectedSidecarExits` set, the 500ms log-tail timeout, the `for await` loop). No test covers "sidecar dies mid-run, watcher fires, log line is emitted with exit code". The parallel-boot integration test is fake-orchestrator-only and never exercises the real Docker watcher.

### 10. 5xx-during-connect treatment is implicit

The retry treats `>=500` as retryable via the message regex, but `runtime-pi`'s typical first error during the boot race is a TCP-level `ECONNREFUSED`, not an HTTP 5xx. If the sidecar happens to return a 503 while `/configure` is still processing (pool path), the regex MUST match for the retry to fire — and that's never tested. Add a test that returns `Response{status:503}` and assert retry behaviour.

### 11. `Math.floor(Math.random() * expCap)` excludes the upper bound

Attempt 0 with `base=50` yields jitter ∈ `[0, 49]` ms, not `[0, 50]` ms. The unit tests assert `≤ 50` so they pass, but the JSDoc says `random(0, min(capMs, baseMs * 2^n))` which reads inclusive. Cosmetic.

---

## Things done well

- Retry policy is **opt-in via an option object**, preserving the legacy single-shot path for all existing callers (tests, future consumers). Clean migration surface.
- AWS full jitter is the right backoff shape for this workload; the rationale is well-commented inline.
- The `extractErrorCode` walk is depth-bounded (8) — no infinite-loop risk on circular `.cause` chains.
- Deadline is checked **before** sleeping, avoiding wasted budget on an unreachable host.
- The integration test asserts `agentCreateResolvedAt < sidecarCreateResolvedAt` — that's exactly the contract the PR is shipping. Future refactors that re-introduce serialisation will trip this.
- The watcher's `expectedSidecarExits` mechanism correctly suppresses the cleanup-path log noise in the happy case.
- `startSidecarAndHealthCheck` is **kept** (still used by the pool's create path), so pool initialization remains synchronous — the only behaviour change is on the per-run path. Conservative and correct.

---

## Suggested follow-up tickets

1. **Operator-tunable retry deadline** (env var `APPSTRATE_MCP_CONNECT_DEADLINE_MS`, default 30s). The issue's acceptance criteria call for a "configurable deadline".
2. **Fast-fail on watcher-detected sidecar death**: hook the watcher up to a signal that aborts the agent's MCP retry, so the user sees the real exit code instead of the deadline message.
3. **Propagate sidecar exit details into the user-facing run record**, not just operator logs. The watcher already has exit code + tail — surface them via the run's failure event.
4. **CI: run integration tests on PRs touching the orchestrator / run-launcher**, or gate the merge on a manual integration run.
5. **Phase 3 (sidecar pool removal)** — per the original issue, gated on warm-path latency metrics post-merge.
6. **AggregateError handling** in `extractErrorCode`.
7. **Tighten the 5xx retry test surface** with an actual 503 Response.

Issue #436, opened 2026-05-13
spike: evaluate Portkey AI Gateway as LLM-side replacement for #431/#432/#433/#434


pierrecabriere:
> **Scope**
> - **Decision gate** for the #427 follow-up roadmap.
> - **Status**: smoke-tested locally on Bun 1.3.11 / macOS arm64 on 2026-05-13. **Pre-validation positive** — Option A (Portkey as stateless proxy, Appstrate DB stays source of truth) is technically feasible. Remaining work: ~4-6h to close the operational and configuration questions before full adoption decision.
> - **Original timebox**: 2 days. **Updated**: ~4-6h based on smoke test pre-validation.

# Background

#427 incident: agent fanning out 8 parallel Gmail fetches → 312 K-token LLM context → OpenAI TPM ceiling → SDK retries exhausted → run reported `success` with null output. #429 patched the reporting bug + added a per-run fan-out cap. The remaining roadmap (#431-#434) builds a SOTA rate-limit/retry/circuit-breaker stack from scratch with `rate-limiter-flexible` + `cockatiel` + OpenTelemetry.

In March 2026, Portkey AI Gateway went **fully open-source (Apache 2.0)** — exposing the same building blocks battle-tested across thousands of production deployments. Before sinking ~2 weeks of careful work into custom code that may be thrown away in 6 months, this spike establishes whether Portkey is a credible drop-in.

# The two architectural options

## Option A — Portkey as stateless proxy (recommended, pre-validated)

DB schema stays as-is. Web app keeps calling `/api/models` and `/api/provider-keys`. At run start, `apps/api` synthesises a Portkey config JSON in memory with the resolved credentials + retry/rate-limit rules, then routes LLM calls through Portkey via the `x-portkey-config` header inline. No Portkey-side virtual-key store, no Portkey DB, no data migration.

| Data | Today | With Portkey Option A |
| --- | --- | --- |
| Provider credentials (encrypted blob) | `model_provider_credentials` | **Same** — Appstrate DB, decrypted in-process per request and passed inline to Portkey |
| Model config (label, baseUrl, cost…) | `org_models` | **Same** — read by API, transformed into Portkey config inline |
| Default model selection | `org_models.is_default` | **Same** |
| TPM bucket runtime state | N/A (to build) | **Portkey runtime** (in-memory single-instance for Tier 0, Redis for Tier 2+) |
| Circuit breaker state | N/A (to build) | **Portkey runtime** |
[...]

Issue #433, opened 2026-05-13
observability: sidecar-side metrics (concurrency, queue, token budget, user-API retries)


pierrecabriere:
> **Scope**
> - **Layer**: Observability — **sidecar-side metrics only** (provider_call concurrency, queue depth, token budget, user-API retries).
> - **Portkey relation**: **Independent of #437.** Portkey emits native OTel metrics for the LLM side — that scope is closed.
> - **Priority**: P3 (prerequisite for calibrating the sidecar caps from #429, #430, #432, #435).
> - **Note**: LLM-side metrics were originally in scope; **dropped after #437 adoption** — Portkey exposes them natively.

Follow-up from #429 (closes #427).

## Problem

#429 introduced two defaults chosen by hand:
- `SIDECAR_PROVIDER_CALL_CONCURRENCY = 3`
- `SIDECAR_RUN_TOOL_OUTPUT_BUDGET_TOKENS = 100_000`

Without metrics, we have no way to verify these are correctly calibrated. We need to observe concurrency wait times, token budget spillover frequency, and user-API retry rates to tune the defaults empirically.

## Recommended stack: **OpenTelemetry**

The 2026 consensus is clear: OpenTelemetry is the Node.js observability standard. Cross-correlation between logs, metrics, and traces via `traceId`/`spanId` shared context.

- `@opentelemetry/sdk-node` + `@opentelemetry/api` for instrumentation
- `@opentelemetry/exporter-prometheus` to expose `/metrics` (Prometheus scrape compat)
- Same endpoint Portkey already exposes its LLM metrics on (Phase 4 of #437) — consolidate dashboards

Verify the stack already in use in `apps/api` before adding deps.

### Concurrency limiter (#429 + #430 + #435)

| Metric | Type | Labels |
| --- | --- | --- |
| `sidecar_provider_call_acquired_total` | counter | `run_id, org_id, provider_id` |
| `sidecar_provider_call_wait_duration_ms` | histogram | `provider_id` |
| `sidecar_provider_call_in_flight` | gauge | `run_id` |
| `sidecar_provider_call_queued` | gauge | `run_id` |

Alert: if `p95(wait_duration_ms) > 5000` → concurrency cap is too low.

### Token budget (#429)

| Metric | Type | Labels |
| --- | --- | --- |
| `sidecar_token_budget_consumed` | counter | `run_id` |
| `sidecar_token_budget_spilled_total` | counter | `run_id, reason` |
[...]

PR #456, opened 2026-05-14
feat(llm-proxy): org-scale TPM bucket per (orgId, modelLabel)

### LINKED ISSUE #431: feat(api): gateway-side TPM bucket per (orgId, modelLabel)
> **Scope**
> - **Layer**: LLM upstream rate limiting (org × model).
> - **Portkey relation**: **Blocked by #436.** If Portkey is adopted, this becomes a configuration task ("wire Portkey TPM limits"). If rejected, build custom with `rate-limiter-flexible`.
> - **Priority**: P1 (org-scale protection; #429 only protects per-run).

Follow-up from #429 (closes #427 at org scale). Item **#1** from the original incident analysis, deferred as platform-side architectural work.

## Problem

#429 added a per-run fan-out cap (`SIDECAR_PROVIDER_CALL_CONCURRENCY=3`) on `provider_call`. This protects **one** run, but nothing prevents **N concurrent runs in the same org** from collectively blowing past the upstream model's TPM window.

Concretely: 10 runs × 3 concurrent provider_calls × ~100 K tokens of accumulated context each = 3 M tokens/min hitting OpenAI. The per-run cap is necessary but **not sufficient**.

### Path A — if #436 adopts Portkey

Configure a TPM limit per `(orgId, modelLabel)` via Portkey virtual keys / route policies. No custom Lua, no Redis primitives in our codebase.

**Acceptance criteria**:
- [ ] Per-org virtual key created on agent run start, bound to the org's TPM allowance.
- [ ] Sidecar credential proxy routes LLM calls through Portkey using the run's virtual key.
- [ ] When the bucket is empty, Portkey returns 429 with `Retry-After` — sidecar surfaces this as a clear `appstrate.error` event (handled by #432 user-API logic + Portkey LLM logic).
- [ ] Integration test: spawn 10 concurrent runs, verify Portkey denies the excess.

### Path B — if #436 rejects Portkey

Add a Redis-backed **token bucket** keyed by `(orgId, modelLabel)` drawn before each LLM upstream hop, using [`rate-limiter-flexible`](https://github.com/animir/node-rate-limiter-flexible) (MIT). SOTA pattern (cf. [TrueFoundry](https://www.truefoundry.com/blog/rate-limiting-ai-agents-preventing-llm-api-exhaustion)):

1. **Bucket config**: each `(orgId, modelLabel)` tuple gets a TPM rate (e.g. 90 % of contractual TPM) and a 1-minute burst.
2. **Draw before send**: estimate tokens (`prompt_chars/3.5 + max_tokens`) and atomically draw.
3. **Local 429 on empty bucket**: refuse with `Retry-After` from refill rate. No upstream hit.
4. **Refill continuously** (token-bucket algorithm, not fixed window).

**Acceptance criteria** (Path B):
- [ ] Redis-backed token bucket with configurable TPM + burst per `(orgId, modelLabel)`.
- [ ] Atomic draw before upstream LLM hop in the sidecar credential proxy.
- [ ] Local 429 + `Retry-After` when bucket is empty.
- [ ] Metric `tpm_bucket.{drawn,denied,refilled}` emitted (see #433).
- [ ] Integration test: 10 concurrent runs, verify denials kick in below the configured ceiling.

## Open questions (regardless of path)

- TPM discovery: hard-coded provider table, or org-level config?
- Multi-model fallback (e.g. `gpt-5.4` → `gpt-5.4-mini`) — separate buckets or shared?
- Org-level config UI for TPM caps (defer to follow-up).

## References

- #427, #429, #436 (decision gate).
- [TrueFoundry 3-layer gateway pattern](https://www.truefoundry.com/blog/rate-limiting-ai-agents-preventing-llm-api-exhaustion).

pierrecabriere:
## Summary

Closes #431

#427 was an incident where 8 parallel agent runs in the same org each respected their per-run fan-out cap but collectively blew past the upstream model's TPM ceiling, exhausted the SDK retry budget, and surfaced as silent null-output `success` runs. PR #429 fixed the per-run side. This PR is the platform-side counterpart: a Redis-backed token bucket via `rate-limiter-flexible` keyed on `(orgId, modelLabel)` that caps cumulative tokens-per-minute across all concurrent runs in the same org, regardless of how well the per-run cap behaves.

Per the #431 acceptance criteria this implements **Path B** (Redis-backed limiter via the platform's existing `rate-limiter-flexible` infrastructure). The Portkey alternative (Path A, gated on #436) was rejected — Portkey itself was dropped today in `e77ff225` ("refactor: drop Portkey gateway, route direct to upstream providers") because the OSS bundle didn't actually ship the features that justified adopting it.

## Highlights

- `apps/api/src/services/llm-tpm-limiter.ts` — `drawTpm({ orgId, modelLabel, estimatedTokens })` returning a typed allow/deny result. Backing store is `RateLimiterRedis` when `REDIS_URL` is set, `RateLimiterMemory` otherwise — wired through the same `getRateLimiterFactory()` the rest of the platform uses. Lua-script `consume()` is atomic across N API instances.
- Bucket capacity is configured via `LLM_PROXY_LIMITS.tpm_buckets` — a strictly-validated map `{ "<modelLabel>": { tpm, burst? } }` with a `"default"` fallback. Empty map = limiter disabled, zero footprint.
- The Redis key is `tpm:${orgId}:${modelLabel}` (per the issue spec) so two distinct models sharing the same `default` capacity still draw from separate buckets — one hot model can't exhaust the default bucket for every other model in the org.
- Token estimation: `ceil(promptChars / 3.5) + max_tokens`, shape-agnostic across OpenAI chat-completions and Anthropic messages (incl. content-as-`{type,text}[]`). Falls back to `DEFAULT_MAX_TOKENS_RESERVATION = 4096` when `max_tokens` is unset. Estimator is deliberately approximate — TPM is a traffic-shaping tool, not a billing meter (`llm_usage` still tracks canonical post-call usage).
- Wired into `proxyLlmCall()` between the response-cache miss and the upstream `fetch()`. Cache hits don't draw (they're free upstream-wise); denies never reach upstream and never mint `llm_usage` rows.
- Deny path returns RFC 9457 `application/problem+json` with `code: "llm_tpm_exhausted"`, `Retry-After` header, and a structured body naming the bucket and the deficit.
- Structured `info` log on every draw (`{ orgId, modelLabel, estimated, remaining, denied }`) — #433 will lift these to proper metrics.
- Backend failure (Redis hiccup) fails open with a `warn` log: better one upstream 429 than locking the whole org out of LLM access.
- Multi-model fallback isolation: a retry on a fallback model draws from the fallback's bucket — per the issue's open question, that is the right behavior.

## Env var

```jsonc
LLM_PROXY_LIMITS='{
  "tpm_buckets": {
    "default": { "tpm": 200000 },
    "gpt-4o-2024-08-06": { "tpm": 50000 },
    "claude-sonnet-4": { "tpm": 80000, "burst": 100000 }
  }
}'
```

Burst defaults to `tpm` (full 60s window = burst). Bucket keys match `ResolvedModel.label`. Omit `tpm_buckets` (or leave it `{}`) to disable the limiter entirely — zero footprint when unused.

## Test plan

- [x] `bun test apps/api/test/unit/services/llm-tpm-limiter.test.ts` — 13/13 pass (budget/deny/isolation/disabled/burst + token-estimation across 6 body shapes)
- [x] `bun test apps/api/test/integration/routes/llm-proxy.test.ts` — 18/18 pass (15 pre-existing + 3 new TPM-bucket: under-budget concurrent allow, over-budget deny with RFC 9457 + Retry-After + `llm_usage` filtered to only allowed calls, no-op when bucket absent)
- [x] `bun test apps/api/test` — 2182 pass, 1 pre-existing flake (`PersistingEventSink > appstrate.metric…writeLedger is on` — fails on `main` too, unrelated to this PR)
- [x] `bun run check` — green across all 19 turbo tasks (typecheck + lint + prettier + verify-openapi + system-package builds). Only the unrelated `packages/ui/.../file-widget.tsx` react-refresh warning is reported.


pierrecabriere:
## Fresh-eye review (independent)

### Verdict
⚠️ Ship after addressing the oversized-draw passthrough + a couple of estimator gaps. Architecturally sound, code is clean, tests are good. Not a blocker on its own — but the clamp behaviour partially defeats #431's purpose.

### What works well
- Clean separation: `drawTpm()` is a small pure-ish service, wired through the existing `getRateLimiterFactory()` (no new infra). Lua-atomic `consume()` is the right primitive for multi-instance.
- Bucket state keyed on `tpm:${orgId}:${modelLabel}` while the *policy* is matched via `tpm_buckets[label] ?? tpm_buckets.default` — distinct models sharing the `default` capacity correctly draw from separate Redis keys (`llm-tpm-limiter.ts:107-120`). Non-obvious correctness win.
- TPM draw correctly placed **after** cache lookup, **before** upstream fetch (`core.ts:99-119`). Cache hits don't draw — claim verified.
- `application/problem+json` shape + `Retry-After` header + structured `retryAfter` body field all consistent with the rest of the codebase (`ApiError` in `packages/core/src/api-errors.ts`).
- `tpm_bucketSchema` and `llmProxyLimitsSchema` both `.strict()` — unknown keys fail-fast at boot.
- Tests don't mock the limiter — the real `RateLimiterRedis` semantics flow through both unit and integration tests.

### Required before merge (must-fix)
- **None blocking.** Everything below is "would improve correctness" rather than "ship-stop."

### Suggestions (non-blocking polish)
- `llm-tpm-limiter.ts:152` — clamping a 5M-token request to 200K capacity lets one oversized call *succeed* the very first time the bucket is empty. The unit test at `llm-tpm-limiter.test.ts:106-122` literally asserts `res.ok === true` for `estimatedTokens: 50_000` against `tpm: 1_000`. That oversized payload still goes to OpenAI — which is exactly the #427 failure mode (upstream rejects, surfaces as null-output). Consider denying outright when `estimatedTokens > capacity` with a different `code` (`llm_request_oversize`) so callers can split the request rather than burn the bucket on a doomed call.
- `llm-tpm-limiter.ts:212` — `logger.warn` for "limiter backend error — failing open" is too quiet. A multi-hour Redis outage would silently disable the limiter across the fleet with no paging signal. Use `logger.error`.
- `proxy-limits.ts:34-42` + `llm-tpm-limiter.ts:113` — when both `tpm` and `burst` are set, `tpm` is *unused* (`capacity = burst ?? tpm`). The field name implies a refill rate that never materialises in a fixed-window scheme. Either drop `burst` (it's redundant) or document loudly that `burst`, when present, **replaces** `tpm` as the window capacity. The PR description ("Burst defaults to `tpm` — full 60s window = burst") is more honest than the inline comment.
- No metric emission (`tpm_bucket.{drawn,denied,refilled}` — issue AC #4). Deferred to #433 — fine, but the PR body doesn't acknowledge this AC line explicitly.
- `policyFor()` does exact-match on `modelLabel`. Versioned suffixes (`gpt-4o-2024-08-06` vs `gpt-4o`) will need separate entries — no prefix/glob match. Worth documenting in `CLAUDE.md`'s env table or `proxy-limits.ts` JSDoc.
- The `denied` log line uses `info` level — at sustained 1k 429/sec on a hammered org this is log volume. Sample or move to `warn`.

### Acceptance-criteria audit (#431 Path B)
- AC1 (Redis-backed token bucket configurable TPM + burst per (orgId, modelLabel)): ✅ — with the caveat that `burst` is a window-capacity override, not a separate refill rate. Issue wording is ambiguous; PR's interpretation is reasonable.
- AC2 (Atomic draw before upstream LLM hop): ✅ — `core.ts:107-119`, before `fetchImpl()` at line 122.
- AC3 (Local 429 + `Retry-After` when bucket empty): ✅ — RFC 9457 problem+json + integer `Retry-After` (seconds), denied call never reaches upstream (integration test `expect(upstreamHits).toBe(2)`).
- AC4 (Metric `tpm_bucket.{drawn,denied,refilled}` emitted): ❌ — structured logs only. Deferred to #433, not delivered here.
- AC5 (Integration test: 10 concurrent runs, denials below ceiling): ⚠️ — the "10 concurrent calls under budget" test exists; the explicit "denials kick in below the ceiling" path is covered by a sequential 3-call deny test rather than 10 concurrent runs over-ceiling. Acceptable, but doesn't quite match the issue's wording.

### Honest gaps acknowledged
The PR body honestly flags: deferral of metrics to #433, fixed-window vs token-bucket semantics ("burst subsumed by tpm"), and the deliberate approximate nature of the estimator. The author **did not** acknowledge: oversized-request passthrough behaviour (the clamp), `tools[]`/function-definition undercounting, and `tool_use`/`tool_result` blocks not being counted at all. The Anthropic `tools[]` undercount can be material — a 50-tool definition can easily be 5-10K tokens and is entirely invisible to the estimator.

### Algorithm + token-estimation deep-dive
`rate-limiter-flexible`'s `RateLimiterRedis` is a **fixed-window counter** (Lua: `INCRBY` + `PTTL`, `PEXPIRE` on first increment). The author's claim is correct. `msBeforeNext` reflects time until the current key TTL expires — so `Retry-After` is an honest "wait until this window resets," not a sliding-window approximation. Multi-instance correctness is guaranteed by the atomic Lua script — 10 concurrent draws from 10 API replicas serialise to a single Redis `INCRBY` per call with no double-spend. The Tier 0 fallback (`RateLimiterMemory`) is per-process — fine for single-instance dev/OSS, but not documented as such in `LLM_PROXY_LIMITS` env docs.

The estimator (`helpers.ts:98-150`) covers OpenAI `messages[].content` (string or `{type:"text", text}[]`), Anthropic `system` (string or array), and `messages[].content` arrays. It **does not** count: `tools[]` / `functions[]` definitions (both OpenAI and Anthropic — multi-thousand-token undercount in tool-heavy agents), Anthropic `tool_use.input` JSON, Anthropic `tool_result.content` (long shell/file outputs are invisible — most likely real-world undercount path for the AppStrate agent fleet, since `tool_result` is exactly what gets injected back into the loop after each `bash`/`Read` call), OpenAI `tool_calls[].function.arguments`, and any image/audio multimodal content. For TPM as a *traffic-shaping* tool this is "fine, approximately right"; for fan-out agents that hammer tools, the cumulative undercount can be 30-50%+ of true token cost. Worth a follow-up issue, not a merge blocker.

Issue #199, opened 2026-04-19
Bootstrap installer hangs at tier prompt — Bun-compiled CLI doesn't receive keypresses on macOS (regression after #190)


pierrecabriere:
## Summary

After PR #190 (`</dev/tty` redirect fix for #184), the `curl … | bash` bootstrap no longer SIGKILLs at the handoff to `appstrate install` — the tier prompt now renders correctly. **But it hangs there**: keystrokes (arrows, Enter) are not received by the CLI, the terminal is left in a broken raw-mode state, and the only recovery is killing the process and running `reset`.

The fix from #190 wasn't wrong; it surfaced a deeper bug. **The bug is in the bun-compiled binary itself**, not in the bootstrap script.

## Reproduction

- macOS (arm64), zsh, terminal emulator (Terminal.app / iTerm2 / Ghostty — same behavior)
- Latest published CLI: `cli@1.0.0-alpha.61` (built with Bun 1.3.11, includes #190's fix)

```sh
curl -fsSL https://get.appstrate.dev | bash
```

Crashes as described.

**Critical observation**: the bug **also** reproduces with the workaround that previously sidestepped #184:

```sh
curl -fsSL https://get.appstrate.dev -o /tmp/inst.sh
bash /tmp/inst.sh
```

This rules out the \`curl | bash\` pipe and the \`</dev/tty\` redirect entirely. The binary itself fails to read keypresses regardless of how it's invoked through bash. The same binary ran directly (`/path/to/appstrate install`) from the user's interactive shell needs to be tested — see TODO below.

## Root cause analysis

The CLI is shipped as a `bun build --compile` standalone binary (the Bun runtime is embedded — users don't need bun installed). The interactive prompt uses `@clack/prompts` (^1.2.0) which calls `process.stdin.setRawMode(true)` and listens for keypress events.

This matches a **documented Bun runtime regression family** on macOS, where `setRawMode(true)` succeeds (terminal goes raw — hence the \"frozen\" feel) but the read loop on fd 0 never delivers keystrokes to JS. clack's keypress handler is registered on \`process.stdin.on('data')\` after \`setRawMode\` — no events fire, prompt waits forever.

The terminal stays in raw mode after kill because no one restores cooked mode on exit, hence the need for \`reset\`.

Relevant upstream issues:

[...]

PR #481, opened 2026-05-17
feat!: AFPS 2.0 migration + provider→integration unification + integrations rollout

pierrecabriere:
## Summary

Complete **Integrations rollout** per `docs/architecture/INTEGRATIONS_PROPOSAL.md`, then the **provider → integration unification** that supersedes it: the legacy `provider` package type, the `provider_call` agent tool, and the connection-profile / app-profile features are removed; the 59 first-party providers are re-modelled as `api_call` integrations; and a final audit-driven cleanup pass removes the residual dead code and unifies naming. The AFPS `tool` package type is also removed (Phase 3 — system tools become built-in runtime tools).

PR is **production-ready** and ready to merge. Each phase landed as its own commit so reviewers can follow the rollout incrementally; the post-Phase-3 work is documented in the **Provider → Integration unification** section at the bottom.

## Roadmap

| Phase | Goal | Status |
|------|------|--------|
| **1.0** | Import a minimal `.afps` integration package into DB (no run) | ✅ Done |
| **1.05** | CLI `afps bundle` vendors deps into the package | ✅ Done |
| **1.1** | Credential layer: multi-auth (`auths`), delivery rules, RFC 8707 audiences, RFC 9728 discovery | ✅ Done |
| **1.2a** | Runtime spawn — orchestrator, supervisor, server resolver, toolsDynamic, SIGHUP, CA planner | ✅ Done (pure code) |
| **1.2b** | MCP HTTP transport OAuth 2.1 — RFC 7591 DCR, scope step-up, RFC 8707 authorize URL | ✅ Done (pure code) |
| **1.2c** | Wire-up — Bun.spawn adapter, openssl CA generator, bootstrap controller | ✅ Done |
| **1.2d** | HTTPS MITM listener — per-host SNI + TLS re-encrypt + credential injection | ✅ Done |
| **1.3** | Marketplace UI + agent editor wiring | ✅ Done |
| **1.4** | Env-delivery boot + container-per-integration runtime (agent-side) | ✅ Done |
| **1.5** | In-run credential refresh + `delivery.http` MITM injection | ✅ Done |
| **E2E** | Cross-phase Playwright / integration suite | ✅ Done |

## Phase 1.0 — landed

- `@appstrate/core/integration` Zod schema (server discrimination, D31, D32, multi-auth + delivery)
- `agent.schema.json` extended with `dependencies.integrations`
- `packageType` enum gains `integration` (migration 0023)
- `parsePackageZip` handles integration zips (manifest + optional `INTEGRATION.md`)
- `integration-service.ts` — read-only DB surface (org+system scoping, broken-manifest tolerance)
- 33 schema unit tests + 19 service tests + 6 end-to-end import tests

## Phase 1.05 — landed

- Schema extended: `server.package` is now a discriminated union on `registryType` (`oci` | `npm` | `pypi`); `npx`/`uvx` author sugars accept either `entryPoint` or `package(npm|pypi)`.
- `@appstrate/core/integration-bundle` — pure bundler with injectable network/subprocess boundaries.
- `afps bundle <manifest>` CLI command in `packages/afps-runtime`.

## Phase 1.1 — landed

- `@appstrate/connect/audience-binding` — RFC 8707 helpers.
- `@appstrate/connect/oauth-discovery` — RFC 9728 → RFC 8414 cascade with SSRF + allowlist + 24h cache.
- `@appstrate/connect/integration-credentials` — multi-auth resolver + delivery planners + URL routing.
- `integration_connections` table (migration 0024) — owner XOR + `coalesce` unique trick.

## Phase 1.2 — landed

All pure / DB-free / network-free with injectable I/O boundaries for hermetic tests.

**Phase 1.2a — MCP Runtime stdio**:
- `SubprocessTransport.killTimeoutMs` default 1s → **10s** (spec §5.4.2)
- `McpHost` namespace disambiguation with `_2`/`_3` suffix on collision + audit event (§5.4.5)
- `@appstrate/connect/integration-runtime` — server resolver (entryPoint + OCI), spawn command builder (D31 — node/bun → `bun`), `compatibility.mcp` / `compatibility.afps` semver-range validation, proxy 6-tuple env injection
- `@appstrate/connect/restart-supervisor` — 1s/5s/15s backoff schedule, injectable clock
- `@appstrate/connect/integration-orchestrator` — top-level parallel spawn + ordered teardown + SIGHUP cred-refresh dispatch
- `@appstrate/connect/tools-discovery` — at-spawn `tools/list` discovery, 24h cache keyed by `(integrationId, userId, connectedAuthsKey)`, drift detection vs `tools.lock.json` (added/removed/schemaChanged)
- `@appstrate/connect/proxy-ca-planner` — per-run CA bundle paths + tmpfs policy (`0444`/`0400`), injectable `CertGenerator`

**Phase 1.2b — MCP HTTP transport OAuth 2.1**:
- `@appstrate/connect/dynamic-client-registration` — RFC 7591 DCR with SSRF protection
- `@appstrate/connect/mcp-http-auth` — `WWW-Authenticate` parser, bearer-bearing `fetch` wrapper with scope step-up retry (capped to defeat auth loops), authorize URL builder (RFC 8707 `resource=` + PKCE S256 mandatory)

## Phase 1.2c — landed

Runtime-pi wire-up that composes Phase 1.2a/b pure modules into a real boot pipeline.

- `runtime-pi/sidecar/integration-spawner.ts` — `Bun.spawn` adapter producing a `SpawnedChildHandle`. SIGTERM grace gated on `exited` promise, escalates to SIGKILL after `killTimeoutMs`.
- `runtime-pi/sidecar/ca-cert-openssl.ts` — real X.509 generator via `Bun.spawn("openssl")`. Root + leaf 2048-bit RSA; leaf carries AKI (POC #3a) + EKU `serverAuth` + SAN.
- `runtime-pi/sidecar/integration-runtime-controller.ts` — `bootstrapIntegrationRuntime` composes `planCaBundle` → fs writes → per-integration `resolveIntegrationServer` + `buildSpawnCommand` + `buildProxyEnvInjection` → `spawnIntegrations(parallel)`. Returns a controller with `caBundle` (for 1.2d listener), `running`, `childFor()`, `refreshCredentials()`, idempotent `shutdown()`.

## Phase 1.2d — landed

Per-integration HTTPS MITM listener that closes the strip/inject loop the spec prescribes (§4.1.4).

- `@appstrate/connect/integration-mitm-planner` (pure) — given a request URL + caller header names + per-auth `HttpDeliveryPlan` map, returns the `MitmAction`.
- `runtime-pi/sidecar/integration-cert-minter.ts` — per-SNI leaf cert minter signed by the run CA via `openssl`. LRU cache keyed by host (default cap 256).
- `runtime-pi/sidecar/integration-mitm-listener.ts` — outer TCP server parses CONNECT, peeks the TLS ClientHello to extract SNI via a hand-rolled RFC 6066 parser, then per-SNI Bun.serve relays raw bytes for TLS termination.
- Why per-SNI Bun.serve: `tls.createServer({SNICallback})` and `new TLSSocket(socket, {isServer: true, ...})` are broken in Bun 1.3.11.
- Controller wire-up: optional `enableMitmListener: true` starts one MITM listener per integration that supplied a `credentialSource`.

## Phase 1.3 — landed

End-user marketplace UI for AFPS integrations + agent editor wiring.

**Backend** (`apps/api`):
- New `integrations` core RBAC resource (read / install / uninstall / connect / disconnect), wired into role-grant matrix + API-key allowlist.
- **13 new REST endpoints** under `/api/integrations/*` — list / installed / detail / install / uninstall / OAuth client CRUD / connect/fields / connect/oauth2 / callback (popup HTML) / connections list & disconnect.
- `services/integration-connections.ts` — orchestrates the multi-auth model: connection upsert keyed by `(package, auth, accountId, owner)`, identity extraction via the manifest's `extractTokenIdentity` JSONPath, v1-envelope encryption for both connection credentials and admin-registered OAuth client secrets.
- Migration `0025_integration_oauth_clients.sql` — per-application OAuth2 client registration table (clientId + encrypted clientSecret + optional redirectUri), unique on `(applicationId, integrationPackageId, authKey)`.

**Connect package** (`@appstrate/connect`):
- New pure module `integration-oauth.ts` — PKCE S256 mandatory, RFC 8707 `resource` on authorize+token requests when audience declared, `client_secret_post` / `client_secret_basic` / `none` (public client) support, scope shortfall/creep classification. State record carries every flow context field so callback never re-fetches the manifest.
- `OAuthStateRecord.integration` — discriminator dispatched by the routes layer.

**Frontend** (`apps/web`):
- Hooks: `use-integrations.ts` (10 hooks — list/detail/install/uninstall/connect-fields/connect-oauth/disconnect/OAuth-client CRUD).
- Pages: `integrations-page.tsx` (marketplace catalogue with all/installed tabs + search), `integration-detail.tsx` (manifest metadata + per-auth section with multi-account list, fields-connect modal, popup-driven OAuth flow, admin OAuth client form with public-client toggle).
- Agent editor: new "Integrations" tab via existing `ResourceSection` with `type="integration"`, writing to `manifest.dependencies.integrations`.
- Sidebar: new "Intégrations" entry under the integrations section.
- i18n: fr + en (~70 new keys per locale).

**Deferred** (documented, non-blocking):
- OAuth Mode B (RFC 9728 discovery) user-facing connect flow — runtime spawn path already consumes it; marketplace requires explicit `authorizationUrl` + `tokenUrl` in the manifest for now.
- OAuth1 marketplace connect — `custom` auth covers as fallback.
- Per-agent multi-account selection at runtime — connection model + UI ship; runtime resolver still picks first matching connection until DB reads through `MitmCredentialSource` land.

## Phase 1.4 — landed

End-to-end agent-side runtime that takes an installed integration + a live `integration_connections` row and surfaces its MCP tools to the agent's LLM. Deliberately re-shapes the original "system-packages migration" plan into a **container-per-integration** architecture so the sidecar image stays language-agnostic (one new runtime = one new Dockerfile, never a fatter sidecar).

**Platform** (`apps/api`):
- `services/integration-spawn-resolver.ts` — resolves the agent's `dependencies.integrations[id]` chain: verifies the package is installed on the app, looks up the actor's `integration_connections` row, decrypts the v1-envelope blob, and builds `spawnEnv` from `manifest.auths.{key}.delivery.env`. Returns `null` on missing/expired/broken so the run continues without that integration's tools rather than hard-failing.
- `env-builder.ts` + `run-launcher/pi.ts` — plumbs `plan.integrations` (typed as `IntegrationSpawnSpec[]`) through to the sidecar launch spec.
- `orchestrator/docker-orchestrator.ts` — when (and only when) `spec.integrations.length > 0`, mounts `/var/run/docker.sock` on the sidecar via `Binds` and overrides `User=0:0`. Runs without integrations keep the default `nobody:nobody` user with no socket access — no privilege uplift for runs that don't need it.
- `services/docker.ts` — `CreateContainerOptions` gains `binds`, `groupAdd`, `user` (HostConfig fields).
- `routes/internal.ts` — new `GET /internal/integration-bundle/{scope}/{name}` (Bearer run-token, dep + install double-check, `application/zip` body). Bundle bytes ship out-of-band so we don't fight Linux env-var size limits.

**Sidecar** (`runtime-pi/sidecar`):
- `Dockerfile` — adds `docker-cli`. Carries NO language interpreter. Final image: 132 MB.
- `integrations-boot.ts` — reads `INTEGRATIONS_TO_SPAWN_JSON` at boot, fetches each bundle via the new internal endpoint, extracts to tmpfs, then spawns one runner container per integration via the mounted socket: `docker create --rm -i --cap-drop ALL --memory 256m --pids-limit 128 -e <spawnEnv> <runner-image> <entry>` → `docker cp` the bundle into `/bundle/` → `SubprocessTransport(["docker","start","-ai",<id>])` to pipe MCP stdio. Wraps each result in an `AppstrateMcpClient` and registers on a shared `McpHost` under a `{namespace}__{tool}` prefix. Auto-detects Docker availability and falls back to host-subprocess `Bun.spawn(["node"|"python3", entry])` when no daemon is reachable (process-mode dev / tests).
- `server.ts` + `app.ts` + `mcp.ts` — non-blocking boot (Bun.serve binds before integrations finish spawning; `tools/list` awaits the boot promise once on first call) so the agent's MCP handshake doesn't race on a not-yet-listening sidecar.

**Runner images** (`runtime-pi/runners/`):
- `node/Dockerfile` — Alpine + `nodejs` + ca-certs. ENTRYPOINT `["node"]`. 75 MB.
- `python/Dockerfile` — Alpine + `python3` + ca-certs. ENTRYPOINT `["python3","-u"]`, `PYTHONPATH=/bundle/vendor:/bundle`. 49 MB.
- `binary/Dockerfile` — Alpine + `libstdc++` + `libgcc` + ca-certs. ENTRYPOINT execs the bundle binary as PID 1. 11 MB.

**Agent runtime** (`runtime-pi/mcp/direct.ts`):
- New `buildIntegrationToolFactories` enumerates every non-first-party tool advertised by the sidecar (i.e. spawned-integration tools) and registers one Pi extension per tool that forwards verbatim to `mcp.callTool`. The LLM sees them under the `{ns}__{tool}` name advertised by `tools/list`.

**MCP transport** (`packages/mcp-transport/src/client.ts`):
- Critical bug fix in `RETRYABLE_CODES`: Bun's fetch errors use PascalCase codes (`ConnectionRefused`, `DNSNotFound`, `ConnectionTimeout`, …) instead of Node's E-prefix codes. Without the Bun aliases, `isFatalConnectError()` would classify every Bun-side connection failure as fatal and bypass the retry loop entirely — the agent container would fail its very first MCP handshake against a still-booting sidecar instead of riding through the warm-path race. The container-per-integration spawn extended that warm-up window, making the race near-deterministic.

**System packages** (`system-packages/`):
- `integration-dice-1.0.0.afps` — pure-stdlib Python MCP server (no auth, hand-rolled JSON-RPC over stdio so the bundle has zero third-party deps — proves the python runner without coupling to platform-specific pydantic wheels).
- `agent-dice-game-1.0.0.afps` — agent that exercises `appstrate_dice__roll_dice` end-to-end, validates the entire chain (resolver → bundle delivery → spawn → MCP call → tool forwarding).

## Phase 1.5 — landed

Closes the in-run credential refresh + `delivery.http` MITM gaps left open by Phase 1.4. For integrations whose manifest declares `delivery.http`, the sidecar now stands up a per-integration HTTPS MITM proxy that injects credentials transparently per upstream request and force-refreshes OAuth2 tokens on `401`. The integration's MCP server never reads the credential — that contract is now invariant.

**Platform** (`apps/api`):
- `routes/internal.ts` — two new endpoints behind the same Bearer run-token + dep-and-install guard:
  - `GET /internal/integration-credentials/{scope}/{name}` — live decrypted credentials + per-auth `HttpDeliveryPlan` + `expires_at`. OAuth2 tokens refresh proactively when within `OAUTH_REFRESH_LEAD_MS` of expiry.
  - `POST .../refresh` — force-refresh every OAuth2 auth on this integration. `403` ↔ refresh-token revoked (connection flagged `needsReconnection`); sidecar translates to `401` for the integration's MCP client.
- `services/integration-credentials-resolver.ts` — live resolver shared by both endpoints. Decrypts the connection blob, runs the OAuth2 refresh (proactive or forced) via the new helper, and assembles the wire payload using `resolveHttpDelivery` from `@appstrate/connect`.
- `services/integration-token-refresh.ts` — `forceRefreshIntegrationConnection` mirrors `@appstrate/connect/token-refresh.forceRefresh` (which targets `user_provider_connections`) but writes back to `integration_connections`. Same `RefreshError` taxonomy (`revoked` vs `transient`); `invalid_grant` flips `needsReconnection`. In-flight dedup + per-connection lock so concurrent refreshes coalesce.
- `services/integration-spawn-resolver.ts` — `resolveDeliveries` unifies env + http delivery into one decrypt-per-auth pass, exposing `httpDeliveryAuths` on the spawn spec so the sidecar knows which integrations need MITM treatment without re-reading the manifest.
- `packages/core/src/sidecar-types.ts` — `IntegrationSpawnSpec.httpDeliveryAuths` field.

**Sidecar** (`runtime-pi/sidecar`):
- `Dockerfile` — adds `openssl` (~3 MB). Without it CA bring-up fails at boot and the sidecar continues with env-delivery-only integrations.
- `integration-credentials-source.ts` — `MitmCredentialSource` factory that talks to the platform endpoints. Exposes `current()` / `deliveryPlans()` cached from the boot-time GET, plus `refreshOnUnauthorized(authKey)` that POSTs to `/refresh` and swaps the cache in place. 5 s per-`authKey` cooldown + in-flight dedup guards against refresh storms when the 401 isn't an expiry issue.
- `integrations-boot.ts` — for each integration carrying `httpDeliveryAuths`:
  1. Mint a per-run CA bundle ONCE via `planCaBundle({ generator: createOpensslCertGenerator() })`. Cert PEM lands on local fs (mode `0444`) for `docker cp`; CA private key never touches disk (memory-only on the sidecar).
  2. Start a per-integration `createIntegrationMitmListener` bound to a `createCertMinter` (lazy per-SNI leaf certs).
  3. Spawn the runner container with `--network container:<sidecar-id>` so `127.0.0.1` reaches the listener bound on the sidecar's loopback (avoids dedicated docker networks). Env: `HTTPS_PROXY`/`HTTP_PROXY`/`NO_PROXY` + `NODE_EXTRA_CA_CERTS`/`SSL_CERT_FILE`/`REQUESTS_CA_BUNDLE`/`CURL_CA_BUNDLE` pointing at the mounted CA. `docker cp` the CA cert to `/etc/appstrate/ca.pem`.
  4. Process-mode fallback uses the same env vars on the subprocess — no `--network container:` (inherits parent NS).
- `delivery.env`-only integrations are unchanged — same baked-in env path as Phase 1.4.

**System packages** (`system-packages/`):
- `integration-mitm-test-1.0.0.afps` — pure-stdlib Python MCP server with a `call_upstream` tool. Its `api_key` auth declares `delivery.http` injecting `X-Mitm-Test-Token`. The integration never reads the API key — proves the injection happens entirely sidecar-side. Used by the Phase 1.5 manual E2E.

**Documentation**:
- `CLAUDE.md` — replaces "Phase 1.5 not yet implemented" with the full landed-state section: wire diagram, endpoint contracts, refresh helper pointer, and the six load-bearing invariants (network model, refresh storm protection, CA hygiene, graceful degradation, `allowServerOverride`, reference integration).

## Phase 2 — Connection management & flat model — landed

Follow-on rollout (21 commits) reworking how connections are picked, governed, and resolved at runtime. Replaces per-`(integration, authKey)` keying with **one connection per integration** — the connection's own `authKey` drives credential injection, so OAuth and `api_key` connections are interchangeable at runtime.

**Resolver cascade (single source of truth):** admin pin → run override → schedule override → member pin → accessible-connections fallback. `insufficient_scopes` is enforced on the **resolved** connection only (412 when the picked connection lacks the agent's required scopes), with in-place incremental-consent upgrade for the actor's own connection. Pure `resolveConnections()` + DB orchestrator `resolveConnectionsForRun()`, 22 unit tests.

**Schema (migrations 0027–0029):** `integration_pins` gains nullable `user_id` (NULL = admin force pin, NOT NULL = member preference pin); flat-connection drop of the old per-authKey unique; member-pin uniqueness via computed-expression UNIQUE INDEX.

**Runtime snapshot path (#199):** admin pins + run/schedule overrides honored end-to-end from kickoff to sidecar MITM credential refresh. `runs.resolved_connections` snapshot pins the exact connection row; `loadActorConnection` accepts a `connectionId` override; scheduler propagates `package_schedules.connection_overrides`.

**API:**
- `GET /api/me/connections` — unified user-scope aggregator (provider + integration, cross-org)
- `GET/PUT/DELETE /api/me/integration-pins` — member-self pin CRUD
- `DELETE /api/me/connections/:id` — single owner-scoped destructive delete (cascades to pins/snapshots/overrides), replacing the dangerous global agent-context delete
- `GET /api/integrations/:packageId/accessible-connections` — own + shared (with `ownerName`, `scopesGranted`)
- `GET /api/integrations/:packageId/consuming-agents`
- `GET /api/integrations/:packageId/agent-resolution/:agentPackageId` — server-side picker verdict (collapses the front-end's 5-query cascade re-implementation into one)
- admin pins / `block_user_connections` toggle / label + share-with-org on connections
- `connectionOverrides` accepted on `POST /run` and schedule create/update

**UI:**
- Unified `/connections` page (source-grouped, inline label edit, share toggle, reuse hint)
- Pre-run `must_choose` picker on the agent page (rich dropdown: own + shared, owner/type, admin-pin lock, per-connection scope badges, collapsed-by-default)
- Centralised pin governance table on the integration detail page (admin)
- Split reconnect vs. upgrade-scopes labels (in-place consent semantics)

**Tech-debt cleanup (final commit):** types hoisted into `@appstrate/shared-types` (were redeclared identically front + back), shared `pickAnyAccessibleConnection` across spawn + credentials resolvers, dead-code removal (`assertCanManageActorConnection`, `fallback_default` enum member, no-filter branch, dead i18n keys), stale `integration_member_assignments` docblock fixed.

> ⚠️ No production data exists, so migrations 0027–0029 are safe to merge as-is (no data backfill). Pin tasks P6 (data migration) and P7 (cleanup after stabilisation) are no-ops in this context.

## Phase 2.1 — Org-wide default connection — landed

Cross-agent governance primitive (6 commits) addressing the "you pin agent-by-agent" scaling gap. One **default connection per `(application, integration)`** covers every consuming agent; per-agent pins become exceptions.

**Resolver — 2 new cascade layers (7 total):**
```
1. admin pin (per-agent force)
2. org default ENFORCE      ← NEW (org-wide force; per-agent pin still wins)
3. run override
4. schedule override
5. member pin
6. org default SOFT         ← NEW (org-wide default; member pin still wins, kills must_choose)
7. fallback
```
`enforce: true` locks members org-wide; `enforce: false` is a soft default a member can override. `ConnectionResolutionSource += org_default | org_default_enforced`. The resolver input is optional → fully inert until an admin sets a default (no behavioural change in OSS).

**Schema:** `integration_org_defaults` (migration 0030, hand-written idempotent SQL — `db:generate` is unusable here, snapshots squashed at 0000). Unique on `(application, integration)`. Same `sharedWithOrg` requirement as admin pins; FK `ON DELETE CASCADE`.

**API (admin):** `GET/PUT/DELETE /api/integrations/:packageId/default` (OpenAPI + verify-openapi +3). Unshare guard extended: refuse `sharedWithOrg=false` while a connection is an org default. `validatePinTarget` reused (not duplicated).

**UI:** integration detail page gains a **Default connection** section (shared-conn select + Enforce toggle) above the per-agent table, now relabelled **Per-agent exceptions**. The agent-page `MemberConnectionPicker` renders an enforced default as the same read-only lock as an admin pin. `IntegrationAgentResolution` carries `orgDefaultConnectionId` + `orgDefaultEnforced`.

**Tests:** +7 pure resolver unit tests (enforce vs pin/override/member precedence, soft kills must_choose, soft falls through when connection gone, scope-deficient default → insufficient_scopes) — 36/36. +7 route tests (DB-backed, CI). `bun run check` green (20/20).

> No prod data → migration 0030 is safe as-is; the table is born empty and the resolver falls back to prior behaviour until a default is set.

## Phase 2.2 — Integrations UX overhaul — landed

Reworks the integrations surface to mirror the providers page and drops the install/uninstall framing in favour of **activate/deactivate** (single commit).

**Lifecycle rename (activate/deactivate):**
- API: `POST /api/integrations/:packageId/activate` + `DELETE /api/integrations/:packageId/deactivate` replace the `/install` routes; list field `installed` → `active`; audit actions `integration.activated` / `integration.deactivated`. **Deactivation is non-destructive** — connections / OAuth clients / pins / org-defaults FK to `(package, application)`, not to `application_packages`, so they survive and are reused on reactivation (mirrors disabling a provider). OpenAPI + verify-openapi + shared-types + hooks updated.

**List page (provider-aligned):** `Activées` / `Installées` tabs, an `Activée` / `Non activée` badge per card, and a single `Configurer` button → detail page. No more install/uninstall buttons on cards.

**Detail page (tabbed):** header carries the `Activer` / `Désactiver` toggle; body split into 3 tabs — **Connexions** (per-auth account list + connect, for everyone), **Accès** (admin: OAuth client configuration, then governance — block member connections, org-wide default, per-agent pin exceptions), **À propos** (metadata). Inactive integrations show an activation prompt in the first two tabs.

**Configure ≠ connect split (admin OAuth client moved to Accès):** the per-auth OAuth client registration form (app-level `clientId`/`secret` registered with the IdP — the analog of a provider's admin-credentials config) moved out of the per-section Connexions view into a dedicated **OAuth client configuration** group at the top of the **Accès** tab. Connexions now only lists accounts; an oauth2 section with no client registered shows a pointer to the Accès tab instead of a missing connect button. `AuthSection` drops its now-unused `isAdmin` prop.

**Fixes & cleanup bundled in:**
- Removed the now-false multi-auth banner (claimed "one auth method per user at a time" + recommended a dedicated end-user for CI — neither holds under the flat connection model).
- Per-auth-section "Ajouter un compte" button is now scoped to its section's method (new `lockToAuthKey` on `InlineConnectButton`) — the OAuth section no longer offers `api_key`. The multi-method dropdown is retained on the agent surfaces where one button stands in for the whole integration.
- Legacy remnants swept: stale install/uninstall docblocks (route header, detail-page header, `RequiredScopesPanel`), `agentEditor.integrations.subtitle` copy (installed → activated), and a dead `notInstalledHint` i18n key removed (fr + en).

`bun run check` green (20/20). Route + e2e tests updated for the renamed endpoints; DB-backed route tests run in CI.

## Quality gate

- `bun run check` — 20/20 turbo tasks green (typecheck, lint, prettier, **verify-openapi (288 endpoints, +13 in 1.3, +1 in 1.4, +2 in 1.5)**, Zod ↔ OpenAPI schemas, code ⊆ spec)
- `bun test` — **5950/5950 pass** (full repo suite after Phase 3 tool-type removal)
- `apps/web` Vite build — clean, 491ms
- E2E smoke: `e2e/tests/integrations/marketplace.api.spec.ts` (7 specs against live API)

## Test plan

- [x] Phase 1.0 unit + DB-backed service + end-to-end import tests
- [x] Phase 1.05 bundler unit + Bun-probe + CLI tests
- [x] Phase 1.1 credential resolver + audience-binding + discovery tests
- [x] Phase 1.2a runtime + supervisor + orchestrator + tools-discovery + CA planner tests
- [x] Phase 1.2b DCR + MCP HTTP auth tests
- [x] Phase 1.2c spawner (14) + openssl CA (7, runs real openssl) + bootstrap controller (10) tests
- [x] Phase 1.2d MITM planner (17 pure) + cert minter (8, runs real openssl) + listener (7, real TCP+TLS) + controller wire-up (3) + subprocess E2E (1) tests
- [x] **Phase 1.3 integration-oauth (15 pure) + routes (15, real DB) + cross-phase E2E smoke (7)**
- [x] **Phase 1.4 manual E2E** — Python dice integration verified against `appstrate-mcp-runner-python` (rolled 3d20 → tool result surfaced verbatim to the LLM → output validated). Node runner additionally verified by hand against a custom Gmail MCP server (real Gmail API call, real OAuth token).
- [x] **Phase 1.5 unit tests** — 10/10 `MitmCredentialSource` factory tests (initial payload exposure, refresh round-trip + bearer header, 403-revoked cooldown, network-error tolerance, 5xx pass-through, in-flight refresh dedup, malformed-JSON recovery, `fetchInitialIntegrationCredentials` success + error paths).
- [ ] **Phase 1.5 manual E2E** — exercise `@appstrate/mitm-test` against a mock upstream (CONNECT preamble → SNI parsed → leaf minted → header injected → response forwarded). Validate `/refresh` round-trip with an expiring OAuth token. Pending: requires a local mock upstream + OAuth provider setup.
- [x] `bun run check` (20/20 turbo tasks, **288 endpoints**)
- [x] `bun test` (**5950/5950**)
- [x] `apps/web` Vite build (clean)

## Phase 3 — Remove AFPS `tool` package type — landed

Removes the `tool` AFPS package type entirely. The five former system tools (`output` / `log` / `note` / `pin` / `report`) become **built-in runtime tools** baked into the runtime image and selected per agent via a new top-level `runtimeTools: string[]` agent-manifest field. `output` is always injected; `log` / `note` / `pin` / `report` are opt-in.

**Why built-in (not integrations):** these tools run in-process and feed the agent-side run-result aggregate/finalize path (they emit the same stdout CloudEvents). Routing them through the integration/subprocess channel would relocate that aggregation boundary and add round-trips for no benefit.

**Core (`@appstrate/core`):** new `runtime-tools-catalog` (`MANDATORY`/`SELECTABLE` lists + `RUNTIME_TOOL_CATALOG`); `packageTypeEnum` drops `tool`; agent manifest drops `dependencies.tools`, gains `runtimeTools`; `toolManifestSchema` + `tool-bundler` removed; schema generator stops emitting `tool.schema.json`.

**Runtime (`runner-pi`):** the 5 tools ported to built-in Pi `ExtensionFactory`s under `runtime-tools/builtin/`, selected by `runtimeTools`; `prepareBundleForPi` registers built-ins instead of loading tool packages (dead scratch-dir machinery removed).

**API (`apps/api`):** tool resolution / `tool-bundle-cache` / `PUT /tools` removed; boot migration rewrites agent `dependencies.tools` → `runtimeTools` (fail-loud on unknown third-party tool ids); library route + OpenAPI group by `agent`/`skill`/`provider`/`integration`; `baseline.json` regenerated.

**Web (`apps/web`):** agent editor gains a `runtimeTools` checklist; tool type + tools page removed; i18n fr/en.

**System packages + infra:** 6 `tool-*.afps` deleted; agent manifests drop `dependencies.tools`; Dockerfiles + build/verify scripts updated.

`afps-runtime` keeps its own AFPS-spec-driven tool support (unchanged) — the platform simply stops feeding it tool deps.

**Cross-repo follow-up (separate PRs, out of scope here):** `registry/` (db enum + migration, publish/discovery/search) and `afps-spec/` (schema + spec) still carry the `tool` type; appstrate compiles against the published `@afps-spec/schema` unchanged.

**Quality gate:** `bun run check` green (20/20). Tests: core 466 · runner-pi 161 · api unit 841 · api integration 1455 · web 78 — all pass.

---

# Provider → Integration unification

After Phase 3, the branch absorbed the full **provider → integration unification** chantier: the legacy `provider` AFPS package type, the `provider_call` agent tool, and the connection-profile / app-profile features are all removed, and the 59 first-party providers are re-modelled as `api_call` integrations. Net effect across this range: **608 files changed, +18.5k / −61.1k** (the large deletion count reflects retiring two parallel feature surfaces).

> No production data exists, so every destructive migration in this range is safe to merge as-is — there is no backfill and no compatibility shim.

## Phase 4 — `api_call` capability + generic tool + provider conversion

- `@appstrate/core` integration manifest gains an `apiCall` capability — declarative HTTP tool surface for an integration's server.
- Generic `api_call` runtime tool replaces the bespoke per-provider tool machinery; `api_upload` covers resumable uploads (S3 multipart, MS / Google resumable, TUS) through the same `ctx.apiCall` bridge.
- `afps-runtime` gains a local + remote `api_call` resolver (`http-call-core.ts`); the CLI `appstrate run` path is wired onto it.
- **59 first-party providers converted** to `apiCall` integration packages in `system-packages/`.
- Provider machinery removed from `runner-pi`, `runtime-pi`, `apps/web`, and the CLI (the old `provider_call` path retired).

## Phase 5 — Drop the `provider` package type + tables

- `packageType` enum drops `provider`; `@appstrate/core` + `@appstrate/shared-types` provider types removed.
- Provider data tables dropped (destructive migration, safe — no prod data).
- Credential proxy re-platformed onto `integration_connections` (no more provider-specific credential path).
- `CLAUDE.md` + docs updated to drop the provider package type.

## Stage D — Remove connection-profiles + app-profiles

- The connection-profile and app-profile features are removed end-to-end: feature code, OpenAPI endpoints (unwired + verify-openapi), and the dead CLI commands.
- `X-Connection-Profile-Id` (the live connection-pinning header) is **retained** — it is a distinct, in-use mechanism, not part of the removed profile feature.

## `api_call` as `server.type` + in-process McpHost

- `api_call` integrations are modelled as `server.type: "api_call"` — a serverless tool surface (no subprocess), resolved by the spawn-resolver into a serverless `apiCall` spec.
- The sidecar routes `api_call` through `McpHost` as a trusted in-process tool factory, collapsing the previously divergent serverless path onto the same host registration the subprocess integrations use.
- `active-integration` gating + scope-only connect: only activated integrations expose tools, and the connect flow requests scopes only (no profile coupling).

## Audit-driven cleanup (final pass)

A 5-category architectural audit (indirection layers, duplicated types, duplicated business logic, dead/legacy code, inconsistent naming) drove a focused cleanup of **15 findings** — **112 files, +838 / −2.85k**:

- **Dead code removed:** `Dependencies.providers` + manifest provider parse/write helpers, the dead `provider` branch in `buildDependencies`, dead connection mutations (`useConnect*`), `invalidation.ts`, the always-`null` `MeConnectionEntry.profile`, the dead OAuth `connectionProfileId`, and the empty `providers`/`tokens` RunPlan fields.
- **Single source of truth:** `buildProxyCredentialsPayload` / `PROXY_INJECTED_FIELD` hoisted into `@appstrate/connect` (was duplicated across the integration resolver and the sidecar).
- **Naming unification:** integration-scoped `provider*` identifiers renamed to `api_call` / `integration` throughout (symbols, events `provider.called` → `api_call.called`, header `X-Provider` → `X-Integration`, sidecar constants, `ctx.providerCall` → `ctx.apiCall`, `provider-tool.ts` → `http-call-core.ts`). Legitimate `provider` naming (OAuth / model providers, better-auth, the live `X-Connection-Profile-Id`, and the `appstrate://provider-response/` wire URI) is intentionally left untouched.

**Out of scope (separate repos, tracked as #252 / #245):** the `afps-spec` mirror of the canonical event rename (`provider.*` → `api_call.*`) + the `integrationId` payload field, and the registry-side mirror.

## Updated quality gate

- `bun run check` — **20/20** turbo tasks green throughout the unification + cleanup ranges.
- Full test sweep green: `packages` + `runtime-pi` (1968) + api unit (774) + the broader suites — **0 failures**.
- Residual-provider grep clean: every removed-mechanism token (`provider_call`, `providerCall`, `X-Provider`, `makeProviderTool`, `provider.called`, `PROVIDER_CALL`) is eliminated; only legitimate / live / historical references remain.

---

# Post-unification — connect acquisition, runtime-tools-as-MCP, DB hardening & review hardening

After the provider→integration unification, the branch absorbed five more workstreams. Net effect of this range: **432 files changed, +14.8k / −8.7k across 26 commits**. Still no production data — every destructive migration here is safe to merge as-is.

## Unified credential acquisition — `ConnectStrategy` (#487)

All credential acquisition (OAuth2-PKCE, field paste, declarative login, code-orchestrated login) and all re-acquisition (refresh) now sit behind one `ConnectStrategy` abstraction with a single write site into `integration_connections`. Per `docs/architecture/CONNECT_ACQUISITION_SPEC.md` (Validé).

- **OAuth2Strategy** — authorization-code + PKCE, automatic proactive + on-401 refresh.
- **FieldsStrategy** — paste-the-bag for `api_key` / `basic` / bare `custom`.
- **LoginStrategy** — declarative single stateless login request (`connect.steps`, length 1).
- **LoginSecretStrategy** (`connect.tool` + `runAt:"run-start"`) — stores the user's login secret; the session is minted fresh inside each agent run's sidecar (MITM input substitution + `runConnectLogin` primitive), with `reauthOn` driving mid-run re-login on upstream 401.
- **OrchestratedStrategy** (`connect.tool` + `runAt:"link"`) — runs the untrusted login tool once in an **ephemeral connect-run** (stripped sidecar, per-run CA, MITM) via `ConnectToolExecutor`, captures the credential bundle, tears down.
- **Kijiji** migrated end-to-end to a `connect.tool` integration and validated against the live service.
- Legacy "profile" terminology dropped; `connect.dependsOn` (cross-integration login) deliberately **deferred**.

## Runtime tools as MCP + opt-in `output` (#501)

- `output` / `log` / `note` / `pin` / `report` are now transport-neutral MCP tool definitions hosted by the sidecar (and registered as Pi extensions on the no-sidecar path) — no longer in-prompt tool listings.
- `output` is **opt-in** per agent (`runtimeTools`) instead of always-injected; required only when the agent declares an `output.schema`.
- The prompt now forbids free-text replies and routes all user communication through tools; dead in-prompt tool-listing plumbing removed.

## `server.type: "bun"` in-process MCP servers (#497)

Bun integrations run as a host subprocess in process mode and in a dedicated runner container in docker mode — same MCP wire, isolation boundary preserved (no third-party bun in the root-privileged sidecar process tree).

## Hard-fail boot + DB hardening

- A declared integration that fails to start now **aborts the run** on every tier, with an `IntegrationBootReport` (timed breadcrumbs) relayed into run logs — replacing silent degradation.
- Migrations squashed into a single baseline; `timestamptz` for all timestamp columns; missing hot-path indexes added, redundant ones dropped; closed-domain + FK invariants enforced.

## PR #481 multi-agent review hardening (#492, #494, #496, #498, #499, #500, #504, #506)

Several deep review passes from the multi-agent audit:

- OAuth credential bug fix, proxy redirect credential-leak hardening, SSRF on userinfo + per-hop redirects, connection upsert races.
- Connection credentials validated against the manifest schema at connect time (#504).
- **Run-event forgery closed (#506)** — third-party MCP tools can no longer forge canonical runtime events (`output.emitted`, `log.written`, …): the re-emit is gated to first-party runtime tools at the agent host, the sidecar strips the forged meta key, and core allowlists canonical event types. Plus dead schema/plumbing removal (unwired OAuth Mode-B discovery, `extensionWrapper`/`OUTPUT_SCHEMA` dances, no-op `installSchema`).
- Iterative dead-code / provider-residue removal across passes.

## Phase 7 — remove `oauth1` auth type (breaking) + connect authoring guide (#507)

- `oauth1` removed from the integration auth-type enum (declarable-but-unconnectable: no connect strategy, signing never implemented; 0 usage, 0 prod data). OpenAPI breaking-change baseline refreshed to record it as intentional. `manifestVersion` not bumped (mirrors MCPB `1.x`).
- New author guide `docs/guides/writing-an-integration-with-connect.md` mapping each manifest auth declaration to its connect strategy.
- `afps-spec` keeps `oauth1` in its legacy provider schema (versioned independently) — appstrate-only by design.

## Cross-repo follow-ups still open (separate repos)

- `afps-spec/` — canonical event rename (`provider.*` → `api_call.*`) + `integrationId` payload field; legacy `tool` / `provider` / `oauth1` provider-schema surface (tracked #245 / #252).
- `registry/` — db enum + migration, publish/discovery/search mirror.

## Quality gate (current HEAD)

- `bun run check` — **20/20** turbo tasks green (tsc + eslint + prettier + verify-openapi).
- `detect:breaking` passes against the refreshed baseline.
- Test sweep green across core / connect / runner-pi / api unit + integration suites — 0 failures.

## System-package hygiene + full verification (#508)

Test-only / synthetic packages that were added during development but shipped as system packages at boot (the build globs every dir under `scripts/system-packages/`) were moved out of the versioned tree. System packages: **69 → 62** — every remaining one is a declarative real integration (the 59 converted providers + the official remote-MCP integrations `gmail-mcp@2.0.0`, `github-mcp`, `clickup-mcp`).

- **Moved out** (→ gitignored `local-test-packages/`): `agent-dice-game`, `agent-github-mcp-test`, `integration-dice`, `integration-kijiji`, `integration-mitm-test`, `integration-gmail-mcp@1.0.0`.
- **`bun-toolkit`** (the `server.type:"bun"` e2e fixture) was moved to a **versioned test fixture** (`runtime-pi/sidecar/test/fixtures/bun-toolkit/`) — committed so its e2e runs in CI, but outside `scripts/system-packages/` so it never ships.
- Dropped dead `connection-profiles` seeding from the e2e schedule-isolation tests (the feature was removed in Stage D; the helper still POSTed the deleted endpoint).

**Full verification (branch tip):** `bun run check` 20/20 · `bun test` 5497 pass / 0 fail · Playwright e2e 91 pass / 0 fail · `detect:breaking` clean · build deterministic (62 packages).

## Legacy cleanup (P7) — closed as no-op

The post-stabilization cleanup task (P7) targeted the residual provider-profile entities — tables `userAgentProviderProfiles` / `userApplicationProfiles` / `appProfileProviderBindings`, the `appProfileId` column, and the `app-profiles` routes. These were already removed earlier in this PR by **Phase 4** (drop provider tables) and **Stage D / D-bis** (remove the connection-profiles + app-profiles feature and the CLI profile commands). Verified **0 references** remain on this branch — the only surviving `provider_call` hits are regression tests asserting it is never emitted, and the `"provider"` hits are unrelated OIDC social-providers. Nothing left to clean here; the only open follow-ups are the cross-repo ones above (#245 / #252).

---

# AFPS 2.0 migration (originally PR #512, squashed into this branch)

Final breaking-change layer on top of the integrations rollout: migrates the platform from **AFPS 1.x → AFPS 2.0** end-to-end. The wire format flips from camelCase to snake_case across every transport (HTTP JSON, manifest files, OpenAPI, SQL columns), the integration manifest grows a `source` discriminant (`local | remote | api`), MCP servers become a first-class `mcp-server` package type (verbatim **MCPB manifest**), and the legacy `tool` / `provider` AFPS types + `TOOL.md` / `PROVIDER.md` companions are fully removed (no back-compat — no production data exists).

**Scope:** every workspace package + the 62 system-package manifests. Migration commit + 28 follow-up commits.

## Casing policy (Zone 1–5 + 13 carve-outs)

The 1.x→2.0 + DTO unification produced a strict, documented casing policy:

- **Wire** (HTTP JSON, AFPS manifests, OpenAPI, OAuth2 fields, SQL columns) → **snake_case**
- **Drizzle TS schema fields** → **camelCase TS / snake_case SQL** (`userId: text("user_id")`)
- **TS internal** (vars, function args, React props, Zustand state) → **camelCase**
- **Universal DB-convention carve-out** (`id`, `*Id`, `createdAt`, `updatedAt`, `notifiedAt`, `readAt`, `runNumber`, `runOrigin`, `contextSnapshot`, `modelCredentialId`) → stay **camelCase EVERYWHERE**
- **Better Auth tables** + 11 other named carve-outs (CloudEvents, BullMQ jobs, logger fields, audit log JSONB shapes, etc.) preserve camelCase with documented justifications

Authoritative reference: `docs/CASING_CONVENTIONS.md`. Automated 100%-coverage audit: `/audit-casing` (dispatches 6 parallel opus agents).

## Niveau-2 scope model end-to-end (Phases 0–7)

Per-tool scope inference + runtime enforcement, all 7 phases landed and individually tested:

| Phase | Surface | Test surface |
|-------|---------|--------------|
| 0 | Schema (`auths.{key}.availableScopes` + `tools.{name}.{requiredScopes, requiredAuthKey, urlPatterns}` cross-validation) | `packages/core/test/integration.test.ts` (+13 cases) |
| 1 | Install-time scope validation (refuse unknown tool / scope outside catalog) | `apps/api/test/integration/routes/packages.test.ts` (+6 cases) |
| 2 | OAuth kickoff scope union (`defaults ∪ caller ∪ getCurrentGrantedScopes`, incremental consent preserved) | `services/integration-scope-resolver.test.ts` (+9) |
| 3 | Runtime `tools/list` allowlist (`McpHost` filter + `tool_excluded_by_allowlist` audit) | `sidecar/test/mcp-host.test.ts` (+5) |
| 4 | Runtime MITM URL envelope (per-request enforcement BEFORE credential injection) | `sidecar/test/integration-mitm-listener.test.ts` (+5) |
| 5 | UI inline missing-scopes badge + "Reconnect to grant" CTA | wire-format covered server-side |
| 5b | Agent-editor `IntegrationToolPicker` (tool + scope checkboxes) | `agent-editor/test/utils.test.ts` (+7) |
| 6 | Refresh-time scope-shrink (`shrinkDetected` → flip `needsReconnection` only when below required floor) | `services/integration-token-refresh.test.ts` (+4) |
| 7 | Remote HTTP MCP (`server.type: "http"` — Bearer injection via `fetch` wrapper, force-refresh on 401) | `sidecar/test/integrations-boot-remote-http.test.ts` (+5) |

## Audit hardening (pass-1 → pass-3)

Three audit passes drove 70+ regression-prevention tests + 1 real code fix into the merged branch:

**Pass-1 — casing audit (`/audit-casing` dispatches 6 parallel opus agents across schema / wire DTO / Drizzle / frontend / carve-outs / cross-repo):**
- Caught 4 real user-visible bugs: pin-connection 400 (frontend sent camelCase, backend Zod demanded snake_case), unread-run badge always-on (missing `readAt == null` guard), schedule next-run preview always visible (frontend read `startedAt` but wire field is `started_at`), OIDC instance-client `redirect_uri=undefined` (interface declared `callback_url` but backend ships `callbackUrl`).
- Tightened `RunWireDto.status: string` → `RunStatus` union (proper architectural fix, unblocked `ACTIVE_RUN_STATUSES.has()` narrowing across every consumer).
- Hooks lying about types: `usePaginatedRuns`, `useScheduleRuns`, `useGlobalRunSync` retyped from Drizzle `Run` to wire `EnrichedRun`.

**Pass-2 — test coverage audit (10 parallel opus agents across schema / sidecar credentials / backend services / backend routes / afps-runtime / runner-pi / connect / niveau-2 scopes / web / system-packages):**
- Found 1 real code regression: `extractDependencies` silently ignored `dependencies.mcp_servers` (AFPS 2.0's first-class dep map — every downstream consumer was blind to mcp-server transitive deps). Fixed + tested.
- Identified 17 test gaps; closed 11 P0/P1 ones inline.

**Pass-3 — gap closure (+70 net new tests across security-critical surfaces):**

| Surface | Tests added | Why |
|---------|-------------|-----|
| `/internal/integration-credentials/*` GET + refresh | +10 | Most credential-sensitive HTTP endpoint in platform (sidecar live-decrypted creds). Dep-and-install gate had zero HTTP coverage. |
| `AES-GCM tamper detection` (`packages/connect`) | +4 | Byte-flip in body / authTag / IV → decrypt MUST throw. Pin the `setAuthTag` invariant. |
| `KID_PATTERN` wire-parse | +1 | Forged kid with whitespace / >32 chars / empty → reject before key lookup. |
| AFPS 1.x camelCase manifest rejection | +2 | The migration's headline contract was unpinned. |
| `/api/me/integration-pins` HTTP | +12 | snake_case body validation, end-user impersonation 401 gates, GET empty-list short-circuit. |
| `POST /run` 412 `missing_integration_connection` envelope | +4 | Contract `MissingConnectionsModal` consumes — was tested at service layer only. |
| OAuth callback CSRF state-mismatch | +2 | `OAuthCallbackError("transient")` branch fires when state never persisted. |
| Admin agent-resolution + org-pin HTTP | +15 | Integration detail page's admin surface — service-only coverage before. |
| `syncSystemPackagesToDb` boot DB sync | +5 | Single most load-bearing code path of the AFPS 2.0 migration. Insert / idempotent re-boot / integrity-drift safety gate / multi-version / multi-package independence. |
| Manifest validator (mcp_servers extraction) | +5 | New AFPS 2.0 dep category, cycle detection through mcp-server hop. |

Net: **5497 → ~5770 tests, 0 regressions**, 4 P0 + 7 P1 audit findings closed, 1 real code fix shipped (`mcp_servers` extraction).

## Final quality gate (branch tip)

- `bun run check` — 20/20 turbo tasks green (typecheck + lint + prettier + verify-openapi + detect:breaking + build:system-packages)
- `bun test` — 5743 pass / 2 fail (both env-bound python+openssl+Docker e2e tests, pre-existing, unrelated to this branch)
- Playwright e2e — 91 pass / 0 fail
- Cross-repo follow-ups still open (separate repos): `afps-spec/` canonical event rename + legacy schema cleanup (#245 / #252), `registry/` AFPS 2.0 alignment (separate `feat/afps-2.0` branch)

---

# Post-AFPS-2.0 stabilization (since #512)

33 PRs landed on top of the AFPS 2.0 squash — hardening, UX, dead-code removal, and a few additive features. No new breaking changes beyond those already documented above; everything below stays within the integrations rollout scope.

## Features (additive)

- **#525** — per-run shared workspace volume + MCP Roots for opt-in mcp-servers (SOTA filesystem pattern).
- **#550** — MCP-spec **auto-DCR** (RFC 7591 + PKCE + public client) for remote MCP integrations; no hand-registered OAuth client needed.
- **#551** — integrations tab in the package library UI.
- **#554** — system-package **conformance harness** (`scripts/conformance/`): spawn each local MCP server, `tools/list`, strict-diff against the manifest; gate tier runs in `bun run check`.
- **#540** — edit/delete packages regardless of scope (gated on org ownership, not scope identity).
- **#509 / #526 / #530** — integration detail page + connection naming unified; connection-renewal flow (modal, picker, live readiness badge); reusable connection picker extracted.
- `a6bef121f` surface resolved connections in run info · `5a8f9ef54` non-blocking connection-readiness badge on the run button · `8432063be` derive Google connection labels from account email.

## Fixes

- **#558** — agent self-provisions its workspace bundle instead of relying on volume seeding → fixes skills never materialising on tmpfs-backed self-hosted Docker (#548/#549).
- **#569** — stream input documents without the `Bun.write(Response)` 100% CPU spin that tripped the 60s heartbeat watchdog.
- **#564** — route agent egress through the proxy on the no-sidecar fast path (proxy was silently ignored).
- **#542** — docker-mode integration runtime: `delivery.env` local-runner egress gate + workspace seeding.
- **#557** — accept `integrations_configuration.<id>.tools: "*"` wildcard via `@afps-spec/schema` 0.6.0 (#547).
- **#556** — `warn` (not `info`) when a serverless integration exposes 0 tools (#545).
- **#534** — wipe per-run CA private key on teardown + race-free leaf serials.
- **#532** — stop `X-Integration-Id` leaking upstream + dedup helpers.
- **#533** — org-settings membership gate, AJV validator cache, guard post-finalize side-data.

## DX / tooling / quality

- **#560 / #563** — bring `runtime-pi` + `sidecar` into the typecheck gate (~103 inherited errors cleared + 2 genuine source bugs); `bun run check` now covers the Docker-image runtime code.
- **#562** — green resolver suite + silence two benign Better Auth warnings.
- **#553** — installer: don't create `~/.bash_profile` when only `~/.bashrc` exists (#527).
- **#502** — declare the `@appstrate/core` workspace dependency in runner-pi.
- `a8cc0d8d0` refresh LiteLLM pricing catalog (#552).

## Refactor / dead-code

- **#522** architectural simplification (dedup, dead code, shared engines) · **#541** architecture analysis cleanup (28 findings) · **#528** drop AFPS 1.x legacy dependency vocabulary · **#529 / #531 / #535 / #536 / #537 / #539** credential-proxy tidy, doc refresh, `actorFilter()` owner predicates, `callAllHooks()` broadcast signup hooks, drop dead agent-side MCP runToken, branch-audit dead-code removal.

## Quality gate note

The "Final quality gate" numbers above predate this stabilization batch. `runtime-pi` + `sidecar` are now inside the typecheck gate (#560), and the resolver suite is green (#562) — re-run `bun run check` + `bun test` at the current tip (`738a9123f`) for fresh totals before merge.


pierrecabriere:
### Forensic review pass — follow-up + merge-time caveat

Latest review pass landed in `3f839b02..8f5d1f85`: dead-code/naming residue purge, OAuth-popup timeout UX fix, dropped tautological tests, **+39 new security tests** (token-exchange/refresh classification, credential-decrypt plane isolation, PKCE vector, sidecar zip-slip guard, credential-proxy route authz incl. session-principal rebind), and removal of dead barrel re-exports + 13 unused i18n keys.

⚠️ **Merge-time caveat (cross-repo):** this PR deletes the published `@appstrate/core/tool-bundler` subpath, which `registry` still imports (`publish.ts`). Do **not** bump `@appstrate/core` in `registry` until that's migrated. Tracked in #511.

Note: the earlier-proposed `requirePermission("integrations","disconnect")` guard on `DELETE /me/connections/:id` was **reverted** — `/me/*` routes are org-context-free so RBAC perms aren't resolved for session callers (it 403'd the dashboard). The route's actor-ownership filter is the correct authz model, consistent with the other `/me/*` destructive routes; `integrations:disconnect` remains valid as an OIDC/end-user grantable scope.

PR #503, opened 2026-05-24
fix(runtime-tools): integration tools can't forge run events (pr #501 review)

pierrecabriere:
Deep review pass on **PR #501** (runtime-tools-as-MCP). Targets `feat/runtime-tools-as-mcp` so the fixes fold into #501 before it merges. Also carries the local stamping commit `379fb7c4` that wasn't yet pushed to the PR branch.

## 🔴 CRITICAL — integration tools could forge canonical run events

PR #501 routed **every** namespaced integration tool result through `reEmitRuntimeToolEvents` in `runtime-pi/mcp/direct.ts`, re-emitting whatever the integration returned under `_meta["appstrate/events"]` into the run's event sink. A compromised/malicious integration MCP server could therefore forge:

- `pinned.set` → written to `package_persistence`, spliced into **future** runs' system prompts (persistent prompt injection), incl. `scope:"shared"` → cross-actor poisoning
- `output.emitted` → overrides `runs.result.output` (what webhooks/billing read)
- `report.appended` / `log.written` → audit-trail tampering

**New vector in this PR** — integration results never passed through any re-emit before. The no-sidecar path was unaffected.

### Fix — clean separation + defence-in-depth
- **`direct.ts`**: the five first-party runtime tools (`output/log/note/pin/report`) are now wired through a dedicated `buildRuntimeMcpToolFactories` that re-emits their `_meta` events (trusted — served in-process by the credential-isolated sidecar). The integration-tool path **no longer re-emits at all**. Selectable runtime-tool names join the claimed set so they take the first-party path. (Also removes the bogus `integration_tool.*` telemetry previously fired for runtime-tool calls.)
- **`mcp-host.ts`**: strips the `appstrate/events` `_meta` key from third-party `CallToolResult`s at the sidecar boundary, enforcing the boundary at the trust anchor independently of the agent-side gate.
- **Regression tests**: `mcp-direct` (first-party `log` re-emits; `evil__steal` forged `_meta` does not) + `mcp-host` (third-party `_meta` strip).

## 🟡 Cleanup (surgical, from the same review)
- **CLI `run.ts`**: removed the dead `OUTPUT_SCHEMA` env-var dance (+ restore) — the schema is now passed explicitly to `buildRuntimeToolExtensions`; fixed stale comments.
- **`bundle-extensions.ts`**: removed the now-dead `extensionWrapper` option and the always-empty `extensionFactories` return (the helper is skills-only); updated the two callers (entrypoint, CLI) + tests.
- **`output.test.ts`**: added the `output` description-contract coverage (mandatory-iff-schema) that the deleted `output-tool.test.ts` guarded.
- **`entrypoint.ts`**: fixed the stale "stdout-JSONL → bridge" comment (no-sidecar path emits straight to `bridgedSink`).
- **CLAUDE.md**: runtime tools are MCP defs in `@appstrate/core/runtime-tool-defs`, not baked-in builtins registered by `prepareBundleForPi`.

## Correctness review (separate pass) — clean
Event-shape parity, runId/timestamp stamping (the `379fb7c4` fix), single-emission, output validation, and selection plumbing all verified against the consuming reducer/finalize — no behavioral regressions.

## Tests / checks
- `bun run check` — green for all touched packages (tsc + eslint + prettier + OpenAPI). The only `check` failure is a prettier/tsc issue in unrelated **uncommitted** concurrent-session WIP files outside this change set — not included in this PR.
- Affected suites pass: `runtime-pi` mcp-direct (16) · output · sidecar mcp-host (26) · runner-pi (incl. updated bundle-extensions).


pierrecabriere:
Closing: the base `feat/runtime-tools-as-mcp` was merged into `feat/integrations` (#501) with substantial follow-on cleanup (in-prompt tool listing removed; `b934a0c4` independently fixed runId/timestamp stamping + hardened finalize). Rather than rebase this branch onto the moved base, the outstanding review findings — chiefly the unfixed CRITICAL event-forgery vector — are being re-captured as a GitHub issue against `feat/integrations`.

PR #541, opened 2026-05-29
refactor: architecture analysis cleanup (28 findings)

pierrecabriere:
## Summary

Implements all 28 confirmed findings from the Karpathy-guided architecture analysis (`claudedocs/ARCHITECTURE_ANALYSIS_2026-05-29.md`). Every change is surgical, scoped to one finding, and carries a verifiable success criterion. Grouped into 7 commits by subsystem.

Overall: the architecture was already healthy — these are localized fixes (layering inversions, copy-paste duplication, package hygiene) plus a few latent correctness bugs.

## Findings fixed by domain

**core-packages** (`72ed0e3`)
- Remove broken `./back-compat` export from `@appstrate/core` (pointed at a non-existent file → hard-crash for any importer).
- Drop phantom runtime deps from `@appstrate/connect` (dragged the db/env credential graph into a credential-isolated package).
- Move runtime Zod schemas/fns out of the types-only `@appstrate/shared-types` into `@appstrate/core`.

**proxy-credentials** (`6a1c541`)
- **Bug fix:** `capResponseBody` getter-cast meant `X-Truncated` was never set on capped non-streaming responses. Read truncated state after drain + integration test.
- Rename misnamed `env-builder.ts` → `run-context-builder.ts`.
- Extract `updateOAuthBlob()` for the two duplicated OAuth round-trips.
- `token-resolver`: acquire lock via infra `getCache().set({nx})` instead of raw redis.
- Move cookie-jar store into `infra/cookie-jar/` (backed by `KeyValueCache`).
- Drop dead `cost_usd` column + downtime-safe migration.

**run-lifecycle** (`d8a6257`)
- Move `computeRunCost` to `state/runs.ts`; move `executeAgentInBackground` out of routes into `run-launcher/` (breaks a route↔service import cycle).
- Extract `applySpecToSidecarEnv()` (deduped across both orchestrators) and `refreshSequence()` (3 duplicated blocks).
- `listOrphanRunIds`: exclude runs a sibling instance is actively heartbeating.
- ⚠️ **Behavioral:** remote-origin `run.started` now fires at first runner event (matching the real `pending→running` transition) instead of at credential-mint time. Downstream `github-action` consumes `run.started` (separate repo) — see note below.
- Guard `runRowToWireDto` against drift with assignability assertions.

**packages** (`28701be`)
- `looksLikeAfpsBundle`: read ZIP central directory via fflate `unzipSync({filter})` instead of a 64 KiB byte-scan that false-negatives on large first entries.
- `package-fork`: read `packages.type` via one query instead of probing 4 types (~8 redundant queries).
- Replace `_set/_resetSystemPackagesForTesting` global mutators with optional params.
- **Bug fix:** import atomicity — `createPackageVersion` now throws on genuine DB failure instead of committing an un-runnable `version_id:null` package; ZIP cleanup fires on failure. (Single-package abort is the accepted trade-off; full all-or-nothing import txn deferred.)

**integrations** (`0acada2`)
- Remove dead `connection_blocked_by_admin` error code.
- Token refresh: write snake_case keys only; drop dead camelCase aliases.
- `resolveAgentIntegrationPick`: carry resolved `source` on the error so the `insufficient_scopes` branch stops re-fetching pins.

**infra-lib** (`79aed68`)
- **Bug fix:** LocalQueue cron used a hand-rolled 5-field parser, so `@daily`/`MON` passed validation but never fired on Tier-0. Replace with `computeNextRun` (cron-parser).
- Move `syncSystemPackagesToDb` + helpers out of `lib/boot.ts` into `services/system-packages.ts`.

**web** (`d642177`)
- **Bug fix:** `saveDraft` missing `version-info` invalidation (stale version banner).
- **Bug fix:** `CORE_APP_SCOPED_KEYS` dead key `agent-memories` → `agent-persistence` (agent-memory leaked across app switches).
- Extract untested manual SSE parser into a pure `parseSSEFrames()` + unit tests.

## Verification

- `bun run check` (tsc + eslint + prettier + OpenAPI + breaking-change detection, all 23 turbo tasks) — green. 2 pre-existing lint warnings only (react-refresh, untouched files).
- Full Tier-0 suite: **5970 pass, 0 fail, 72 skip** (6042 tests / 464 files).

## Behavioral changes to review

1. **Remote `run.started` webhook timing** (run-lifecycle): now fires at first runner event, not at run creation. `github-action`'s `src/stream.ts` consumes this event — confirm it doesn't treat `run.started` as acceptance confirmation before merge.
2. **Multi-package bundle import** (packages): a single failing package now aborts the import (was a logged partial-failure). Eliminates the worse un-runnable-orphan state.

## Rejected findings

18 analysis findings were rejected by adversarial verification (cosmetic, speculative, or based on misread code) and intentionally **not** implemented. See the appendix of `claudedocs/ARCHITECTURE_ANALYSIS_2026-05-29.md`.


pierrecabriere:
## Opus deep-audit — verdict + follow-up fixes

Ran a multi-agent Opus audit of the whole branch (per-domain review + cross-cutting dead-code/over-engineering sweeps + adversarial verification of every finding).

**Verdict: net improvement — cleaner, more coherent, more robust than base. No serious over-engineering, no harmful dead code, no functional regressions.** 15 PR-introduced findings (1 medium, rest low/nit), 3 pre-existing (out of scope), 2 rejected.

### Fixed in `04b500f5b`
- **medium** — LocalQueue cron catch-up was unbounded: a long process freeze could enqueue a large synchronous burst on the in-memory (Redis-absent) fallback. Now capped per poll + backlog coalesced with a warn.
- **low** — bundle orphan-row cleanup was a non-atomic SELECT-then-DELETE (TOCTOU vs a concurrent same-id import). Now a single self-guarding `DELETE … NOT EXISTS(version)`.
- **low/nit** — dropped tautological drift-guard assertions (the `: RunWireDto` return annotation is the real guard); `import type { z }` in shared-types; corrected an `updateOAuthBlob` docstring.

### ⚠️ Behavior changes bundled in this "cleanup" (reviewers/consumers note)
1. **`X-Truncated` header** now actually fires on capped non-streaming credential-proxy responses (was a no-op). Non-streaming capped responses are now buffered (bounded by `max_response_bytes`).
2. **Remote `run.started` webhook** now fires at the real `pending→running` transition (first runner event), not at run creation. A remote run that produces zero events never emits `run.started`. Confirm `github-action`'s `src/stream.ts` tolerates terminal-without-started.
3. **Cron catch-up semantics** (Tier 0/1 in-memory queue): missed occurrences now fire (drift) up to a per-poll cap, then coalesce — vs base firing ≤1/poll.

### Out of scope (pre-existing, not introduced here)
- `LocalQueue` ignores `WorkerOptions.limiter` (amplifies the cron-burst class; the cap above mitigates).
- Dead re-export of `modelCostSchema`/`tokenUsageSchema` in `run-launcher/types.ts`.

PR #605, opened 2026-06-07
feat(integrations): per-URL TLS-client routing via curl/curl-impersonate (#403)

### LINKED ISSUE #403: feat: provider extension `x-tlsClientByUrl` to route specific URLs through `curl` for TLS-exotic upstreams

pierrecabriere:
Closes #403.

## What

Adds a `_meta["dev.appstrate/tls-client"]` vendor extension that lets an integration declare, **per upstream URL**, which HTTP client the sidecar uses: the default Bun/undici `fetch`, or `curl` — optionally `curl-impersonate` for browser TLS mimicry. This defeats upstreams (Cloudflare / Akamai Bot Manager, JA3/JA4 fingerprinting) that reject the Node-native ClientHello.

```jsonc
"_meta": {
  "dev.appstrate/tls-client": {
    "routes": [
      { "url_pattern": "https://*.amisgest.fr/**", "client": "curl", "impersonate": "chrome" },
      { "url_pattern": "https://api.exotic.com/open/**", "client": "undici" }
    ]
  }
}
```

First matching route wins; unmatched URLs keep the default path.

## SOTA alignment

Researched the current landscape before designing. Plain `curl` only differs from Node's fingerprint — it does **not** itself look like a browser (its OpenSSL JA3 is also detectable). The maintained SOTA is **curl-impersonate** (lexiforest fork): it replicates a browser's exact cipher suites, extension order, GREASE values, and HTTP/2 settings via `--impersonate <target>`. So the extension is designed around an opt-in `impersonate` target rather than assuming plain curl is enough — matching what actually gets past JA3/JA4 in 2026. (TLS fingerprint is one signal among IP reputation / JS challenges / rate — documented as a limitation, not a silver bullet.)

## How

- **core** (`@appstrate/core/integration`): `TLS_CLIENT_META_KEY`, `TlsClientRoute`, `getTlsClientRoutes()`, install-time `superRefine` validation. `tlsClientRoutes` added to `IntegrationSpawnSpec`.
- **resolver** (`integration-spawn-resolver.ts`): copies resolved routes onto the spawn spec; dropped for remote HTTP MCP (those requests don't traverse the MITM listener).
- **sidecar**:
  - `curl-runner.ts` — a `fetch`-shaped wrapper over the `curl` CLI (`Bun.spawn`, `-q`, `-D -`, `--compressed`, manual redirect, body via stdin, `--impersonate`, response cap + timeout).
  - `tls-client-router.ts` — first-match URL→client resolver returning a drop-in `fetch`. Reuses `matchesAuthorizedUriSpec` glob semantics.
  - Wired into the MITM listener's upstream `fetch` at boot. **Nothing else in the path changes** — credential injection, SSRF re-check, 401 retry, and response passthrough are untouched; only which client makes the listener↔upstream TLS handshake differs. No routes → default `fetch` returned unchanged (zero overhead).
- **Dockerfile**: `curl` added to the sidecar image. `curl-impersonate` stays opt-in (vendor a build, set `TLS_CLIENT_CURL_IMPERSONATE_BIN`).
- **docs**: `INTEGRATIONS_RUNTIME.md` section.

## Scope (v1)

Covers the **MITM listener path** — local integrations using `delivery.http` to a TLS-exotic provider (the architecture behind the named use cases: `petitspas-to-drive`/Amisgest, `fizz-to-drive`). The `api_call` proxy and the plain CONNECT egress listener still use the default client; both inject a `fetch` the same way, so extending routing to them is mechanical follow-up (noted in code + docs, not yet wired).

## Tests

- core: route resolution (snake→camel, ordering, malformed-entry skipping) + schema validation (all error paths).
- `curl-runner.test.ts`: argv construction + response parsing via a fake spawn, **plus** real-`curl` integration tests against a local server (GET/POST/headers/gzip/3xx-no-follow/404). Auto-skip when `curl` is absent.
- `tls-client-router.test.ts`: routing logic, first-match precedence, init passthrough, `onRoute` telemetry.

Quality gate: `bun run check` (tsc + eslint + prettier + OpenAPI, 29 tasks) green; new tests + full sidecar suite (562) + core integration suite (105) green.

## Note on the issue's wording

The issue predates the provider→integration migration and refers to "provider extension `x-tlsClientByUrl`". Implemented as an `_meta["dev.appstrate/..."]` reverse-DNS vendor extension to match the codebase's existing convention (`dev.appstrate/api`, `dev.appstrate/workspace`, `dev.appstrate/connect`).


pierrecabriere:
Closing — not the right mechanism to merge as-is.

Implementation works and is fully tested, but on review it's the **crude tier** of TLS impersonation, not SOTA:

- **Spawns the `curl` CLI per request** — no connection pooling, no HTTP/2 reuse, full-body buffering, process-spawn overhead. The SOTA approach is an in-process impersonating TLS stack (curl_cffi-style binding to libcurl-impersonate, or a small reqwest-impersonate / utls egress), not shelling out per request.
- **Doesn't impersonate out-of-the-box**: the image ships plain `curl` (curl-impersonate left opt-in/unbundled), so by default it only defeats a naive JA3-on-Node check, not real Cloudflare/Akamai.
- **Partial impersonation even with curl-impersonate**: the MITM rebuilds headers (lowercased/reordered) and appends injected `-H` after curl's browser defaults, so the header-order + HTTP/2 fingerprint may still not match a real browser.

More importantly, the underlying diagnosis is unverified — "curl works, fetch doesn't" is frequently NOT JA3 (it's HTTP/2 vs 1.1, header casing/order, missing UA, undici connection behaviour, compression). Shipping a curl-spawn subsystem inside the credential-isolating sidecar for 2 integrations, on an unverified TLS-fingerprint hypothesis, isn't justified.

The router abstraction (`makeTlsClientFetch`, a drop-in `fetch`) is reusable if we revisit. Branch `feat/tls-client-routing` kept. See #403 for the diagnosis-first path forward.

PR #606, opened 2026-06-07
feat(mcp): inbound MCP server exposing the platform API over /api/mcp

pierrecabriere:
## Summary

Adds a built-in **`mcp` module** that serves the entire platform REST API as a **Model Context Protocol server** (Streamable HTTP, stateless) at **`/api/mcp`**, derived from the live OpenAPI spec. Implements `claudedocs/MCP_SERVER_PLAN.md` (Phases 1–3 + the cheap parts of 4) and prepares the ground for the external chat app in `claudedocs/CHAT_APP_PLAN.md`.

## Design

- **Progressive disclosure, not 250 tools.** ~250 operations as individual MCP tools ≈ a 360K-token tool list — dead on arrival for every client. Instead, three fixed tools — `search_operations`, `describe_operation`, `invoke_operation` — let the model discover on demand. ([Anthropic: code execution with MCP](https://www.anthropic.com/engineering/code-execution-with-mcp), [advanced tool use](https://www.anthropic.com/engineering/advanced-tool-use))
- **Single source of truth.** The operation catalog is built lazily from `buildOpenApiSpec()` (the same assembly behind `GET /api/openapi.json`). New endpoints become invokable automatically; `verify-openapi`'s *code ⊆ spec* guarantee keeps it complete. The MCP server's own paths are excluded from the catalog (no recursive self-invocation).
- **In-process coupling.** `invoke_operation` dispatches through the platform's own Hono app (`app.fetch`) via a new generic `lib/platform-app.ts` holder, re-entering the **full auth pipeline + RBAC** with the caller's auth forwarded verbatim. An MCP call can never exceed what the same credential could do over REST: `mcp:invoke` gates the tool, the underlying operation still enforces its own permission (defence in depth).

## Chat groundwork (auth/scope)

- New `mcp:read` / `mcp:invoke` RBAC resource (read incl. `viewer`; invoke excl. `viewer`; both API-key- and end-user-grantable).
- `getAppstrateScopes()` now merges module end-user-grantable scopes into the OIDC requestable vocabulary, mirroring the end-user filter already in `claims.ts` — avoids hardcoding module scopes in the core-typed scope list (which broke `@appstrate/web`'s typecheck).
- **`llm-proxy:call` added to the `member` role** (and to the OIDC vocab) so ordinary members — not just admins — can run completions through the platform. ⚠️ **Reviewers: confirm this policy change.** It's deliberate (per the plan, to power a member-facing chat) but widens who can spend LLM credits.
- `mcp` added to the default `MODULES` set (+ all mirrors: env schema, registry, compose-defaults table, `.env.example`, CI).

## Discovery / spec compliance

- RFC 9728 Protected Resource Metadata served at `/.well-known/oauth-protected-resource`.
- The `WWW-Authenticate: … resource_metadata=` challenge on 401 (RFC 9728 §5.1) is a **documented follow-up** — the platform auth pipeline owns the 401 for `/api/*`. Not needed by the first-party chat (it authenticates via its OIDC token); only matters for tokenless generic clients (Claude Desktop), which is Phase 4.

## Out of scope (follow-ups)

- RFC 7591 Dynamic Client Registration / CIMD for zero-config generic MCP clients (Phase 4).
- Embeddings-backed search (keyword/tag ranking ships first; revisit if recall hurts).
- Token audience-binding (RFC 8707) — relevant when third-party clients arrive.

## Testing

- **Unit** (no DB): catalog build + exclusion, `search`/`describe`/`invoke`, in-process dispatch + auth-header forwarding, `mcp:invoke` gating, path interpolation, missing-param errors.
- **Integration**: public PRM, unauthenticated 401, authenticated `initialize` handshake through the real middleware chain.
- `bun run check` green (all 29 turbo tasks). Regression-checked: oidc (430), apps/api integration (1368), apps/api unit (834), llm-proxy (15), webhooks (61), core-providers (5), registry/loader. **0 failures.**

## Clarifications & review follow-ups (commit 50ff158b6)

**Endpoint shape.** The server is mounted **per-org at `/api/mcp/o/:org`**, not the bare `/api/mcp` implied by the summary. Each org's canonical resource URI (`${APP_URL}/api/mcp/o/<orgId>`) is the RFC 8707 audience a token is bound to; all `/api/mcp*` paths are excluded from the invokable catalog.

**Phase 4 is co-shipped, not incidental.** This PR intentionally bundles the auto-discovery program — better-auth `1.7.0-beta.4` major bump, `@better-auth/cimd`, RFC 7591 DCR, RFC 8707 `resources` columns (migration 0006) — on top of Phases 1–3, because the 1.7 oauth-provider is the mechanism that enforces cross-org audience confinement. It is a dependency, not scope creep, but it does widen the diff and the blast radius — review accordingly.

**RBAC widening (confirm).** `llm-proxy:call` is granted to the **`member`** role (and added to the OIDC vocab) so ordinary members can drive the member-facing chat. This widens who can spend LLM credits. On billed deployments, confirm cloud quota enforcement gates member-initiated calls.

**Default-on.** The `mcp` module is added to the default `MODULES` set (+ all mirrors). Self-hosters tracking the default get it without opting in.

### Production blocker fixed in this PR

Migration 0006's additive `resources` columns would be **silently skipped** on the production platform DB: drizzle's postgres-js migrator applies by `__drizzle_migrations` timestamp watermark, and prod's watermark is corrupted to a future date — so 0006's `when` falls below it and never runs, while the pinned better-auth 1.7 oauth-provider expects those columns at token-mint time.

Fix (50ff158b6):
- 0006 made idempotent (`ADD COLUMN IF NOT EXISTS`).
- `reconcileOAuthResourceColumns()` at boot detects the missing columns via `information_schema` and runs the additive DDL out-of-band of the watermark — a no-op on healthy DBs, a loud self-heal on the drifted prod DB (logs an error telling the operator to realign the watermark before the next schema release).
- Added the two zero-coverage tests review flagged: `invoke_operation` write-path (POST body serialization + content-type; GET sends no body) and the `orgIdFromMcpAudience` exact-match invariant (rejects nested/decorated/empty/wrong-prefix audiences).

> **Still recommended before/at deploy:** realign the prod `__drizzle_migrations` watermark so *future* migrations are not skipped too — the self-heal only covers 0006.


pierrecabriere:
## Fresh-eyes review round (2 independent reviewers) — fixes applied in af1b449

### Real defects found + fixed
- **HIGH — `invoke_operation` hang**: `response.text()` on a streaming (`text/event-stream`) operation never resolves. The platform exposes SSE GET ops (`streamRun`, run logs), so this was reachable through the tool surface and would hang the request promise with no timeout. `readResponse()` now refuses streaming bodies, summarises non-text/binary bodies (downloads/tarballs) without decoding, and caps buffered text at 100K chars with a truncation flag.
- **SOTA — tool annotations**: added `annotations` to all three tools (search/describe = `readOnlyHint`+`idempotentHint`; invoke = `destructiveHint`, non-idempotent, open-world) so clients prompt for confirmation correctly.
- **RFC 9728 discovery**: PRM now also served at the path-insertion variant `/.well-known/oauth-protected-resource/api/mcp` (§3.1) and advertises `scopes_supported`.
- **Docs/correctness**: `GET /api/mcp` documented as 405 (stateless → no GET stream); corrected the `enableDnsRebindingProtection` rationale (rests on auth, not CORS).

### Test gaps closed
- End-to-end through the transport: `tools/list` (3 tools + annotations), `tools/call search_operations`, and `invoke_operation` reaching a **real route in-process** (asserts a real HTTP status — proves dispatch end to end). This was the biggest gap — previously only `initialize` was exercised over HTTP.
- RBAC: API key without `mcp:read` → 403; with `mcp:read` but not `mcp:invoke` → tool `isError`.
- Regression guard: `member` role has `llm-proxy:call` (`permissions.test.ts`).

### Verified clean
- `bun run check` (29 turbo tasks) green. mcp module 17 tests, permissions 16, plus the prior regression sweep (oidc 430, apps/api integration 1368, unit 834) — 0 failures.

### Deliberately deferred (documented, not needed for the first-party chat)
- Large-result spill to MCP resources (`resources/list`+`read`) — stateless transport needs a content-addressed store; mitigated for now by the SSE-refusal + binary-omit + char cap.
- `WWW-Authenticate: … resource_metadata=` challenge on the 401 — the platform auth pipeline owns the `/api/*` 401; PRM is discoverable via both well-known URLs in the meantime.
- `structuredContent`/`outputSchema` on the read tools.

PR #645, opened 2026-06-09
feat(runs): return full Run DTO from trigger + document model pin discoverability

### LINKED ISSUE #635: Runs: org-default model change applies silently to in-flight testing — model pin at trigger exists (modelId) but is undiscoverable
## Problem

While testing an agent, the org default model was changed mid-session; the next triggered run picked up the new default (`model_source: "org"`) rather than the model visible at trigger time. The tester had to cancel the run and had no obvious way to pin the model.

The capability exists — `runAgent` accepts a `modelId` override (#544 touched its validation) — but it is not surfaced:

- Not prominent in the MCP `describe_operation` output / OpenAPI description for `runAgent`.
- No UI affordance at trigger time showing "this run will use model X (source: org default)" with a pin option.

## Proposal

1. **Docs/MCP:** call out `modelId` in the `runAgent` operation description as the way to pin a model per run; document `model_source` semantics (`org` = resolved at run start, not trigger time).
2. **Response transparency:** run object already exposes `model_source` — good; consider also echoing the resolved model id at trigger time in the `runAgent` response so the caller can detect drift immediately.
3. **UI (optional):** trigger dialog shows resolved model + source, one-click pin.

Mostly a discoverability/doc fix; no behavior change strictly required.

pierrecabriere:
## Summary

Fixes #635 — the org-default model is resolved at run creation, so changing it mid-session silently applies to the next triggered run. The `modelId` per-run pin already existed but was undiscoverable, and callers had no way to see the resolved model at trigger time.

Two parts: **(1)** make the pin discoverable (OpenAPI/MCP/UI), and **(2)** return the full run resource from the trigger so callers see the resolved model — and everything else — without a follow-up GET.

### Model resolution precedence (investigated, unchanged)

`routes/runs.ts` → `resolveRunPreflight` → `buildRunContext` → `resolveModel`:

1. request `modelId` (per-run pin, validated by `assertExplicitModelExists` → 404 on unknown)
2. agent model setting (`application_packages.model_id`)
3. org default model (`org_models.is_default`)
4. system default

`model_source` (`system` | `org`) is computed in `run-context-builder.ts` from the resolved model and persisted on the run row.

### Changes

**a) OpenAPI / MCP discoverability** — the `runAgent` operation description and its `modelId` body field now state the full precedence chain, that the org default is resolved at run creation (not ahead of time), the pin semantics, and the 404 contract. MCP `describe_operation` serves the raw OpenAPI operation node from `buildOpenApiSpec()`, so it picks this up with no MCP-side change (verified against the built spec). Also clarified `model_source` timing on the `Run` schema.

**b) Trigger returns the full Run DTO** — `POST /agents/:scope/:name/run` now returns the **created run resource**: the same shape and serializer as `GET /runs/:id` (`getRunFull`), so callers get `status`, `version_ref`, `agent_scope`, and the resolved `model_label` / `model_source` (org-default drift detection per #635) in one call, no follow-up GET. OpenAPI response is `allOf: [Run, { runId }]`; the legacy `runId` field is kept alongside the DTO's `id` for backward compat (web `navigate`, CLI, github-action read `runId`).

> Amended after review. The first pass echoed only `{ runId, model_label, model_source }` — an arbitrary cherry-pick: the same drift argument applies to `version_ref`, `agent_scope`, etc., and growing the response field-by-field is the incoherence #635 itself flags. Both values are already persisted on the run row at creation, so the principled move is to return the resource. This trigger endpoint is one of ~25 mutating endpoints returning a stub instead of the resource — uniform sweep tracked in #646.

**c) UI transparency** — the run-trigger modal (`RunModal`) shows the resolved model and its source before triggering: « Modèle : X (défaut de l'organisation) » / « (réglage de l'agent) », mirroring the server-side `resolveModel` cascade including the stale-pin fallback. No per-run model picker in the trigger modal today (one exists in the schedule form's `RunOverridesPanel`); per the issue this part is optional and display-only keeps the change minimal — pinning is available via the API `modelId` documented in (a).

**d) Tooling** — `detect-breaking-changes.ts` now flattens `allOf` in `getProperties` / `getRequired`. A top-level `allOf` response previously read as zero properties, so a `$ref`-composed resource response spuriously flagged every baseline field as removed. Also unblocks the #646 convention work, where create responses compose `allOf: [Resource, …]`.

### Tests

- Route-level integration tests covering the trigger response: org default resolved at trigger, explicit `modelId` pin beats the default, and the response is the full Run DTO (`id` === `runId`, `status`, `agent_scope`) matching the persisted run-row snapshot.
- First tests to take the trigger route to a 200; to keep the fire-and-forget kickoff away from a real container runtime (slow Docker pulls were racing the next test's `truncateAll`), they run against a fake orchestrator via a new `_setOrchestratorForTesting` seam (same pattern as `_setRunLimitsForTesting`), draining the in-flight tracker before finishing. `runs.test.ts` stable across consecutive runs.

### Validation

- `runs.test.ts`: 45 pass / 0 fail
- `bun run check` (turbo check + verify-openapi, 29/29 tasks) ✓
- `detect:breaking`: 0 breaking / 46 additive (full Run DTO fields now surfaced) + 1 unrelated additive
- Full API suite (first pass): 6395 pass / 0 fail. One pre-existing warning on `/api/mcp/o/{org}`, unrelated.


pierrecabriere:
## Amended after review — return the full Run DTO instead of a cherry-picked echo

The original echo added only `model_label` + `model_source` to the trigger response. On review that's an arbitrary cherry-pick: the same org-default-drift argument applies to `version_ref` (#641), `agent_scope`, etc., and growing the response field-by-field is exactly the incoherence this issue flags. Both values are already persisted on the run row at creation.

**Change**: `POST /agents/:scope/:name/run` now returns the **full Run DTO** — same shape and serializer as `GET /runs/:id` (`getRunFull`) — so callers get `status`, `version_ref`, `agent_scope`, and the resolved model in one call, no follow-up GET. Legacy `runId` kept alongside the DTO's `id` for backward compat (web `navigate`, CLI, github-action all read `runId`). OpenAPI response is now `allOf: [Run, { runId }]`.

**Tooling**: `detect-breaking-changes.ts` now flattens `allOf` in `getProperties`/`getRequired` — a top-level `allOf` response previously read as zero properties and spuriously flagged every baseline field as removed. This also unblocks the broader convention below.

The OpenAPI cascade docs + UI `ResolvedModelHint` from the first pass are unchanged — they remain the real discoverability fix for #635.

Verified: `runs.test.ts` 45 pass, `bun run check` 29/29, breaking detector = 0 breaking / 46 additive.

This trigger endpoint is one of ~25 mutating endpoints that return a stub instead of the resource — tracked for a uniform sweep in the convention issue below.

PR #659, opened 2026-06-10
refactor(api)!: DELETE → 204 + documented bulk counts (batch A, #657)

pierrecabriere:
Batch **A** of #657 — trivial 204s + documented bulk counts. Every change below is a deliberate breaking wire change; the OpenAPI breaking-changes baseline is updated in this PR.

## DELETE / ack endpoints → `204 No Content`

| Endpoint | Removed body |
| --- | --- |
| `DELETE /api/orgs/{orgId}` | `{ ok }` |
| `DELETE /api/orgs/{orgId}/members/{userId}` | `{ ok }` |
| `DELETE /api/orgs/{orgId}/invitations/{invitationId}` | `{ ok }` |
| `DELETE /api/schedules/{id}` | `{ ok }` |
| `DELETE /api/integrations/{packageId}/oauth-clients/{authKey}` | `{ deleted }` |
| `DELETE /api/integrations/{packageId}/pins/{agentPackageId}` | `{ deleted }` (idempotent — 204 whether the pin existed or not) |
| `DELETE /api/integrations/{packageId}/default` | `{ deleted }` (idempotent) |
| `DELETE /api/agents/{scope}/{name}/persistence/memories/{id}` | `{ deleted }` |
| `DELETE /api/agents/{scope}/{name}/persistence/pinned/{id}` | `{ deleted }` |
| `PUT /api/notifications/read/{runId}` | `{ ok }` (idempotent ack) |
| `POST /api/welcome/setup` | `{ ok }` (profile lives under `GET /api/profile`) |

## Bulk mutations → stay `200` with documented snake_case counts

| Endpoint | Change |
| --- | --- |
| `DELETE /api/agents/{scope}/{name}/runs` | `{ deleted }` → `{ deleted_count }` (required, documented) |
| `PUT /api/notifications/read-all` | `{ updated }` → `{ updated_count }` (required, documented) |
| `DELETE /api/agents/{scope}/{name}/persistence` | keeps `{ memories_deleted, checkpoint_deleted }` — now `required` + field descriptions |

## Mechanics

- OpenAPI paths updated to match (204 responses, bulk-count schemas), `verify-openapi` green
- Breaking-changes baseline regenerated (`bun scripts/detect-breaking-changes.ts --update-baseline`)
- Tests **replaced** (not added) to assert the new shapes: agents, integrations, integrations-admin-pins, notifications, organizations, runs, schedules, welcome
[...]

Issue #657, opened 2026-06-10
REST strict: mutations return the bare resource everywhere — remove all legacy/compat fields (no prod data)


pierrecabriere:
## Context

Supersedes the additive posture of #646. There is **no production data**, so backward compatibility is not a constraint. Full audit (33/33 route files, `/audit-rest-payloads`, 2026-06-10) found ~44 logical non-conformant mutation endpoints (~56 counting the packages factory ×4-type expansion), including the compat shims deliberately kept by #645–#651.

## The strict rule

1. `POST` create → `201` + bare created resource (same serializer as GET detail, OpenAPI `$ref` direct — no `allOf` compat envelope)
2. `PUT`/`PATCH` → `200` + bare updated resource
3. `DELETE` → `204` empty
4. **Remove**: id aliases (`runId`, `packageId`, …), `{success}`/`{ok}` booleans, op scraps (`active`, `activated_at`, `lock_version`*, `message`, `warnings`, `restored_version`, `validation`, `deleted`), every `allOf:[Resource,{compat}]`
5. Exceptions (keep operation-result shape): one-time secrets, OAuth/flow init, presign envelope, multi-entity import reports, synthetic tests, transport passthroughs, browser auth flows, **bulk mutations** (decision 2026-06: `DELETE /runs`, `DELETE …/persistence`, `PUT /notifications/read-all` keep a documented `{deleted_count}`/`{updated_count}` — an action over a set is not a single resource)

\* if `lock_version` is genuinely part of the package resource (draft optimistic-lock state), it belongs INSIDE the resource DTO, not stapled on mutation responses.

## Fix batches (one PR each)

- [ ] **A — trivial 204s**: orgs DELETE org/invitation/member, schedules DELETE, integrations DELETE oauth-clients/pins/default + deactivate, agents persistence memories/pinned DELETE, notifications read/:runId → 204; bulk ops → documented `*_count` shapes
- [ ] **B — proxies + models + credentials**: POST/PUT `{id}` → bare resource; PUT default `{success}` → resource; mpc pair/redeem → credential resource
- [ ] **C — runs**: POST run → `201` bare Run (strip `runId`); cancel → updated Run; inline → Run DTO; sink/extend + runs-remote `runId` → `id`; sync web `use-mutations` (`data.runId` → `data.id`), CLI, github-action
[...]

PR #660, opened 2026-06-10
refactor(api)!: bare resources for proxies/models/credentials mutations (batch B, #657)

pierrecabriere:
Batch **B** of #657 — proxies, models, and model-provider-credentials mutating endpoints now return the **bare resource**: direct `$ref` to the resource schema, no `allOf` wrapper, no duplicated top-level `id`, no legacy `success` flag. POST returns **201**.

Refs #657

## Per-endpoint changes (breaking)

| Endpoint | Before | After |
| --- | --- | --- |
| `POST /api/proxies` | 201 `OrgProxy` + duplicated top-level `id` (allOf) | 201 bare `OrgProxy` |
| `PUT /api/proxies/:id` | 200 `OrgProxy` + duplicated top-level `id` (allOf) | 200 bare `OrgProxy`; `404` documented |
| `PUT /api/proxies/default` | 200 `{ success: true, proxy: OrgProxy \| null }` | 200 bare *effective* default `OrgProxy`; **204** when no default remains in effect |
| `POST /api/models` | 201 `OrgModel` (with `{ id }` stub fallback) | 201 bare `OrgModel` (re-projection failure → 500) |
| `PUT /api/models/:id` | 200 `OrgModel` (with `{ id }` stub fallback) | 200 bare `OrgModel`; `404` documented |
| `PUT /api/models/default` | 200 `OrgModel` allOf `{ success: true }`, or `{ success: true }` when cleared | 200 bare *effective* default `OrgModel`; **204** when no default remains in effect |
| `POST /api/model-provider-credentials` | 201 `ModelProviderCredentialInfo` (with `{ id }` fallback) | 201 bare non-secret `ModelProviderCredentialInfo` |
| `PUT /api/model-provider-credentials/:id` | 200 `ModelProviderCredentialInfo` (with `{ id }` fallback) | 200 bare non-secret `ModelProviderCredentialInfo`; `404` on vanished row |

### Removed fields

- `success` — `PUT /api/proxies/default`, `PUT /api/models/default`
- `proxy` envelope — `PUT /api/proxies/default` (resource is now the body itself)
- duplicated top-level `id` (allOf overlay) — `POST/PUT /api/proxies`, documented legacy `id`-stub fallbacks on models/credentials

### Secrets

Credential endpoints keep returning the non-secret `ModelProviderCredentialInfo` projection (same shape as GET/list) — API keys / OAuth tokens are **never** echoed back.

### pair/redeem decision

`POST /api/model-providers/oauth/pair/redeem` **keeps its operation-result shape** (`{ ok, credentialId, availableModelIds }`) as a documented flow-completion exception: the connect-helper's pairing bearer is single-use and consumed by this very request, so the caller cannot follow up with a GET, and `availableModelIds` is registry-derived operation info (the provider's featured models the helper prints in its terminal summary), not part of the credential resource. The dashboard obtains the created credential via `GET /pairing/{id}` polling + the credentials list. Decision documented in the route handler and the OpenAPI description.

## Implementation notes

- Web hooks (`use-proxies`, `use-models`, `use-model-provider-credentials`) typed against the returned bare resources, default-setter hooks handle the 204 (`undefined`) case.
- Integration tests replaced to assert bare shapes, 201/204 statuses, and the no-secret-echo invariant.
- Breaking-change baseline regenerated (`bun scripts/detect-breaking-changes.ts --update-baseline`) — also picks up additive drift from recently merged feature PRs (#638, #641, #642, #651). Will conflict with #659's baseline; regenerate after whichever merges second.

## Validation

- `bun run check` — 29/29 tasks, OpenAPI verification ALL CHECKS PASSED
- `bun scripts/detect-breaking-changes.ts` — PASSED against regenerated baseline
- Touched integration tests: proxies + models + model-provider-credentials (42 pass) and model-providers-oauth* (23 pass)

### DIFF
--- a/apps/api/src/openapi/baseline.json
+++ b/apps/api/src/openapi/baseline.json
@@ -796,7 +796,7 @@
         },
         "responses": {
           "200": {
-            "description": "Agent proxy updated",
+            "description": "Agent proxy updated — returns the affected proxy setting",
             "headers": {
               "Request-Id": {
                 "$ref": "#/components/headers/RequestId"
@@ -808,12 +808,28 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "type": "object",
-                  "properties": {
-                    "success": {
-                      "type": "boolean"
+                  "allOf": [
+                    {
+                      "type": "object",
+                      "required": ["proxyId", "resolved"],
+                      "properties": {
+                        "proxyId": {
+                          "type": ["string", "null"]
+                        },
+                        "resolved": {
+                          "type": "boolean"
+                        }
+                      }
+                    },
+                    {
+                      "type": "object",
+                      "properties": {
+                        "success": {
+                          "type": "boolean"
+                        }
+                      }
                     }
-                  }
+                  ]
                 }
               }
             }
@@ -1284,7 +1300,7 @@
         },
         "responses": {
           "200": {
-            "description": "Agent model updated",
+            "description": "Agent model updated — returns the affected model setting",
             "headers": {
               "Request-Id": {
                 "$ref": "#/components/headers/RequestId"
@@ -1296,12 +1312,25 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "type": "object",
-                  "properties": {
-                    "success": {
-                      "type": "boolean"
+                  "allOf": [
+                    {
+                      "type": "object",
+                      "required": ["modelId"],
+                      "properties": {
+                        "modelId": {
+                          "type": ["string", "null"]
+                        }
+                      }
+                    },
+                    {
+                      "type": "object",
+                      "properties": {
+                        "success": {
+                          "type": "boolean"
+                        }
+                      }
                     }
-                  }
+                  ]
                 }
               }
             }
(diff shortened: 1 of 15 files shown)


pierrecabriere:
## Review fix applied (5e53c79eb)

**Finding**: the credential-reachability gate (`loadInferenceCredentials`) only ran in the derive-label branch of `POST /models`. With an explicit `label`, a dead credential (needs-reconnection OAuth) skipped it → row inserted, then the bare-resource re-projection 500'd deterministically, leaving a phantom row filtered out of every list. Same root cause gave `PUT /models/:id` a misleading 404 after a successful write.

**Fix**: gate hoisted out of the branch — runs before the insert on POST and before the update when the body re-points `credentialId` on PUT → `400 invalid_request` before any write. The 500-on-re-projection path is now genuinely race-only.

**Tests**: 2 regressions added (POST dead credential → 400 + no row; PUT switch to dead credential → 400 + pointer unchanged). Suite: 15/15 pass (tier0 PGlite). `bun run check` 29/29.

PR #662, opened 2026-06-10
refactor(api)!: bare Run resource from run mutations — drop runId alias, 201 on trigger (batch C, #657)

pierrecabriere:
Refs #657 — batch **C (runs)** of the strict bare-resource rule. Deliberate **breaking wire changes** (no prod data).

## Endpoint changes

| Endpoint | Before | After |
|---|---|---|
| `POST /agents/{scope}/{name}/run` | `200` + `allOf:[Run, {runId}]` | **`201`** + bare `Run` resource (same serializer as `GET /runs/{id}`, direct `$ref`) |
| `POST /runs/{id}/cancel` | `200 {ok: true}` | `200` + the **updated Run resource**, read AFTER `synthesiseFinalize` — `status: cancelled`, final `cost`/`completed_at` |
| `POST /runs/inline` | `202 {runId, packageId}` | `202` + bare `Run` resource (the shadow package id is the resource's own `packageId` field) |
| `POST /runs/remote` | `201 {runId, ...sink credentials}` | envelope **kept** (one-time sink credentials are legit operation data, #657 exception) but `runId` → **`id`** |
| `PATCH /runs/{runId}/sink/extend` | `200 {runId, expiresAt}` | `200 {id, expiresAt}` — operation result, not a resource: `sink_expires_at` is deliberately not on the Run DTO, and no first-party client reads this body (runtime keep-alive is `POST /events/heartbeat`) |
| `POST /runs/inline/validate` | `{ok: true}` | unchanged — legit dry-run operation result (noted only) |

## Removed fields / status changes

- **Removed**: `runId` alias (trigger, inline, remote, sink/extend), `ok` boolean (cancel), the `allOf:[Run,{runId}]` compat envelope, the inline `{runId, packageId}` envelope.
- **Status codes**: trigger `200` → `201` (created). Inline stays `202` (accepted, async), cancel stays `200` (update), remote stays `201`.

## Consumers updated (same PR)

- **web**: `useRunAgent` — `api<EnrichedRun>`, `data.runId` → `data.id` (only `.runId` response read in apps/web).
- **CLI**: `remote-runner.ts` `triggerRun` parses the bare resource's `id` (strict, fail-fast on alternate keys, hints updated); `report.ts` `startReportSession` reads the remote envelope's `id`.
- **MCP**: server instructions no longer say "returns a runId" — now "returns the created run resource (use its `id`)".
- **github-action**: separate PR in `appstrate/github-action` (reads `data.runId`) — linked below.

## Tests replaced

- `runs.test.ts`: trigger echo tests expect `201`, `body.id`, `body.runId` undefined; cancel tests assert the bare updated Run (`id`, `status: cancelled`) instead of `{ok}`.
- `runs-cancel-convergence.test.ts`: pins that the response body carries the post-finalize state (`status: cancelled` + final `cost`).
- `runs-remote-registry.test.ts`: envelope `id` + `runId` removal pinned.
- CLI `run-remote-runner.test.ts`: trigger mocks `201` + `{id}`, shape-rejection tests renamed/retargeted; `run-report-source.test.ts` envelope fixture updated.

## Baseline

`apps/api/src/openapi/baseline.json` regenerated via `bun scripts/detect-breaking-changes.ts --update-baseline`. Against the committed baseline the script reports exactly the 5 intended BREAKING entries above (plus pre-existing non-breaking drift from already-merged PRs — the committed baseline was stale).

⚠️ **The regenerated baseline will conflict with the other open #657 batch PRs** — regenerate at merge time on whichever lands last.

## Validation

- `bun run check`: 29/29 tasks green (incl. OpenAPI verify, 203 endpoints).
- Touched API integration tests: 162 pass / 0 fail (runs, cancel-convergence, remote-registry, inline-run, inline-run-validate, version-selection, events-ingestion, run-wait).
- CLI tests: 42 pass / 0 fail.

### DIFF
--- a/apps/api/src/modules/mcp/router.ts
+++ b/apps/api/src/modules/mcp/router.ts
@@ -126,7 +126,7 @@ Organization → Applications (id \`app_…\`, one default) → Agents → Runs.
 This MCP server is scoped to ONE organization — the one this endpoint serves — and every operation runs against it plus its default application; you never send those ids per call. To act in another organization, connect that organization's own MCP server (its URL carries its id). Within the org, operations use the default application unless an operation takes an explicit application id.
 
 ## Beyond the per-operation schemas
-- Runs are asynchronous: triggering one returns a runId, then it moves pending→running→success|failed|timeout|cancelled. To wait for completion, call the run get operation with \`query: { wait: true }\` — the server long-polls up to ~55 s and returns the run when it goes terminal; a still-non-terminal response just means call it again (no sleep needed).
+- Runs are asynchronous: triggering one returns the created run resource (use its \`id\`), then it moves pending→running→success|failed|timeout|cancelled. To wait for completion, call the run get operation with \`query: { wait: true }\` — the server long-polls up to ~55 s and returns the run when it goes terminal; a still-non-terminal response just means call it again (no sleep needed).
 - Streaming/SSE operations (live logs, realtime) cannot be called through this server; fetch logs or poll instead.
 - Wire JSON is snake_case, except universal id/timestamp fields (id, createdAt…) which stay camelCase.`;
 
--- a/apps/api/src/openapi/baseline.json
+++ b/apps/api/src/openapi/baseline.json
@@ -796,7 +796,7 @@
         },
         "responses": {
           "200": {
-            "description": "Agent proxy updated",
+            "description": "Agent proxy updated — returns the affected proxy setting",
             "headers": {
               "Request-Id": {
                 "$ref": "#/components/headers/RequestId"
@@ -808,12 +808,28 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "type": "object",
-                  "properties": {
-                    "success": {
-                      "type": "boolean"
+                  "allOf": [
+                    {
+                      "type": "object",
+                      "required": ["proxyId", "resolved"],
+                      "properties": {
+                        "proxyId": {
+                          "type": ["string", "null"]
+                        },
+                        "resolved": {
+                          "type": "boolean"
+                        }
+                      }
+                    },
+                    {
+                      "type": "object",
+                      "properties": {
+                        "success": {
+                          "type": "boolean"
+                        }
+                      }
                     }
-                  }
+                  ]
                 }
               }
             }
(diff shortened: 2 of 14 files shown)


pierrecabriere:
Companion github-action PR: appstrate/github-action#10 (reads `data.id` instead of `data.runId`).

pierrecabriere:
Applied the two review fixes in cb606f3:

1. **e2e trigger status assertions (200 → 201)** — `e2e/tests/uploads/upload-flow.api.spec.ts:82,145` asserted `200` on `POST /agents/:scope/:name/run`; the trigger now returns `201` with the bare run resource. Grepped all of `e2e/` for other run-trigger status assertions or `runId` reads from trigger responses — these were the only two (the `400` magic-byte rejection and the `404/409/410` reuse assertions are unaffected). E2e tests can't run locally without the full stack, so the assertions were aligned with the documented behavior rather than verified live.

2. **Race-fallback vs strict `$ref Run`** — `apps/api/src/routes/runs.ts:174,427` can return `{id}` only when a concurrent teardown races the post-create read, which contradicted the strict `Run` schema. The 201 (`/agents/.../run`) and 202 (`/runs/inline`) response descriptions in `apps/api/src/openapi/paths/runs.ts` now note the body may degrade to `{id}` only under that rare race. The schema's required list is unchanged.

Gate: `bun run check` — 29/29 tasks passing (tsc, eslint, prettier, OpenAPI validation).

pierrecabriere:
**Real-conditions validation (local Docker, tier3) + stale e2e spec fix** (69256ae37)

Ran the full validation matrix on a local merge of all six #657 branches (merge-train):

- `bun run check` — 29/29 green (baseline regenerated cleanly after the 6-way merge; only `baseline.json` conflicts between branches, zero code conflicts)
- Full API integration suite in **tier3** (real PostgreSQL/Redis/MinIO/DinD via test compose): **1482/1482 pass**
- Full Playwright e2e suite against real PostgreSQL + `RUN_ADAPTER=docker` (real containers): **96/97 → 97/97 after the fix below**

The one failure was a **stale spec on main**, not a regression from this PR: `upload-flow.api.spec.ts` still asserted single-use uploads (`reusing an upload:// URI twice fails with 409`), but since #642 consumed uploads are retained for `UPLOAD_RETENTION_HOURS` and the URI is re-consumable within that window — the second trigger correctly returns 201. Nobody noticed because the e2e CI job is label-gated. Spec updated here (this PR already owns the e2e trigger-status updates): reuse now asserts a second 201 with a distinct run id. Verified green locally under the docker adapter.

PR #676, opened 2026-06-12
feat(web): typed OpenAPI client — end-to-end type unification, legacy fetch layer removed

pierrecabriere:
## Summary

Unifies backend↔frontend typing on the OpenAPI spec (chosen over tRPC: the REST API is the product — GitHub Action, MCP, API keys, external consumers — so the spec stays the single source of truth and the SPA now consumes it through a generated, compile-time-checked client).

### Generated types + drift gate
- `scripts/generate-api-types.ts`: assembles the core spec + built-in module contributions (same discovery as `verify-openapi`) → `apps/web/src/api/schema.d.ts` via openapi-typescript (committed, 170 paths). Deterministic.
- `verify:api-types` (`--check`) wired into `bun run check`: spec change without regeneration fails CI.

### Typed client
- `apps/web/src/api/client.ts`: openapi-fetch `client` + openapi-react-query `$api`. Middleware injects `X-Org-Id`/`X-Application-Id` and throws `ApiError` (RFC 9457) on non-2xx.
- Custom `pathSerializer`: keeps `@` literal and the `/` separator inside scoped package ids (`@scope/name`) — Hono regex routes match the raw path, so the default `encodeURIComponent` 404s every `@scope/{name}` route (same root cause as the known MCP `%40` bug). Unit-tested.

### Full migration — legacy layer deleted
- All hooks, pages, components and module UIs (oidc, webhooks) migrated; **`apps/web/src/api.ts` is deleted**. Permanent ESLint guard bans the old import specifiers (relative and `@/api`).
- SSE-patched caches (`use-global-run-sync`) and app/org-switch resets keep their pinned legacy query keys — typed fetch, unchanged keys, documented in code.
- `uploadClient` moved to `api/uploads.ts` (typed descriptor step); multipart via `bodySerializer`, blobs via `parseAs: "blob"`.
- Only sanctioned untyped call: `cloudApi` in `use-billing.ts` (cloud-module routes are absent from the OSS spec by design).
- Latent wire bugs fixed by the typing: webhooks create missing the spec-required `level`/`applicationId`; oidc sending `fromName: null` rejected by server Zod.

### Spec hardening (compensates the generated-types trade-off)
- `verify-openapi` step 6: every 2xx JSON response must declare a schema (13 justified allowlist entries).
- Response-validation tests now fail (not warn) on undocumented fields, core + webhooks module; PATCH/DELETE end-users coverage added. Caught and fixed: `GET /api/applications` returning undocumented `hasMore`.
- Spec accuracy fixes: `required` on list envelopes / `EndUserObject` / upload descriptor, `additionalProperties` on metadata.

### Known debt
None outstanding — the spec gaps were closed and **0 `as never` casts remain** in `apps/web/src` (the earlier "8 casts" debt was resolved in follow-up commits).

### Post-merge audit pass (commit `d1bf2267a`)
A multi-agent audit of the migration confirmed the gate's guarantees and closed what it doesn't catch:
- Spec gap: `GET /api/applications/{applicationId}/packages` omitted `hasMore` (the same drift fixed for `GET /api/applications`) — added to the schema + a response-validation test (the endpoint had no coverage, hence the miss).
- `run-detail` SSE log patch unified on the wire `RunLog` type (dropped an `as unknown as RunLog` Drizzle-`Date` cast); `useRunLogs` and the patcher now agree on the cache element type.
- OSS bug: the billing page fired the cloud-only `/billing` fetch (404) in OSS mode — now gated on `features.billing`.
- Dead-code cleanup: removed the `getCurrentOrgId` re-export (its comment cited the deleted `api.ts`), the dual-stack org-list invalidation, and a redundant uploads cast.
- Added `api/schema.assert.ts` — compile-time pins on load-bearing generated shapes, since `skipLibCheck` excludes `schema.d.ts` from tsc.

Deferred (documented, safe): over-permissive `null` on `runOrigin` / `AgentVersion.createdAt` response schemas (tightening trips `detect:breaking`); workspace npm modules absent from the spec-route scan (latent, no HTTP surface today).

## Test plan
- `bun run check`: 31/31 (typecheck, lint, format, verify-openapi ×7, verify:api-types, detect:breaking, conformance, module gates)
- Web: tsc clean, 103/103 unit tests, `vite build` OK
- API integration: response-validation 25/25, webhooks module 4/4, end-users routes 7/7
- Not run: Playwright e2e. No rendered-markup changes, but data-fetching / mutation / error-mapping / cache-invalidation call sites changed across 61 `.tsx` files — an e2e pass before merge is recommended.


pierrecabriere:
## Post-merge conflict resolution + merge-readiness re-analysis

**Conflicts with `main` (#674) resolved** (merge commit `e7ec5ac70`), typed-client style kept throughout:
- `use-schedules`: both list hooks unwrap the new `{ object, data, hasMore }` envelope
- `use-member-integration-pins`: DELETE query params renamed to `agent_package_id` / `integration_package_id`
- response-validation test: strict `extraFields` + envelope assertions combined
- `schema.d.ts` regenerated from the merged spec

**Re-analysis** (3 parallel review agents: completeness sweep, query-key coupling audit, behavioral diff vs main) found and fixed in `f6fc867f1`:
1. **Dead `["integrations"]` invalidations (5 sites)** — use-integrations was re-keyed to typed keys in this PR, so invalidations in use-editor-state / use-mutations / use-me-connections matched nothing; integration verdicts & lists went stale after writes. All five now call `invalidateIntegrationQueries()`.
2. **Bundle-import warnings restored** — the server has always returned required `warnings` (AFPS §7.7) but the spec omitted it; the typed migration silently dropped the toasts. Spec fixed, schema regenerated, hook surfaces them again.
3. **Module query-key contributions emptied & documented** — `["oauth-clients"]`/`["webhooks"]` app-switch reset prefixes matched nothing post-migration (oauth-clients data is org-scoped; webhook keys self-scope via app id in the key).
4. **Last 4 raw `fetch()` pages migrated** (welcome, claim, invite-accept ×2) + the oidc token exchange documented as deliberately untyped (end-user realm must not receive platform org headers). `cloudApi` + the SSE hooks are now the complete untyped surface, as the PR description claims.

**Audited clean**: SSE patcher ↔ pinned-key coupling (every run/schedule key matches exactly), all typed-mutation invalidation path strings, all 8 `as never` casts wire-identical to legacy, multipart/blob paths, enabled guards, pathSerializer edge cases.

**Pre-existing on main, deliberately untouched** (follow-up material): the `["packages","agent"]` dead-key family (SSE patcher + 4 mutation sites invalidate a key whose owners use `"agents"`/different shape) — identical dead code before this PR.

**Validation**: `bun run check` 30/30 · `detect:breaking` 0 breaking · web tests 100/100 · web tsc clean · response-validation 23/23 · schedules + pins + import-bundle tests pass.

PR #691, opened 2026-06-16
fix(web): unify missing-connections modal with the connection picker

pierrecabriere:
## Problem

The run-kickoff **412 recovery modal** ("Connexions d'intégration requises") and the **Connexions-tab dropdown** showed *different* connection lists for the same integration. Reported case: the modal offered 2 Gmail connections, the tab dropdown showed 3 (one needing renewal).

## Root cause

Both surfaces call the same endpoint (`GET /api/integrations/{pkg}/agent-resolution/{agent}`), but the modal applied an extra filter:

- Modal (`missing-connections-modal.tsx`) filtered must_choose candidates down to the 412 payload's `candidate_connection_ids`.
- The resolver (`integration-connection-resolver.ts`) populates that field with **healthy connections only** (`!needsReconnection`).
- The dropdown (`integration-connection-picker.tsx`) renders the **full** resolution candidate list.

So a connection needing reconnection appeared in the tab dropdown (with a renew button) but was silently dropped from the modal.

## Fix

Delete the bespoke must_choose list + `InlineConnectButton` + the candidate-id filter, and embed the shared **`IntegrationConnectionPicker`** in `override` mode in each actionable row — the same component the Connexions tab (pin mode) and schedule editor (override mode) already use.

The picker:
- lists every accessible connection (incl. needs-reconnection, with inline renew),
- exposes connect / renew / upgrade / add flows,
- writes the pick into the per-run `connection_overrides` map (override mode = transient, never persists a member pin).

Structural failures (`integration_not_active`, `package_not_found`, `not_installed_or_invalid_manifest`) keep a plain message — no connection pick can fix them.

Single source of truth for the connection-pick UX; the modal can no longer disagree with the dropdown.

## Notes

- Re-run gating simplified: only `must_choose` rows require a pick before re-run; other actionable rows re-run freely (a fresh 412 reopens the modal). Wiring to `run-agent-button.tsx` (`onRetryWithOverrides`) unchanged.
- Net **-42 lines**, single file.
- `agentPackageId` is always passed by the only caller; the picker is held behind a spinner until manifest + resolution land.

## Test plan

- [x] `apps/web` typecheck clean
- [x] eslint + prettier clean
- [x] `bun test src/components/integration-connect` (7 pass)
- [ ] Manual: trigger a run 412 on an agent with a Gmail connection needing renewal → confirm the modal now lists it with a renew button, and renewing in-place clears the row

### DIFF
--- a/apps/web/src/components/integration-connect/connection-status-badge.tsx
+++ b/apps/web/src/components/integration-connect/connection-status-badge.tsx
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: Apache-2.0
+
+import type { ReactNode } from "react";
+
+/**
+ * Health states a connection pill can express. `needsReconnection` is a
+ * RECOVERABLE warning (the owner can renew) — it must read as amber, never
+ * red/destructive, on every surface.
+ */
+export type ConnectionStatusTone = "connected" | "needsReconnection" | "missingScopes";
+
+const TONE_CLASSES: Record<ConnectionStatusTone, string> = {
+  connected: "border-emerald-500/40 bg-emerald-500/10 text-emerald-700 dark:text-emerald-400",
+  needsReconnection: "border-amber-500/40 bg-amber-500/10 text-amber-600 dark:text-amber-400",
+  missingScopes: "border-destructive/40 bg-destructive/10 text-destructive",
+};
+
+/**
+ * Connection-state pill shared across every surface that flags a connection's
+ * health — the member connections page and the integration detail page (the
+ * Connexions picker keeps its own dropdown-row affordances). Owns the
+ * tone→color mapping so the same state can never render red on one surface and
+ * amber on another: `needs_reconnection` was previously a destructive (red)
+ * badge on the integration detail page but an amber pill in preferences.
+ *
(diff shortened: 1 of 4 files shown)


pierrecabriere:
## Follow-up commit: shared `ConnectionStatusBadge`

Broader pass for other unification opportunities in this feature. Findings:

- **No second large component-reuse win exists** — the 412-modal↔dropdown divergence (this PR's first commit) was the big one.
- Genuine leftover: `needs_reconnection` rendered **red** (destructive) on the integration detail page vs **amber** on the member connections page — same recoverable state, two colors.

Fix: extracted `ConnectionStatusBadge` (tone→color SSoT: connected=emerald, needsReconnection=amber, missingScopes=destructive), routed both surfaces through it. amber everywhere now.

Dropped two over-claimed "opportunities" after verifying:
- `connectionDisplayLabel` in preferences — different type (`MeConnectionEntry.identity`, not `account_id`) + one site edits the label string itself. Not applicable.
- `canRenewConnection` predicate — integration detail gates renew on "OAuth client registered" (admin), the picker on ownership (`is_own`). Different conditions, not a duplication.

Picker untouched (its dropdown-row renew/marker affordances would mix primitives for no gain).

PR #705, opened 2026-06-17
feat(notifications): per-recipient notifications table (#667)

### LINKED ISSUE #667: Notifications: shared read-state on runs rows — refactor to dedicated per-recipient notifications table
## Problem

Notifications are not a real entity — they piggyback on `runs` rows via two columns (`notifiedAt` / `readAt`, `packages/db/src/schema/runs.ts:103-104`). This produces an incoherent scoping model:

1. **Read-state is global, not per-user.** `readAt` is a single timestamp on the run row. When one user marks a notification read, it becomes read for everyone.
2. **Mark-as-read fails silently for non-owners.** `markNotificationRead` (`apps/api/src/services/state/notifications.ts:23-40`) scopes the UPDATE with `actorOrOrgFilter`: `(userId = actor OR endUserId = actor) OR userId IS NULL`. When user B tries to mark user A's run notification as read, the UPDATE matches 0 rows and the route still returns `204` — no `403`/`404`, no mutation. Net effect: **org members see all notifications (runs list is org-visible) but can never mark another member's notifications as read.**
3. **Schedule/end-user runs (`userId IS NULL`)**: any org member can mark them read — for everyone at once.
4. **No dedicated list endpoint.** The bell filters the runs list client-side (`apps/web/src/components/notification-bell.tsx:119`: `notifiedAt != null && readAt == null`).
5. **No test** covers "user B cannot mark user A's notification".

Root cause: three models mixed — org-wide visibility, globally-stored read-state, per-user-scoped writes.

## Proposed design

Standard in-app notification model (fan-out on write / eager insertion — one row per recipient, per-user read-state by construction). Validated against current best practice: the existing design is effectively a degenerate fan-out-on-read with a shared read flag, which is the documented anti-pattern precisely because it can't support per-user mark-as-read.

### New table `notifications` (packages/db)

```typescript
export const notifications = pgTable("notifications", {
  id: uuid("id").primaryKey().defaultRandom(),
  orgId: uuid("org_id").notNull().references(() => organizations.id, { onDelete: "cascade" }),
  applicationId: text("application_id"),          // same scoping as runs
  userId: text("user_id").notNull(),              // recipient
  type: text("type").notNull(),                   // "run_completed" — extensible (invitations, billing…)
  runId: uuid("run_id").references(() => runs.id, { onDelete: "cascade" }),
  payload: jsonb("payload"),                      // agent_id, status… render without join
  readAt: timestamp("read_at", { withTimezone: true }),
  createdAt: timestamp("created_at", { withTimezone: true }).notNull().defaultNow(),
}, (t) => [
  index("idx_notifications_recipient").on(t.userId, t.orgId, t.readAt),
  index("idx_notifications_run").on(t.runId),
]);
```

snake_case wire/SQL, camelCase Drizzle — per `docs/CASING_CONVENTIONS.md`. `type` + `runId` play the standard `entity_type`/`entity_id` roles. No separate content table: payload is tiny, real content lives in the referenced `runs` row, fan-out is bounded by org size (revisit only if rich org-wide broadcasts appear).

### Fan-out (creation, in `finalizeRun` / run-event-ingestion)

- Run with `userId` → 1 notification for that user.
- Schedule or end-user run (`userId IS NULL`) → 1 notification per org member.
- Replaces `notifiedAt` as the trigger; pg `NOTIFY` on insert for realtime (reuse `packages/db/src/notify.ts`).

### API (replaces the 4 current endpoints)

| Endpoint | Behavior |
|---|---|
| `GET /notifications?unread=true&cursor=` | Paginated list for current recipient |
| `GET /notifications/unread-count` | User-scoped count |
| `GET /notifications/unread-counts-by-agent` | Kept — via `payload.agent_id` |
| `PUT /notifications/{id}/read` | `204` if recipient, **`404` otherwise** (no more silent no-op) |
| `PUT /notifications/read-all` | Bulk, user-scoped |

### Frontend

`apps/web/src/hooks/use-notifications.ts` + `notification-bell.tsx`: consume `GET /notifications` instead of client-side filtering of the runs list. Mutation shapes unchanged.

### Migration + backfill

1. Create table.
2. Backfill: run with `notifiedAt` + `userId` → 1 row (copy `readAt`); run with `userId IS NULL` → 1 row per current org member (copy `readAt` — preserves existing read state).
3. Deprecate then drop `runs.notifiedAt`, `runs.readAt`, `idx_runs_notification` in a **follow-up release** (soft transition).
4. OpenAPI: new paths; remove `notified_at`/`read_at` from the run schema (breaking — keep read-only for one release).

⚠️ **Known prod blocker**: the `__drizzle_migrations` watermark in prod is future-dated (~Sep 2026), so new migrations silently skip. Must be fixed **before** any release containing this migration.

### Tests

- Per-user read: A marks read → B still unread.
- B marks A's notification → `404`.
- Fan-out on schedule run → N rows (org members).
- Backfill correctness, read-all, idempotency, counts.

### Scope

~10 files: db schema + 2 migrations, `services/state/notifications.ts` (rewrite), `run-event-ingestion.ts`, `routes/notifications.ts`, OpenAPI paths/schemas, 2 web files, integration tests.

## Open decisions

1. **Recipients for schedule/end-user runs**: all org members, or only members with access to the application? (Proposed: all org members — simple, equivalent to current behavior.)
2. **Drop of `runs.notifiedAt`/`readAt`**: same release or next? (Proposed: next release, after the prod migration watermark is fixed.)

pierrecabriere:
Closes #667.

## Problem

Notifications piggybacked on `runs` rows (`notifiedAt` / `readAt`), mixing three incompatible models — org-wide visibility, globally-stored read-state, per-user-scoped writes:

- **Read-state was global.** One user marking a notification read flipped it for everyone.
- **Mark-as-read failed silently for non-owners.** A non-recipient's `UPDATE` matched 0 rows but the route still returned `204` — no mutation, no error.
- No dedicated list endpoint (the bell filtered the runs list client-side).

## Solution

Dedicated `notifications` table, **fan-out on write** — one row per recipient, so read-state is per-user by construction.

> **No backfill / no backward-compat.** There is no production data yet (the DB is reset on deploy), so this PR drops `runs.notifiedAt` / `runs.readAt` outright in the same migration and ships the clean model with zero legacy/transition scaffolding.

### Schema (`packages/db`)
- Recipient modelled as a single **polymorphic `(recipientType, recipientId)` tuple** — the undistorted `Actor` shape (`@appstrate/connect`: `{ type: "user" | "end_user", id }`). One indexable tuple instead of the nullable `{userId, endUserId}` pair → no XOR CHECK, one feed seek instead of an OR-across-two-columns bitmap, one dedup unique index instead of two. Extends to future recipient kinds (team, org, system) by a new `recipientType` value, no schema change.
- `recipientType` follows the codebase's polymorphic-actor convention (`package_persistence.actorType`): `text` + a TS union (`"user" | "end_user"`) for compile-time safety **+ a CHECK** for DB integrity — **not** a `pgEnum` (reserved for closed-set semantics; would need `ALTER TYPE` to add a kind).
- `type` + `runId` are the standard entity_type/entity_id roles → extensible beyond runs (invitations, billing). `run_id` is **text** to match `runs.id` (`exec_`/`run_`-prefixed), `ON DELETE CASCADE`.
- **3 indexes**: a partial `idx_notifications_unread` `(org, app, recipient_type, recipient_id, created_at DESC, id DESC) WHERE read_at IS NULL` backs every hot read (unread count, unread feed, unread keyset); `idx_notifications_run` backs the run-id FK cascade + by-run lookups; `uq_notifications_run_recipient_type` is the exactly-once dedup guard.

### Recipient cleanup (replaces FK cascade)
`recipientId` carries no FK (it spans two tables), so deleting a user/end-user does not cascade their notifications. `orgId` / `applicationId` FKs still cascade; the narrow "delete one user/end-user but keep the app" case is handled explicitly, each in a **transaction**:
- `deleteEndUser` drops the end-user's notifications across the app.
- `removeMember` drops the departing member's notifications across the org.

### Fan-out
In the `finalizeRun` CAS winner (exactly-once), best-effort (never fails the terminal run):
| Run trigger | Recipients |
|---|---|
| Dashboard user | that user |
| End-user / impersonation | that end-user |
| Schedule / actor-less | org admins/owners only |

⚠️ **Behaviour change:** end-user runs no longer appear in every org member's bell — they notify the end-user only (see Open Decisions).

### API
| Endpoint | Behaviour |
|---|---|
| `GET /api/notifications?unread=&limit=&startingAfter=` | **keyset-paginated** (`(created_at, id)`), recipient-scoped; returns `{ data, has_more }` + `Link: rel="next"` cursor |
| `GET /api/notifications/unread-count` | recipient-scoped |
| `GET /api/notifications/unread-counts-by-agent` | via `payload.agent_id` |
| `PUT /api/notifications/{id}/read` | `204` if recipient, **`404` otherwise** (no silent no-op) |
| `PUT /api/notifications/read-all` | bulk, recipient-scoped |
| `PUT /api/notifications/read/{runId}` | first-class by-run convenience (run-detail holds the run id, not the notification id) |

**Pagination is keyset, not offset:** the cursor is a plain notification id whose `(created_at, id)` coordinate is resolved by a scalar subquery **scoped to the caller's tenant + recipient**, so the row-value comparison stays consistent with the `ORDER BY`. Deep pages are a tight `idx_notifications_unread` range scan (no OFFSET count, no exact `COUNT(*)`) and are skip/duplicate-safe under concurrent inserts; a stale/foreign cursor resolves to an empty page.

### Frontend
Bell + run-detail consume the new endpoints; near-live refresh via the existing terminal-`run_update` notification-query invalidation (no new SSE channel).

## Tests
`notifications.test.ts` rewritten to the new model (**53 tests**): per-user read isolation (the bug), `404` on cross-member mark, fan-out by trigger (user→1, end-user→1 to the end-user/0 to members, schedule→N admins), the **real `finalizeRun` convergence path** (synthesise → CAS → awaited fan-out) incl. idempotent re-finalize, exactly-once under a concurrent double fan-out (unique index), counts-by-agent (incl. read-exclusion), read-all scoping, by-run mark, org + cross-app isolation, polymorphic recipient NOT-NULL + CHECK integrity, recipient cleanup on user/end-user deletion (recipient-typed), and **keyset edges**: a walk asserting no skip/duplicate, pagination across rows sharing an identical `created_at` (the tuple tiebreak), a stale cursor → empty page, and `has_more` on an exactly-`limit` final page. Full API suite green (2501 pass, 0 fail).

## Adversarial review & hardening

Reviewed across 4 independent lenses (correctness/concurrency · security/tenant-isolation · test quality · industry-standard conformance / over-engineering). **No critical or high-severity bug found.** Findings acted on:

- **Cursor subquery scoped** to the caller's `(org, app, recipient)` — was a global PK lookup (a weak existence-oracle for notification UUIDs + a way for a stale/foreign cursor to borrow another tenant's timestamp). No data ever leaked (the outer query is always scoped), but the smell is gone and a stale cursor now consistently yields an empty page.
- **Cleanup hooks made atomic** (transactions).
- **Dropped a reader-less index** (`idx_notifications_feed`): the bell only queries unread, so the partial unread index serves every hot read; the full twin only added write-amp. Re-add (with a retention policy) when a paged full-history feed ships.
- **+7 tests** closing the gaps the review surfaced (real finalize path, keyset tiebreak/stale-cursor/boundary, counts read-exclusion).
- Documented the bounded actor-less-fan-out vs member-removal orphan race (harmless, invisible, org-cascade-reaped).

Deliberately **not** changed (conscious trade-offs):
- **Polymorphic recipient over the `{userId,endUserId}` FK pair** — forfeits a free `ON DELETE CASCADE` (hence the explicit cleanup), but matches how dedicated notification systems decouple the recipient; cleanup is now atomic + tested.
- **SSE refetch breadth** — `run_update` is org/app-scoped, so every connected client refetches the bell on every terminal run. The naive "gate on recipient" fix is incorrect (the event carries the run actor, not the notification recipients → would break admin live-refresh for actor-less runs). Proper fix = a dedicated per-recipient notification channel — deferred.

## Out of scope (roadmap, not regressions)
Batching/digest/throttle · user preferences/opt-out · multi-channel (email/push) · websocket **push** of the notification object (today = invalidate-and-refetch) · retention/pruning reaper · a paged full-history feed (would re-add the dropped feed index).

## Open decisions
1. End-user runs → end-user only (this PR) vs also org members (admin oversight)?


pierrecabriere:
## Review follow-up (second pass)

Pushed fixes for a critical self-review. Summary:

**Fixed**
- **Bounded actor-less fan-out.** Correction to the original analysis: the scheduler copies `schedule.userId`/`endUserId` onto the run, so *owned* schedules already produce a single notification to the owner — no spam. The all-members fan-out only ever triggered for **owner-less org/system schedules** (both null). That case now fans out to **org admins/owners only** (decision: bounded set, the people who manage org/system schedules; plain members still see the run in `/runs`). Backfill branch 3 matches.
- **Duplicate guard.** Added two partial unique indexes `(run_id, user_id, type)` / `(run_id, end_user_id, type)`; the fan-out insert uses `onConflictDoNothing` so a re-fire is a safe no-op. New test asserts a second fan-out creates no duplicate.
- **Un-deprecated `PUT /notifications/read/{runId}`.** It's a legitimate first-class "mark by run id" op (run-detail holds the run id, not the notification id), not a transitional alias. PR2 no longer needs to couple to it.
- Dropped a spurious `404` declared on `GET /notifications`.

**Accepted tradeoffs (not bugs, documented)**
- *Fan-out reliability:* the old `notifiedAt` flag was set atomically in the terminal CAS UPDATE; the new fan-out is a separate best-effort post-CAS INSERT (consistent with the sibling memory/log side-effects there). A transient DB fault means a missing notification while the run is still correctly terminal. Inherent to any separate-table design; no retry added.
- *Retention:* notifications inherit `runs`' lifetime via `ON DELETE CASCADE`. Runs have no TTL today, so neither do notifications. With the bounded fan-out, cardinality now tracks `runs` (≈1 row/run for the common case), so this is no worse than the table it replaces.

Verification: `bun run check` green; full API suite **2464 pass / 0 fail**; all migrations apply clean on a fresh DB.

Open decisions unchanged (still need a maintainer call): keep `runs.notifiedAt`/`readAt` for PR2 drop (recommended); prod `__drizzle_migrations` watermark must be fixed before this ships.

pierrecabriere:
## Second-pass review follow-up

Re-reviewed this branch adversarially (5 parallel reviewers across DB/migration, concurrency, security, API/frontend, tests) and fixed everything actionable. Pushed in `07a95f60e`.

### Confirmed CRITICAL — fixed
- **Run-list & schedule-card badges were stuck "unread" forever.** Those surfaces still derive `unread` from the legacy global `runs.readAt`, but the mark paths had stopped writing it (only `notifications.readAt` was updated). The "dual-write soft transition" was only half-done — the read-clearing side was dropped. Completed it: all three mark paths now mirror onto `runs.readAt` via a shared `legacyMarkRunsRead` helper (verbatim port of the pre-#667 queries). The two components move onto the notifications source in the column-drop follow-up.

### Medium — fixed
- **M2** `markNotificationRead` non-atomic UPDATE-then-SELECT → spurious 404 if a concurrent run-delete cascaded the row away between statements. Now a single atomic `UPDATE … SET read_at = COALESCE(read_at, now()) … RETURNING`.
- **M1** Fan-out moved off the finalize critical path (fire-and-forget `void … .catch`, like the sibling workspace-delete) so a slow INSERT no longer delays the terminal status broadcast / webhooks.
- **M3** operationId reshuffle that would silently retarget SDK consumers: by-run keeps its shipped `markNotificationRead` id; the new by-id endpoint took a fresh `markNotificationReadById`.
- **M4** `idx_notifications_unread` couldn't serve the OR-across-recipient query — split into two partial indexes `(org_id, application_id, user_id)` / `(org_id, application_id, end_user_id)`, each `WHERE read_at IS NULL`. Migration 0012 regenerated.
- **M5** run-detail auto-mark keyed on `notifiedAt` (which the SSE patch omits) → never fired for a run finalizing while the page was open. Re-keyed on terminal `status`.
- **H1** backfill INSERTs made idempotent (`ON CONFLICT DO NOTHING`) — safe to re-apply during the watermark realignment below.

### Design conflict resolved
H2 (make fan-out transactional) was **rejected**: putting the INSERT in the finalize CAS tx would let a notification failure roll back an already-terminal run, violating the "a notification write must never fail the run" invariant. Kept best-effort + applied M1 instead.

### Tests
+13: run-not-found→0, positive end-user fan-out, CHECK rejects both/neither recipient, null-`agent_id` grouping skip, list pagination/ordering/Link, cross-org mark 404, by-run non-recipient no-op + idempotency, read-all unread-subset, **cross-application isolation**, and the `runs.readAt` dual-write. Notifications suite 37/37; `bun run check` 31/31 green; full API suite passes (the only red is flaky Docker/DinD timeouts that pass in isolation).

### Not a code change — ⚠️ deploy blocker
Prod `__drizzle_migrations` watermark is future-dated (~Sep 2026), so 0012 silently skips on prod and the `notifications` table is never created → runtime errors on the live finalize + bell paths. **Realign the watermark before deploying.**

### Accepted tradeoffs (documented, not bugs)
- Fan-out durability is best-effort (consistent with memories/logs/workspace-cleanup side-effects); the atomic `runs.notifiedAt` flag still backs the run-list badge if a fan-out is dropped this release.
- Cross-recipient isolation rests on user-ids never carrying the `eu_` prefix (true under Better Auth's default id scheme).

pierrecabriere:
## Third-pass critical review + hardening (`cfa60e891`)

Re-ran an adversarial review with 6 independent reviewers (DB/migration, concurrency/finalize, API/OpenAPI, frontend, security/multi-tenancy, tests). **The prior second-pass fixes (C1/M1–M5/H1) all hold up** — finalize fan-out is genuinely post-CAS and non-blocking, exactly-once + ON CONFLICT covers all recipient shapes, mark atomicity has no TOCTOU, schema↔SQL↔snapshot has zero drift, backfill partitions correctly + is idempotent, every query is org+app scoped, cross-org/cross-app isolation is blocked & tested, the CHECK is unbypassable, the operationId strategy is non-breaking, and the SSE status-keyed auto-mark is correct.

New findings addressed in this commit:

| # | sev | fix |
|---|-----|-----|
| **N1** | MED | `legacyMarkRunsRead` is now genuinely best-effort — the authoritative per-recipient UPDATE commits first, then the legacy `runs.readAt` mirror swallows+logs failures instead of propagating a **500 on an already-committed mark-read**. All three mark paths. |
| **N2** | LOW | `recipientFilter` no longer rests on the unenforced "user ids never collide with `eu_` end-user ids" invariant. The full `Actor` is threaded through the service and the recipient column is selected with the shared **type-aware `actorFilter`** (same helper the rest of the codebase uses) — removes the only type-blind OR in the module. |
| **N4** | LOW | All `read_at` writes use the DB clock (`now()`) — single timestamp source, no app-side `new Date()` divergence. |
| **N5** | LOW | `notifications.application_id` is **NOT NULL** — every notification is app-scoped; a NULL would be unreachable by `scopedWhere`. 0012 SQL + snapshot regenerated; `db:generate` reports no drift. |
| **N3** | — | +5 tests: failed-status payload, **concurrent** exactly-once (exercises the partial unique indexes, not just app logic), dual-write cross-run isolation + pinned pre-NULL, end-user mark path (validates type-aware filter incl. wrong-actor-type no-op). CHECK assertions tightened to verify the rejection is specifically `notifications_one_recipient` (SQLSTATE 23514). |

**N6** (refresh `baseline.json`) intentionally **not** bundled — it pulls ~3.2k lines of pre-existing endpoint drift from other merged PRs into this diff. `detect:breaking` already passes (new endpoints classified additive); the refresh belongs in a standalone maintenance commit.

Not changed (transitional / self-resolving, documented in code): bell↔run-list badge divergence with ≥2 admins and the non-recipient `markRead` no-op — both dissolve when the legacy `runs.notifiedAt`/`readAt` columns are dropped in the follow-up.

**Gates:** notifications suite 42/42 · typecheck (api/web/db/shared-types) · lint · format · verify:openapi · verify:api-types · detect:breaking (0 breaking, 20 non-breaking) — all green. (DB tests run under the PGlite Tier-0 path; the local Docker daemon is down.)

⚠️ **Deploy blocker unchanged (ops, not code):** the prod `__drizzle_migrations` watermark is future-dated, so migration 0012 will be silently skipped and the `notifications` table never created. Realign the watermark before deploying this branch.

PR #707, opened 2026-06-17
fix(integrations): remote-MCP OAuth — auto-provisioning guard + MCP-spec refresh alignment

pierrecabriere:
## Problem

Connecting the **ClickUp MCP** integration with a `client_id`/`client_secret` created on the ClickUp **REST API** portal fails at the authorize redirect:

```
mcp.clickup.com/oauth/authorize?client_id=L8NTR0...&...
{"error":"invalid_client","error_description":"Invalid client_id"}
```

Root cause: a remote MCP integration whose auth declares `token_endpoint_auth_method: "none"` (`@appstrate/clickup-mcp`, `@appstrate/notion-mcp`) provisions its OAuth client **automatically** at connect time via DCR/CIMD. Its authorization server (`mcp.clickup.com`) is a *different* OAuth server from the REST API portal (`app.clickup.com`) and only accepts a public client it issued itself.

Both the UI and the API let the user store a manual client anyway. Once stored, `ensureIntegrationOAuthClient` returns the stale client and **skips DCR**:

```ts
// integration-connections.ts
if (existing) return resolved; // <- never runs DCR again
```

…so the wrong `client_id` is sent to the MCP authorization server.

Verified against the live `mcp.clickup.com/.well-known/oauth-authorization-server`:
- `registration_endpoint` present → DCR is the intended path
- `token_endpoint_auth_methods_supported: ["none"]` → manual confidential creds are **impossible** there

## Fix

- **Backend** — `PUT /api/integrations/{packageId}/oauth-clients/{authKey}` now rejects (`400`) a manual client when `usesAutoProvisionedClient(manifest, auth)` is true. Guards the API/curl path, not just the UI.
- **Frontend** — the manual client form is hidden by default for auto-provisioned auths, behind an opt-in *"register manually"* escape hatch (for the rare public-client server without DCR). When a manual client is already stored, a warning + delete button lets an admin clear it and restore auto-registration.

## Why no schema refactor

Inventory: of 47 oauth2 integrations, **2** are auto-DCR (public) and **45** are manual confidential. The two combinations the binary `token_endpoint_auth_method` derivation can't express (manual-public, auto-confidential) have **zero** real integrations. Decoupling acquisition-mode from client-type stays YAGNI; the SOTA direction (MCP CIMD, Nov 2025) moves *away* from manual creds, not toward them.

## Recovery for existing victims

Delete the stored client (now exposed via the warning's delete button, or `DELETE .../oauth-clients/{authKey}`), then reconnect with no credentials → DCR self-registers a valid client.

## Tests

- New: `refuses a manual client on an auto-provisioned (remote MCP) auth (400)` + asserts nothing persisted.
- Full `integrations.test.ts` suite: 38 pass.
- `bun run check` green (typecheck, eslint, prettier, detect:breaking — 0 breaking).


pierrecabriere:
## Update — second fix added: MCP-spec refresh alignment

This PR now carries two related remote-MCP OAuth fixes.

### 1. Auto-provisioning client guard (original)
A remote MCP auth with `token_endpoint_auth_method: "none"` self-registers via DCR/CIMD. Entering a manual `client_id`/`secret` for it (UI form was always shown) stored a client that made `ensureIntegrationOAuthClient` skip DCR → wrong client_id sent → opaque `invalid_client`. Now the API rejects a manual client on auto-provisioned auths (400) and the UI hides the form behind an opt-in escape hatch + offers delete-to-restore.

### 2. MCP-spec refresh alignment (new)
Two connect-time refresh gaps, both capability-driven via RFC 8414 `grant_types_supported` (now projected by `resolveOAuthEndpoints`):

- **DCR never requested the refresh_token grant** (`grant_types: ["authorization_code"]` hardcoded) → the AS registered a client not authorised for refresh → never issued a refresh token (Claude Code #7744). Now registers `["authorization_code","refresh_token"]` **when the AS advertises it**.
- **The guard refused every short-lived token without a refresh token** — even when the AS can't issue one. ClickUp MCP advertises no refresh grant; the hard refusal made it unconnectable. Now refusal applies only when the AS DOES support refresh (real offline-access misconfig, e.g. Google without `prompt=consent`); otherwise the connection persists and re-authorises at expiry.

### End-to-end validation (real servers, real consent)
- **Notion MCP** (advertises refresh grant): connection persisted **with** refresh_token; back-dated the access token, ran an agent → auto-refresh fired (expiresAt advanced), refreshed token authenticated against Notion (all `tools/list` + tool calls succeeded, zero 401).
- **ClickUp MCP** (no refresh grant): connection now persists (access-only, ~10y token) instead of hard-failing.

### Conformance
No impact on the system-package tool-declaration check. The merge `gate` tier tests only local MCP servers (no OAuth). The nightly monitor's remote tool-parity calls `resolveOAuthEndpoints` but does not read the new `grantTypesSupported` field — additive, behaviour unchanged.

Added the `integration` label so the new `integration-oauth-refresh-token-guard.test.ts` cases run in CI.

Issue #723, opened 2026-06-18
Multiple OAuth clients per integration (system default + per-org BYO-app)


pierrecabriere:
## Summary

Allow an integration to have **multiple OAuth clients** simultaneously: a **system client** provided by env (enabled by default, shared across all orgs) **plus** per-org/per-app **custom clients** (BYO-app), without removing the system one. A connection records *which* client minted it, so token refresh/revoke resolve the right credentials.

This unblocks the SaaS/all-users story: today the platform's Google OAuth app credentials only work for the org that configured them. Every other org hits a dead connect flow because no `integration_oauth_clients` row exists for it.

## Background & decision

- Today integrations resolve OAuth client credentials by `(applicationId, integrationId, authKey)` with a **unique constraint** → exactly one client per auth key.
- Connections (`integration_connections`) **do not reference a client**; refresh re-resolves by `(integrationId, authKey)`. So >1 client is impossible without breaking refresh.
- We evaluated three options:
  - **A. Env-only system default** — simplest, but no per-org BYO-app without losing the system one.
  - **B. Fork the integration** to a new id and attach own creds — generic (fork already exists) but causes **manifest drift** (fork is a full snapshot, no upstream sync) and forces repointing agents. Fork stays the right tool for editing scopes/tools, **not** for swapping a credential.
  - **C. Multiple clients per integration (this issue)** — system + custom coexist, per-connection selection. Mirrors the existing **model-provider** pattern (`SYSTEM_PROVIDER_KEYS` + `mergeSystemAndDb` + `source: built-in|custom` + system-first credential resolution). Most aligned with the industry connector pattern (one connector, N credentials, per-connection override) and **zero drift**.

**Chosen: Option C.**

### Industry / security note

Sharing one OAuth **client** across orgs is the **standard** for third-party API connectors (Zapier, Notion, Slack, Make all use a single vendor-registered app for all customers). Tenant isolation lives at the **token** layer (per-user encrypted tokens — already implemented), not at the client_id layer. The client_secret stays server-side and is never exposed to orgs, so there is **no cross-tenant data risk**. The residual risks are **shared-fate** (shared Google quota; one org's abuse can get the whole app suspended) — mitigated by the custom-client escape hatch + per-org rate limiting + abuse monitoring. (Do not conflate with SSO/IdP federation, where per-tenant client isolation *is* required — different problem.)

## Design

Unify system + custom clients as a single list, exactly like model providers:

[...]

PR #724, opened 2026-06-18
feat(integrations): multiple OAuth clients per integration (system + BYO-app)

pierrecabriere:
Implements #723.

## What & why

Lets an integration auth be served by a **shared, env-provided system OAuth client** **and/or** the org's **own per-application client**, coexisting.

This unblocks the SaaS story: today the platform's Google OAuth app credentials only work for the org that configured them, so every other org hits a dead connect flow. With a system client, every org can connect Gmail/Drive/… **out of the box** using the platform's verified app — the standard connector pattern (Zapier/Notion/Slack). Tenant isolation stays at the **token layer** (per-connection encrypted tokens, already implemented); the shared `client_id` only identifies the app to the IdP, so there is no cross-tenant data risk.

## Design

- **`SYSTEM_INTEGRATIONS` env registry.** A single env var declares which integrations the deployment **offers out of the box**. Membership = the **auto-active** policy: a listed integration is on by default (usable without any prior user action) until an org explicitly opts out (a sticky `application_packages.enabled = false`). Each entry **MAY** ship one or more shared OAuth clients (`clients: [{ id, auth_key, client_id, client_secret? }]`) — the SaaS connector pattern. An entry **MAY** ship **no** clients: a remote-MCP integration relying on Dynamic Client Registration (RFC 7591) is still offered/auto-active and provisions its client lazily on first connect.
- **`isSystemIntegration(id)`** is the membership predicate driving auto-active. Activation rule: an `application_packages` row EXISTS → its `enabled` flag wins (sticky); no row → auto-active iff `isSystemIntegration`.
- **Custom client** = the existing org per-application `integration_oauth_clients` row (BYO-app), `source: custom`. Exactly one default per `(app, integration, auth_key)` — DB-enforced (`idx_ioc_one_default`). Matches the industry norm (Nango/Paragon/Merge: many connections, one app).
- **`integration_connections.client_ref`** (new, nullable) pins **which client minted** a connection as a **flat client id** — the env id of a system client or the `integration_oauth_clients.id` (UUID) of a custom client. **No prefix/sentinel scheme.** Resolution is **system-first then DB-by-id**, the direct analogue of the model-provider `loadInferenceCredentials`. Token **refresh + revoke** resolve the same client by this id.
  - **Invariant:** `NULL ⟺ a non-oauth2 auth` (api_key / login_secret — no OAuth client). Every oauth2 connection carries a non-null id. Big-bang reset → no legacy rows, no data migration.
- **Connect precedence** (no per-connect picker — the default is an admin choice via `setDefaultIntegrationClient`, the `setDefaultModel` analogue): the org's flagged-default custom client → the default system client → the first custom client. Auto-provisioned remote-MCP auths (DCR/CIMD) keep their own client.

## Security

- **System client_id is never returned to the front.** It is a deployment secret; the clients-list endpoint returns an opaque, stable **fingerprint** (`sys_<sha256[:16]>`, display-only) in place of the real id, and `client_ref` (not the fingerprint) is the connect/refresh handle. The real id is also redacted from invalid-entry logs.
- System secret stays server-side (env / secret-manager), never returned by any GET, never logged. Same posture as `SYSTEM_PROVIDER_KEYS.apiKey`.
- Cross-integration / cross-app escalation blocked on **both** connect (`resolveConnectClient`) and refresh (`resolveIntegrationClientById`): a system id is re-validated to serve the requested `(integration, auth_key)` (`resolveSystemClientForAuth`), and a custom id must belong to `(applicationId, integration, auth_key)`. Custom-by-id lookup is UUID-shape-guarded so a non-UUID `client_ref` returns null instead of a Postgres uuid-cast error.
- Residual risk of a shared app is **shared-fate** (shared Google quota / suspension), not cross-tenant data. BYO-app is the escape hatch.

## Unification (DRY)

- **`resolveIntegrationClientById`** — single by-id resolver, shared by connect + refresh.
- **`listIntegrationClients`** uses the generic `mergeSystemAndDb` helper (same as org-models / org-proxies / model-credentials).
- The two flat system registries (`SYSTEM_PROVIDER_KEYS`, `SYSTEM_PROXIES`) share the `loadSystemRegistry` skeleton; org-models/org-proxies share `createDefaultPointer` (org default as a pointer). Deleted the bespoke DSL (`SYSTEM_CLIENT_REF_PREFIX`, `parseClientRef`, …).

## Changes

| Area | Change |
|---|---|
| env | `SYSTEM_INTEGRATIONS` (`packages/env`), docs (`docs/ENV.md`, `.env.example`) incl. shared-quota caveat + Notion MCP |
| registry | `integration-client-registry.ts` + boot init; fail-fast throw-on-uninit |
| schema | nullable `client_ref` on `integration_connections` + migration `0014` |
| connect | `OAuth2Strategy` → `resolveConnectClient` (default-only); threads the chosen client id through OAuth state → callback → connection row |
| refresh | `resolveIntegrationClientById` (system env vs custom DB row by id); public clients carry no secret |
| API | `GET …/auths/{authKey}/clients` lists available clients (built-in client_id is a `sys_` fingerprint; secrets never returned); `PUT …/default-client` sets the default |

## Tests

- Registry **unit**: env parsing, skip-invalid/duplicate, public/clientless (DCR) entries, `resolveSystemClientForAuth`, reset semantics.
- Service **integration**: `client_ref` round-trip, `resolveIntegrationClientById` (system / custom-by-id / cross-app + cross-integration escalation guards / unknown / public no-decrypt), `listIntegrationClients` merge + default + **no real client_id leak** (asserts the `sys_<hash>` fingerprint, not the real id).
- Route **HTTP**: list shapes, connect out-of-the-box with the system client (200), default-client happy path + **unknown ref → 400** + **cross-app ref → 400**, cross-integration system id → 400.
- Activation precedence (auto-active, sticky opt-out, explicit enabled row).
- End-to-end proxy refresh with a system-pinned connection.

`bun run check` green (31/31).

## Out of scope (follow-up)

- **Per-user quota partitioning (`quotaUser`) for shared system clients → #725.** Must land before enabling a system client on a quota-strict provider in prod.
- White-label / anonymous models (the same masking idea applied to model ids) → #727.

> **Deploy note:** prod must rename + reshape `SYSTEM_INTEGRATION_CLIENTS` → `SYSTEM_INTEGRATIONS` (now `[{ id, clients?: [...] }]`) at the next deploy, or the Gmail system client is lost.


pierrecabriere:
## Critical review + big-bang cleanup (no production data → reset)

Ran a 4-lens parallel review (correctness, legacy/back-compat, security, standards/tests). **No blockers or vulnerabilities.** Applied the findings for a 100%-clean, no-legacy result.

**Security** — clean: system secret never returned/logged/serialized; system-client pinning enforces integration+auth match; refresh + connection lookups both `applicationId`-scoped (no tenant escape); connect `client_ref` regex anchored/ReDoS-free; system redirect URI is server-controlled.

**Legacy removed (big-bang):**
- `parseClientRef` → closed set `{"system:<id>","custom"}`, throws on malformed (no null/garbage→custom shim).
- `client_ref` invariant: **NULL ⟺ non-oauth2 auth**; every oauth2 connection pins a non-null ref. Refresh guards null (skip) instead of guessing. All "pre-multi-client / legacy row" framing deleted.

**Robustness hardening:**
- Refresh re-validates the pinned system client still serves this `(integration, authKey)` (mirrors connect guard) — env reshuffles can't cross creds.
- Explicit `system:` ref on an auto-provisioned (DCR/CIMD) auth → rejected, not silently coerced.
- Registry `id` constrained to `^[\w.-]+$` (matches the wire-addressable set).
- Lazy-init kept (over throw) as deliberate idempotent env-memoization — robust under the shared-process test runner, identical to boot in prod.

**Tests added:** strict `parseClientRef`; null-ref → skip; system-client remap → skip; route-level custom-client connect + no-silent-downgrade (403) + cross-integration reject (400); `packages/connect` begin→state→callback round-trip; **credential-proxy end-to-end refresh of a system-pinned connection** (proves the `client_ref` pin threads from the connection SELECT through to the refresh context). Existing oauth2 fixtures updated to pin `client_ref="custom"` per the invariant.

Full API suite **2540 pass / 0 fail**, connect **256 pass**, `bun run check` green.

pierrecabriere:
Follow-up tracked in #725 — per-user quota partitioning (`quotaUser`) for shared system integration clients.

Not a blocker for this PR (`SYSTEM_INTEGRATION_CLIENTS` is opt-in, defaults to `[]`). A shared system client funnels every org's API calls through one upstream project quota (e.g. Google's per-project Gmail quota), so a noisy org can degrade others. The BYO-app override shipped here is the SOTA mitigation for heavy tenants; #725 adds the fair-sharing primitive for orgs that stay on the shared client. Must land before enabling a system client on a quota-strict provider in prod.

PR #729, opened 2026-06-19
feat: white-label model aliases (LLM-gateway alias pattern)

### LINKED ISSUE #727: White-label / anonymous models (Appstrate small·medium·large) — LLM-gateway alias pattern
## Goal

Add **white-label ("anonymous") models**: let an operator expose a model under an Appstrate-branded vanity name (e.g. `Appstrate small` / `medium` / `large`) while the real backing model — a fine-tune on DeepSeek, a GPT variant, whatever — stays hidden server-side. On the Cloud plan a user picks `Appstrate medium` and never learns what is behind it.

This is the model analogue of the **system-integration client masking** that just shipped on the multi-client-integrations PR (see *Precedent* below). The two share a *principle* but **must NOT be force-merged into one abstraction** — see *Why not unify* below. This issue is to be implemented from scratch in a later, separate effort.

---

## Motivation / use case

Cloud wants to sell capability tiers, not model names:

- Decouple the commercial offer ("small/medium/large") from the underlying vendor, so the backing model can be swapped (cost, availability, a better fine-tune) without changing what the user selected.
- Protect a proprietary fine-tune: don't reveal it's built on DeepSeek/GPT/etc.
- Same precedent already exists in the cloud world: **Azure OpenAI "deployment name"** decouples the name you call from the real model/version; **LiteLLM model aliases** map a public `model_name` to a hidden real model.

---

## Precedent shipped in this PR (read first)

The multi-client-integrations PR already does the **integration** half of the "hidden real identity, public-safe projection" principle:

- `SYSTEM_INTEGRATIONS` env registry: integrations the deployment offers, optionally shipping shared OAuth clients whose `client_id`/`client_secret` stay server-side. See `apps/api/src/services/integration-client-registry.ts`.
- `GET /api/integrations/{id}/auths/{authKey}/clients` no longer leaks a system client's real `client_id`. For `built-in` clients it returns an opaque `sys_<sha256[:16]>` fingerprint instead — `fingerprintSystemClientId()` at `apps/api/src/services/integration-connections.ts:870` (used at `:908`).

That masking is **one-way, display-only**: the fingerprint is never a key (the real handle is `client_ref`, the env id), and the real `client_id` legitimately reaches the IdP in the OAuth `authorize` URL (inherent to OAuth, not a UI leak). No request/response rewrite is involved.

---

## Why NOT unify integration-masking and model-anonymisation (design decision — do not revisit lightly)

They look alike at 30,000 ft ("hide a real identity, show an opaque token") but have **opposite change axes**, so a single shared "anonymizer" abstraction would be the wrong cut (anti-SRP, anti-KISS, and premature → YAGNI):

| | System integration client mask | White-label model |
|---|---|---|
| Direction | **one-way**, cosmetic | **bidirectional**, in the inference data-path |
| Surface | one list endpoint | inference req + resp + streaming, model list, run env, base URL |
| Public token | **throwaway**, display-only (`sys_<hash>`) | **load-bearing**: user-selected, stored in agent config, routed, rewritten |
| Real handle | `client_ref` (env id) — fingerprint keys nothing | the alias **is** the key, resolved on every call |
| Token origin | deterministic hash of the secret | operator-chosen vanity name |
| Reversibility | never (hash→real never read back) | constant live lookup (every inference) |

**What is already DRY and SHOULD be reused** (this is the correct shared layer, not a new super-abstraction):

- `apps/api/src/lib/system-registry.ts` → `loadSystemRegistry()` — the parse→validate→dedupe→log skeleton shared by `SYSTEM_INTEGRATIONS`, `SYSTEM_PROVIDER_KEYS`, `SYSTEM_PROXIES`. The model registry already uses the `SYSTEM_PROVIDER_KEYS` variant (`apps/api/src/services/model-registry.ts`).
- The "never return the secret field in accessors" discipline (integration `clientSecret`, model `apiKey`) — same principle, separate code per shape, which is correct.

Do **not** introduce an "anonymous integration ∪ anonymous model" union type. Build white-label models as their own subsystem **on top of the existing model registry skeleton**.

---

## Industry standard (research)

The robust + unified pattern is the **LLM-gateway model-alias** pattern. Reference impl: LiteLLM (also Portkey, OpenRouter, Kong/Vercel AI Gateway). Four mechanisms, all standard:

1. **Alias map** — public `model_name` → hidden `litellm_params` (real `model`, provider, `base_url`, key). Client only ever uses the alias.
2. **Request rewrite** — gateway swaps alias → real model in the outgoing body (a plain reverse proxy can't; rewriting the provider-specific body is the gateway's job).
3. **Response rewrite** — gateway rewrites the response `model` field back to the alias, **including streaming chunks**. Confirmed current default: LiteLLM ≥ 1.81 returns the alias; a bug report asking to restore the upstream name was closed **not-planned** (BerriAI/litellm#21665). So alias-in-response is intended behaviour, not a workaround.
4. **Listing / observability** — `/models`, `/model/info` expose the alias only (BerriAI/litellm#5524). Cost/usage public-tagged on the alias; the real model stays in the private/admin ledger.

**Limit the industry accepts:** gateways hide *identity/metadata*, NOT behavioural side-channels (tokenizer, context window, latency, refusal style). The bar is "no identifying metadata", never "behaviourally indistinguishable". This issue adopts the same honest bar.

Sources:
- https://docs.litellm.ai/docs/completion/model_alias
- https://github.com/BerriAI/litellm/issues/21665 (response returns alias; closed not-planned)
- https://github.com/BerriAI/litellm/issues/5524 (aliases must show in /models)
- https://docs.litellm.ai/docs/wildcard_routing
- https://openrouter.ai/blog/insights/llm-gateway/

---

## Leak-surface inventory (audited against current code)

Two threat models. Be explicit about which each surface serves.

**Threat A — dashboard user** (selects a model in the UI, reads run detail). Fully closeable.

| Surface | file:line | Leak | Action |
|---|---|---|---|
| `GET /api/models` wire schema | `apps/api/src/openapi/paths/models.ts` (`OrgModel` schema) + route `apps/api/src/routes/models.ts:121` | `apiShape`, `baseUrl`, `modelId`, `credentialId`, `contextWindow`, `maxTokens`, `reasoning` all on the wire | For white-label models, project to `{ id(alias), label }` (+ optional Appstrate-owned context/cost or null); strip the rest |
| Featured picker | `apps/api/src/data/featured-models.json` | real ids (`deepseek-chat`, `gpt-…`) | exclude white-label backings; weekly auto-gen must not surface them (see memory: featured-models.json is auto-overwritten) |
| Org-models service | `apps/api/src/services/org-models.ts` (`listOrgModels` / `resolveModelMetadata`) | full metadata cascade | add a white-label projection step |
| Settings UI | `apps/web/src/pages/org-settings/models.tsx`, `apps/web/src/components/model-form-modal.tsx`, `apps/web/src/hooks/use-models.ts` | renders `modelId`/provider/baseUrl | hide for white-label entries |
| Run detail | `apps/api/src/openapi/paths/runs.ts` → `model_label`/`model_source` | **already opaque** ✅ (stores label, not modelId) | none |
| Agent model config | `apps/api/src/routes/agents.ts:148` `GET/PUT …/model` → `{ modelId }` | the alias id is fine to store/return; ensure it's the alias, not the real id | verify resolution happens server-side only |

**Threat B — adversary who can read/control the agent runtime** (custom agent code, container logs, raw LLM responses). Reducible, NOT fully closeable.

| Channel | file:line | Leak |
|---|---|---|
| Agent container env | `runtime-pi/entrypoint.ts:583,587-588` (`MODEL_ID`), header `MODEL_API` | real model id + protocol family handed to the agent |
| Base URL | `apps/api/src/services/orchestrator/sidecar-env.ts` (`PI_BASE_URL`) | provider endpoint (`api.deepseek.com`) |
| **Upstream response echo** | `runtime-pi/sidecar/app.ts` `passUpstream` (verbatim forward; key-swap precedent at `:470-471`) | provider returns `"model":"…"` → agent sees the real id (non-stream **and** SSE chunks) |
| `llm_usage` at rest | `packages/db/src/schema/runs.ts:452` `model`, `:455` `realModel`, `:458` `api` | real id retained for billing/audit (admin only, **not** via a user API today) — keep private; never expose on a user surface |

---

## Proposed design (build on the existing model registry)

1. **Virtual / white-label model entry.** A model registry entry (env via `SYSTEM_PROVIDER_KEYS`-style, and/or a DB `models` row with a `whiteLabel`/`virtual` flag) where the public `id` is the alias (`appstrate-medium`) and the binding `{ providerId, modelId, apiShape, baseUrl, credentialId, contextWindow, maxTokens, cost }` is hidden. Reuse `loadSystemRegistry`.
2. **Listing projection.** `listOrgModels` / `/api/models` returns only `{ id, label }` (+ optional Appstrate-owned numbers) for white-label entries. Update `OrgModel` OpenAPI schema so the white-label projection is a documented response shape; run `bun run generate:api`.
3. **Bidirectional `model` swap in the sidecar** — the model analogue of the existing API-key placeholder swap (`runtime-pi/sidecar/app.ts:470`):
   - request body `model: <alias>` → `<real>` before forwarding upstream;
   - response `model: <real>` → `<alias>` on the way back, **including each streaming chunk** (`passUpstream` / the SSE path).
   - The agent then only ever sees the alias. `MODEL_ID`/`MODEL_API` given to the container = alias + a neutral value where possible; `PI_BASE_URL` = an Appstrate endpoint, not the provider's.
4. **Usage ledger.** Keep `realModel`/`api` in `llm_usage` (private, billing/audit). User-facing usage/cost is tagged to the alias/label (run already uses `model_label`).

---

## Scope / tasks

- [ ] Virtual-model registry entry + Zod schema (hidden binding) on top of `loadSystemRegistry`; resolution `alias → real` server-side only.
- [ ] `/api/models` + `org-models.ts` projection to `{ id, label }` for white-label; OpenAPI `OrgModel` update + `bun run generate:api`.
- [ ] Frontend: hide `modelId`/provider/baseUrl for white-label entries (models settings page, model-form modal, model picker).
- [ ] Sidecar bidirectional `model` swap (request alias→real, response real→alias, **streaming included**); neutralise `MODEL_ID`/`MODEL_API`/`PI_BASE_URL` exposure to the container.
- [ ] Keep `llm_usage.realModel`/`api` private; verify no user-facing endpoint returns them.
- [ ] Featured-models: ensure white-label backings never surface.
- [ ] Tests: list projection hides real fields; sidecar swap rewrites request + non-stream response + streaming chunks; run record/usage shows alias only; a security test asserting the real model id never appears in any user-facing response (mirror the integration test `expect(JSON.stringify(body)).not.toContain(realId)`).

## Non-goals / honest caveats

- **Not** behaviourally indistinguishable: tokenizer, context window, latency, refusal style remain observable. We hide *metadata*, matching the industry bar. Don't market "zero leakage" absolutely.
- **Not** a merge of integration-masking and model-anonymisation into one abstraction (see *Why not unify*). Reuse the shared registry skeleton only.
- Threat B (adversary controlling the agent runtime) is *reduced*, not eliminated.

## Acceptance criteria

- A white-label model selected by a dashboard user surfaces only its alias + label across `/api/models`, agent config, run detail, usage — verified by a test that greps the real model id / provider / base URL out of every user-facing response.
- An agent run against a white-label model: the agent's request/response `model` field (incl. streaming) is the alias; the real id appears only in the private `llm_usage` ledger.

pierrecabriere:
Closes #727.

Adds **model aliases**: an operator exposes a model under an Appstrate-branded vanity name (e.g. `appstrate-medium`) while the real backing model — a fine-tune on DeepSeek, a GPT variant, whatever — stays hidden server-side. The LLM-gateway alias pattern (LiteLLM / Portkey / OpenRouter / Kong). A dashboard user picks `appstrate-medium` and never learns what's behind it.

> Named **model alias** (the industry-standard term, confirmed via research) rather than "white-label" — the flag/column is `aliased`, which also avoids billing/tier vocabulary in the Apache-2.0 core.

Built **on top of the existing model registry**, not a new abstraction — the issue's "do not unify with integration-masking" decision is respected: the only shared layer is the `loadSystemRegistry` skeleton. The alias **is** the registry `id`, so resolution (alias→real) falls out for free; everything else is projection + a sidecar swap.

## Phases (one commit each, each green on `bun run check`)

1. **Registry flag + DB column + resolution** — `aliased` on the `SYSTEM_PROVIDER_KEYS` model schema + `ModelDefinition`, an `org_models.aliased` column (migration 0015), carried through `listOrgModels`. `ResolvedModel` gains `aliased` + `aliasId` (public id) alongside the real `modelId`.
2. **Listing projection (Threat A — dashboard user)** — `projectAliasedModel()` strips the real binding (provider/protocol, endpoint, upstream id, credential, AND catalog-derived capability/cost — a distinctive window/price would itself fingerprint the backing) at `GET /api/models` + the effective-default response. NOT inside `listOrgModels`, so operator create/update still see the full resource. `OrgModelInfo` binding fields nullable; web hides them behind an "Alias" badge (fr/en).
3. **Sidecar bidirectional `model` swap (Threat B reduction)** — the container is handed the alias as `MODEL_ID`; the sidecar rewrites `model` alias→real upstream and real→alias back, for non-stream JSON **and** streaming SSE (frame-buffered, Anthropic `message_start` nesting handled). Match by exact value at known JSON paths — never a blind string replace, so a model id inside generated content is never clobbered. A model alias forces the sidecar (never skip); the real id/endpoint/key never enter the container. `MODEL_API` (protocol family) necessarily stays — documented honest caveat.
4. **Featured-picker exclusion** — the weekly generator excludes alias backings (explicit `FEATURED_MODELS_EXCLUDE` list ∪ `SYSTEM_PROVIDER_KEYS` aliased entries). Exclusion lives in the generator (survives the weekly overwrite), logs what it drops.
5. **Private usage ledger** — test locks `listLlmUsageForRun` to its id/costUsd/source projection so `real_model`/`api` (retained for billing/audit) can't leak via a future column add.

## Honest bar (non-goals)

- Hides *metadata*, not behaviour: tokenizer, context window, latency, refusal style remain observable — matches the industry bar.
- Threat B (adversary controlling the agent runtime) is **reduced, not eliminated**: `MODEL_API` is still handed to the agent (it needs the protocol family to format requests).
- Not a merge of integration-masking and model-anonymisation — shared layer is the registry skeleton only.

## Tests

50 alias tests, all green: registry parse, projection pure-fn + route security (real id absent from the list, present on the operator create response), DB list/resolve, swap pure-fns (incl. chunk-split SSE + content-not-clobbered), `/llm` proxy end-to-end (request real upstream, alias back, SSE), launcher wiring (sidecar forced, swap descriptor, alias `MODEL_ID`, real id absent from agent env), featured exclusion, private-ledger guard. `bun run check` passes (31/31 tasks).


pierrecabriere:
## Second-pass critical review + hardening (`2adfaa36e`)

Ran a deep multi-angle re-review (security / DRY-KISS / test-coverage / docs / data-path). It found that the first pass **did not actually close Threat A** — the dashboard/API leak was open through two surfaces the initial implementation missed. All verified leaks are now fixed.

### Verified leaks closed
| Sev | Where | Leak | Fix |
|---|---|---|---|
| 🔴 Critical | `llm-proxy/core.ts` (`/api/llm-proxy/*`) | a **second** user-reachable inference path echoed the real `model` id verbatim in JSON/SSE responses **and cached it** | rewrite real→alias on every branch + cached body; scrub error prose |
| 🟠 High | run DTO (`run-pipeline.ts` → `state/runs.ts`) | real backing `modelCredentialId` cross-references to the backing model via `/api/model-provider-credentials` | nulled for aliased runs |
| 🟡 Med | `routes/models.ts` create | omitted label → derived from the real backing → survives projection | require explicit label (route + `SYSTEM_PROVIDER_KEYS` registry) |
| 🟡 Med | swap | aliases on url-model protocols (google/azure/bedrock) silently broke & never swapped | rejected up front via `isAliasableApiShape` |

### Quality
- **DRY:** swap logic lifted to a single shared `@appstrate/core/model-swap`, consumed by both the sidecar and the gateway (sidecar re-exports it) — previously the gateway had no swap at all.
- **Defense-in-depth:** `projectAliasedModel` switched denylist → allowlist (a future `OrgModelInfo` field defaults to hidden).
- **KISS:** dropped the dead trailing-empty guard in the SSE transform.
- **UI:** hide Edit for aliased custom models (projected binding can't round-trip).

### Docs (the review asked "is there a guide?" — there wasn't)
- New operator guide `docs/architecture/MODEL_ALIASES.md` (both creation paths, constraints, threat model).
- `ENV.md` now documents the `aliased` flag + `FEATURED_MODELS_EXCLUDE`.
- `@appstrate/core` CHANGELOG entry for the new export.

### Tests (+ fixed a pre-existing typecheck break in the Phase 5 ledger test)
New: gateway swap (JSON / SSE / error-scrub / ledger keeps real id), create guards (label-required, url-protocol rejected), registry skip-on-invalid, and swap-module gaps (multibyte chunk split, flush no-trailing-newline, exact-vs-substring match, `scrubModelText`, `isAliasableApiShape`).

`bun run check` green (31/31). Alias suites pass serially: 18 (llm-proxy) + 23 (models) + 1 (ledger) + 21 (sidecar swap) + 9 (unit).

### Remaining honest caveat (unchanged)
Threat B (adversarial in-container agent) is **reduced, not eliminated** — `MODEL_API` (protocol family) stays observable. Closing it fully needs a protocol-normalizing gateway, out of scope here.

pierrecabriere:
### Second-pass review — one additional leak found & fixed (`bf7f1d36c`)

Re-audited the full diff with a focus on *every JSON location the model id can occupy per protocol*, not just the paths the first pass touched.

**Finding (🔴, fixed):** the swap rewrote only top-level `model` and Anthropic's `message.model`. But `openai-responses` and `openai-codex-responses` (codex) are both in `ALIASABLE_API_SHAPES`, and the **Responses API streams its model id nested under `response.model`** (`response.created` / `response.completed` snapshot events), never top-level. The sidecar serves those shapes for every agent run, so an aliased codex/responses model leaked the real backing id in the SSE stream — we *claimed* the shape was aliasable but couldn't fully mask it.

Fix: extended the shared `rewriteModelRealToAlias` (one function, used by both the sidecar and the gateway) to cover `response.model` alongside `message.model`. Added non-stream + streaming tests for the Responses snapshot shape.

**Everything else verified sound** — no further changes needed:
- **Gateway** registers only `openai-completions` / `anthropic-messages` / `mistral-conversations` (all top-level `model`), so it was fully covered already; Responses shapes are run-only (sidecar).
- DRY (single `@appstrate/core/model-swap`), KISS, allowlist projection, run-DTO `modelCredentialId` nulling, create/registry guards (label required + aliasable-shape check), UI edit-hidden, and the operator guide (`docs/architecture/MODEL_ALIASES.md`) all hold up.
- `bun run check` green (31/31); alias suites green (48 integration+unit, 23 sidecar swap).

Threat B caveat unchanged: `MODEL_API` (protocol family) stays observable to in-container code.

pierrecabriere:
### OAuth codex alias verification — second leak found & fixed (`29a938e28`)

Traced an aliased **codex** (`openai-codex-responses`, OAuth subscription) end to end. Wiring is correct: `modelSwap` rides into the sidecar's OAuth config (`pi.ts:192`), request `model` is swapped alias→real *after* the wireFormat `transformBody`, and every OAuth return branch (401-retry fallback + final) passes the swap to `passUpstream`. Streaming responses are now masked by the `response.model` fix above.

**Finding (🔴, fixed):** the platform gateway blind-scrubs error-prose bodies, but the in-container sidecar's `passUpstream` ran only the exact-field `swapResponseModelJson` — which misses the real id when it sits in free-form error prose (`"the model \`deepseek-chat\` does not exist"`) instead of a top-level `model` field. An upstream 4xx/5xx for any aliased model (codex included) leaked the backing id straight to the agent, breaking the Threat-B "error prose scrubbed" guarantee documented in `MODEL_ALIASES.md`.

Fix: error branch in `passUpstream` blind-scrubs any non-2xx body (safe — error bodies carry no generated content), mirroring the gateway. Both api_key and OAuth paths funnel through `passUpstream`, so one branch covers both. 2xx keeps the exact-field swap. Added an api_key proxy test for the prose-error case.

`bun run check` green (31/31); sidecar swap+proxy suites green (24).

PR #688, opened 2026-06-15
feat: subscription model engines (Claude Code + Codex) for chat & agents via official vendor tools — no fingerprint forging

otarbes:
## Summary

Adds **subscription model engines** for **Claude** (Claude Pro/Max via `claude-code`): running a **chat turn** and an **autonomous agent run** on a user's own subscription, always through the **official vendor tool** (the Claude Agent SDK / `claude` binary), with **no fingerprint forging**. Plus the removal of all OAuth-forging primitives and a recorded compliance posture.

> **Scope note (was rescoped):** this PR originally also carried the **Codex** agent runner (token-vend + egress-lock mechanism). That slice was **split out to follow-up PR #767** to keep this PR focused on the more complete, more ToS-defensible Claude path. In this PR, **codex is an inference / model provider only** — connectable + models listable, but a codex agent run is hard-refused (`UnrunnableOauthProviderError`) and codex has no chat surface.

85 commits · 209 files · ~+22.7k/−2.7k.

---

### 1. Packaged chat module (`packages/module-chat`)
Opt-in workspace surface (append to `MODULES`). Conversational UI + streaming backend, engine-agnostic on the AI-SDK UI message stream so the client doesn't care which engine answers. Consumes the platform only over its own public REST/MCP surface (loopback-authed), never importing apps/api internals.

### 2. Claude Code — chat **and** agents (official Agent SDK, no forging)
- **Chat:** drives `@anthropic-ai/claude-agent-sdk` `query()` in-process; SDK message stream → UI message stream mapper; in-process MCP for local tools. (`packages/module-claude-code`)
- **Agents:** `packages/runner-claude` (`ClaudeAgentRunner`, AFPS Runner) drives the official `claude` binary in the sandboxed run container; the sidecar gains an **oauth-passthrough `/llm` mode** that swaps the user's own bearer + ensures the documented `oauth-2025-04-20` beta — nothing else. The real subscription token never enters the agent container.
- The binary signs its **own** client fingerprint — we forward its user-agent / x-app / anthropic-beta untouched.

### 3. Removed ALL OAuth-subscription fingerprint forging
Identity headers, system-prepend, `wireFormat` body transforms, originator spoofing — gone (`a2c664f7c`). There is **no forging fallback**: a subscription credential that resolves to a non-official engine is hard-refused (`UnrunnableOauthProviderError`), never forged onto the Pi loop.

### 4. Engine selection (one place)
[...]

PR #757, opened 2026-06-24
refactor: architectural simplifications for the subscription-model-engine PR

pierrecabriere:
Architectural simplifications for PR #688 (subscription-model-engine), stacked onto `feat/module-chat`. Driven by a cross-analysis (6 Opus subagents + Codex CLI) — full report in `claudedocs/pr-688-simplifications.md`.

**Net: −677 lines (836+/1513−), 45 files.** `bun run check` green (34/34); targeted unit+integration suites green (252 tests).

## Applied

| # | Change | Effect |
| --- | --- | --- |
| **S1** | Split the integration-connection **readiness** refactor out (bulk `connection-readiness` endpoint + resolver/web/test churn) — restored the per-integration `agent-resolution` endpoint. **Kept** the MCP `get_me` / operation-index / `/api/me/context` work (genuinely coupled to chat). | biggest scope cut; readiness can land as its own PR |
| **S2** | Factored the duplicated Claude `tool_result` block parsing into a shared `parseToolResultBlocks` (`runner-claude/claude-blocks.ts`), used by both the agent mapper and the chat ui-stream mapper. | removes drift risk on the SDK block field names |
| **S3** | Deleted the `subscription-gateways` provider-id→handler registry + the core `chatGateway` field; mount `/claude-code-sdk` directly, gated on the `claude-code` engine being registered. | −1 file, −1 core contract field |
| **S4** | Inlined `subscription-token.ts` (single caller) into the gateway as a claude-specific resolver. | −1 misleading "shared" file |
| **S5** | Removed `finalizeFailure` from the `@appstrate/afps-runtime` public API; inlined the ~6-line failure epilogue into each runner (Pi already hand-rolled it). | smaller public runtime API |
| **S6** | Renamed `engine-select.ts` → `subscription-run-policy.ts` (it selects nothing — `engineForProvider` lives in core). | clearer boundary |
| **S8** | Deleted the `computeCodexCost`/`CodexModelCost` pass-through wrapper; call shared `computeTokenCost` directly. | −1 indirection hop |
| **S9** | Refreshed `module-chat/README.md` (post-#754 stale claims: `wait_for_run`, `X-Chat-Session-Id`, shipped model-select/thread-list, `CHAT_CLAUDE_MAX_CONCURRENCY` moved). | doc accuracy |
| **S10** | Dropped the dead `opts.system` param from `buildTranscriptPrompt`. | dead code |
| **S11** | Inlined the positional `buildSdkEnv` wrapper; deleted its now-redundant test (coverage lives in `runner-claude`). | −1 shim, −1 duplicate test |

## Evaluated and rejected (with rationale) — these failed the "same robustness/hygiene" bar

- **S6 guard-merge** — the 3 subscription guards fire fail-fast at 3 distinct launch stages; merging them into one late call would delay the early checks and lose fail-fast. Kept as-is (rename only).
- **S7 (strip `chatHandler`/`ChatEngineInput` from core)** — the `chatHandler`-on-registry indirection is exactly what keeps `module-chat` decoupled from the opt-in `module-claude-code` (module-chat has zero dep on it). Stripping it forces module-chat to import the opt-in module, breaking the zero-footprint module invariant. Only the value-positive part (drop `chatGateway`, see S3) was applied.
- **S12 (derive proxy-map from registry)** — `chat-families.ts` is deliberately dependency-free for the browser bundle (the client model-select imports it); deriving it from the server-only `proxyTarget`/provider registry would pull server code into the client bundle. Intentional allowlist.
- **S13 (loopback-bearer → `inProcess.dispatch`)** — `inProcess.dispatch` is just `getPlatformApp().fetch()`; auth + RBAC still run, so the bearer-only CSRF gate still fires (a forwarded cookie is rejected). The subscription chat path bakes the bearer into the SDK env for an HTTP-reachable gateway (a spawned/real-HTTP client can't use in-process dispatch), and `firstPartyLoopback` is the gate that stops a personal subscription being spent as a bare proxy. Removing it breaks the subscription security gate. **Security boundary — left intact.**

## Not changed (anti-findings — verified load-bearing)
Binary resolvers (correct 3-tier split, not duplicated), `token-cost`/`tool-result` shared helpers, `runtime-event` journal/drain (producer/consumer pair), `credential-vend` + egress allowlist, oauth-passthrough, the two divergent event mappers, `resolved-model-cache`.


pierrecabriere:
## Re-audit (Codex CLI + 3 Opus agents) — post-cleanup

Ran a full re-audit on the simplified branch. Result: **clean**, one real bug found and fixed.

**Correctness** (Opus): all S1–S11 changes verified correct; `tsc` green for `apps/api` + `apps/web`. S5 inlined epilogues byte-equivalent to the deleted `finalizeFailure`; S2 parser preserves both mappers' behavior; S3 gateway still loopback-gated + zero-footprint.

**Security/robustness** (Opus): all 5 invariants intact (subscription bearer-only gate, module zero-footprint, fail-fast guards, codex vend egress/port-pin, no-forging). All 4 rejected items (S6-merge / S7-destructive / S12 / S13) confirmed correctly rejected.

**Remaining simplifications** (Opus): PR now well-factored; no further real-value architectural cut survives the robustness bar (5 candidates evaluated and rejected with rationale).

**Bug found + fixed** (Codex, high-confidence): the S1 readiness split kept the **PR-added** `manifestHasRequiredAuth` backend run-blocking (a `required-auth` integration with no tools/scopes selected → `412`) while reverting the frontend that surfaced it — so the launch badge read `ready` while `POST /run` returned `412`. This required-auth blocking did **not** exist at base (verified: base resolver inert-check has no `hasRequiredAuth`, base core has no `manifestHasRequiredAuth`), so it belongs to the readiness feature and travels with it. Fully reverted it (core + resolver + 3 tests) to restore base/frontend consistency. `bun run check` 34/34, affected suites green.

PR #817, opened 2026-07-02
fix(models): stop aliased models leaking the real provider in errors, headers and DTOs

pierrecabriere:
## Context

A run on an aliased model surfaced `502 "LLM request failed: ConnectionRefused (api.deepseek.com)"` in its run log — the real backing hostname reached the agent, the run log, and the UI. An alias's whole contract is that callers/agents never learn the backing provider. A full audit of the alias masking surface found this plus several sibling leak paths.

Root cause: `ModelSwap = {alias, real}` only knew the **model id** — no masking layer knew the backing hostname or provider id, so any error prose or locally-synthesized message naming them passed through unmasked by construction.

## Changes

**Root fix — hostname-aware scrub**
- `ModelSwap` gains optional `realHost` (`packages/core/src/sidecar-types.ts`); `scrubModelText` also masks the backing hostname (replaced with a neutral `upstream` marker). New `hostnameOf` helper.
- `realHost` populated at both swap construction sites: run launcher (`run-launcher/pi.ts`) and platform gateway (`llm-proxy/core.ts`).

**Sidecar (reported bug)**
- `llmFetchErrorResponse` masks the hostname in the synthesized 502 whenever a swap is configured — unconditional, does not depend on `realHost` being populated. Covers ConnectionRefused / DNS / TLS / 30-min timeout on both api_key and oauth paths. Non-aliased models keep the hostname (debugging value).

**Platform gateway (`/api/llm-proxy/*`)**
- SSRF-refusal 400 no longer embeds the resolved `baseUrl`; OAuth-subscription rejection no longer names the `providerId`; claude-code-sdk gateway preset-mismatch / SSRF messages likewise. All details stay in server logs.
- Aliased responses now forward a strict response-header allowlist (`content-type`, rate-limit family, `retry-after`, `x-request-id`) — previously `server: cloudflare`, `cf-ray`, `anthropic-*`, `openai-organization` fingerprinted the backing on every live response. Non-aliased behavior unchanged; mirrors the sidecar and cache-replay posture.
- Mid-stream SSE **error frames** (`type:"error"` or top-level `error` key) are prose-scrubbed; content deltas remain exact-field-only (generated text mentioning the real id is never clobbered).

**Finalize boundary (defense-in-depth)**
- `finalizeRun` scrubs the runner-supplied terminal error against every alias swap visible to the org (`listAliasSwapsForOrg`, metadata-only credential resolution, degrades to no-op on failure) before it fans out to `runs.error` → run DTO, `run_update` SSE, terminal run_log, and webhook payloads.

**Credentials & schema**
- Alias-only built-in credential label no longer falls back to the provider `displayName`/`providerId` (was naming the hidden backing on `GET /api/model-provider-credentials`).
- Never-serialize guard note on `llm_usage.real_model`/`api` (persisted server-side only; no route serializes them today).

## Known-accepted (not addressed here)

- `RUN_ADAPTER=process` inherits the full platform env into the agent process (documented "no isolation" posture) — aliasing offers no protection there; real fix is env filtering, separate work.
- `MODEL_API` (protocol shape) reveals the provider *family* to the agent — SDK contract, unavoidable.

## Tests

- Sidecar: 3 new proxy tests (alias 502 hostname masked / non-alias hostname kept / `realHost` error-body scrub) + unit tests for `hostnameOf`, host scrub, SSE error-frame scrub, content-delta anti-clobber. Suite: 621 pass.
- API: 3 new metering tests (alias header allowlist / non-alias passthrough / `realHost` body scrub). Unit + gateway: 1034 pass; integration (llm-proxy, claude-code-sdk, subscription-rejection, credentials, models routes): 74 pass; finalize path: 62 pass.
- `bun run check` green (typecheck, lint, prettier, OpenAPI, type-coverage).

### DIFF
--- a/apps/api/src/routes/llm-proxy.ts
+++ b/apps/api/src/routes/llm-proxy.ts
@@ -234,6 +234,12 @@ async function handleProxy(
       throw invalidRequest(err.message);
     }
     if (err instanceof LlmProxyUnsupportedSubscriptionError) {
+      // The backing provider id is server-log-only — the caller-facing
+      // message deliberately doesn't name it (model-alias masking).
+      logger.warn("llm-proxy: rejected OAuth-subscription model", {
+        providerId: err.providerId,
+        orgId,
+      });
       throw invalidRequest(err.message, "model");
     }
     if (err instanceof LlmProxyModelApiMismatchError) {
--- a/apps/api/src/routes/llm-proxy.ts
+++ b/apps/api/src/routes/llm-proxy.ts
@@ -243,6 +243,14 @@ async function handleProxy(
       throw invalidRequest(err.message, "model");
     }
     if (err instanceof LlmProxyModelApiMismatchError) {
+      // For an aliased model the message is generic — the backing apiShape is
+      // server-log-only (model-alias masking, same posture as the branch above).
+      logger.warn("llm-proxy: model/endpoint apiShape mismatch", {
+        presetId: err.presetId,
+        expected: err.expected,
+        actual: err.actual,
+        orgId,
+      });
       throw invalidRequest(err.message, "model");
     }
     // No `status`: the upstream attempt produced no response, which the
--- a/apps/api/src/services/llm-proxy/claude-code-sdk-gateway.ts
+++ b/apps/api/src/services/llm-proxy/claude-code-sdk-gateway.ts
@@ -137,9 +137,13 @@ async function resolveClaudeSubscriptionToken(
     throw invalidRequest(`Model preset "${presetId}" is not enabled for this org`);
   }
   if (resolved.providerId !== CLAUDE_CODE_PROVIDER_ID) {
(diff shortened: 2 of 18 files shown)


pierrecabriere:
Closing in favor of a redesign: instead of scrubbing upstream error prose/headers after the fact (blacklist — every forgotten surface is a new leak), the successor PR synthesizes neutral error responses for aliased models at the proxy boundaries (whitelist by construction) and deletes the blind-scrub machinery. Same leaks closed (hostname in synthesized 502s, fingerprinting response headers, apiShape/providerId/baseUrl in gateway messages, credential label fallback), less code. Successor PR linked shortly.

pierrecabriere:
Superseded by #831 (synthesize neutral error envelopes for aliased models instead of scrubbing upstream surfaces).

PR #827, opened 2026-07-05
feat(cli): prompt for public URL on install — --app-url / APPSTRATE_APP_URL

### LINKED ISSUE #822: installer: `appstrate install` hardcodes APP_URL=http://localhost:3000 — no public-URL prompt for remote deployments
## Problem
`appstrate install` never asks for the platform's public URL. `generateEnvForTier` hardcodes:

```
appUrl = "http://localhost:3000"
...
APP_URL: appUrl,
TRUSTED_ORIGINS: appUrl,
```
(`apps/cli/src/lib/install/secrets.ts` ~L161/L167-168)

For any **remote deployment** (e.g. behind a reverse proxy at `https://example.com`) this default is wrong, and everything derived from the public URL breaks:
- OAuth/social redirect URIs
- `TRUSTED_ORIGINS` (CORS) → browser calls rejected
- verification / invitation email links
- the printed `<APP_URL>/claim` bootstrap URL

Today the operator must hand-edit `.env` after install.

## Proposal
Prompt for the public URL during install (interactive), with a flag for non-interactive:
- `--app-url <url>` (+ `APPSTRATE_APP_URL` env)
- interactive default `http://localhost:<PORT>` (current behavior preserved for local)
- set both `APP_URL` and `TRUSTED_ORIGINS` from it (TRUSTED_ORIGINS could accept a comma list later).

## Open question — port handling (needs a decision)
Public URL and host bind port are **two different things** and currently conflate:
- **`PORT`** = host-published port the container binds (`${PORT:-3000}:3000` in compose). This is the *upstream* a reverse proxy forwards to.
- **`APP_URL`** = the public URL users hit. Behind TLS this is `https://example.com` with **no port** (443, terminated by the proxy).

So for a remote deploy they diverge:
- `APP_URL=https://example.com` (no port) — what the browser/OAuth/emails use
- host `PORT=3000` (or whatever) — the proxy's upstream target, *not* in the public URL

Proposed resolution:
1. Keep `--port` as the **host publish port** (proxy upstream), default 3000 — unchanged.
2. New `--app-url` is the **public URL**, independent of `--port`.
3. When `--app-url` host is not `localhost`/`127.0.0.1` **or** scheme is `https`, auto-enable `TRUST_PROXY=true` (so client-IP / forwarded-proto are honored behind the proxy) and don't append `:PORT` to the public URL.
4. When left at default (local), behave exactly as today (`http://localhost:<PORT>`).
5. Document that the installer does **not** provision the reverse proxy / TLS — it only wires `APP_URL`/`TRUSTED_ORIGINS`/`TRUST_PROXY`; the proxy (Caddy/Traefik/nginx) forwarding `example.com → host:PORT` is the operator's responsibility (link an example).

Alternatives to weigh: accept host+port together and infer scheme; or ask "is this behind a reverse proxy?" and branch. To be defined.

## Context
Found while testing the installer on a remote host (PR #818 work). Independent of firecracker — general self-hosting DX.

pierrecabriere:
Closes #822

## Problem

`appstrate install` hardcoded `APP_URL=http://localhost:3000`. For any remote deployment behind a reverse proxy, everything derived from the public URL shipped broken — OAuth redirect URIs, `TRUSTED_ORIGINS` (CORS), email links, and the printed `/claim` bootstrap URL — until the operator hand-edited `.env`.

## What changed

The public URL is now a first-class install input, **independent of `--port`** (which stays the host bind port the proxy forwards to, per the issue's port-handling resolution):

- **`--app-url <url>` flag + `APPSTRATE_APP_URL` env** for non-interactive installs (`curl | bash`, CI). Validated strictly via `parseAppUrl` — http/https origin only, no path/query/fragment/credentials — so a typo fails at install time instead of surfacing as a broken OAuth callback later.
- **Interactive prompt** on fresh Docker-tier installs, defaulting to `http://localhost:<port>` (Enter preserves today's behavior exactly). Tier 0 stays prompt-free (local dev), but honors the flag/env for scripted use.
- **`APP_URL` and `TRUSTED_ORIGINS`** are both set from it (single value for now; comma-list can come later).
- **`TRUST_PROXY=true` auto-enabled** when the URL is non-localhost **or** https (`isRemoteAppUrl`) — TLS is always proxy-terminated in front of the platform, so forwarded-proto / client IP must be honored. Elided on local installs (defaults-not-written policy).
- **No `:PORT` appended** to a user-provided public URL — it only ever derives the port for the local default.
- **Upgrades inherit** the existing `APP_URL` (consistent with `mergeEnv`'s existing-wins semantics) and warn when a divergent `--app-url` is passed — the same contract as `--port` on upgrade.
- **Healthcheck, dev-server readiness poll, and browser open now target the local bind URL**, not the public URL: on a remote deploy the public domain doesn't route until the operator configures the proxy, which happens after the install. A post-install note states explicitly that the installer wires `APP_URL`/`TRUSTED_ORIGINS`/`TRUST_PROXY` but does **not** provision the reverse proxy/TLS, with the forwarding target spelled out.

## Scope note (per issue discussion)

Stays **single-URL**: no MinIO/S3 public-endpoint prompt. The storage follow-up (filesystem default for single-node, GitLab-style proxy-download as a separate issue) is intentionally out of scope here.

## Docs

- `examples/self-hosting/README.md` — Production Considerations rewritten around `--app-url`, the `--port` vs public-URL distinction, and a minimal Caddy example; explicit "the installer does not provision the proxy".
- Root `README.md` — remote-deploy one-liner with `--app-url`.

## Tests

- `parseAppUrl` (normalization, rejections), `isRemoteAppUrl` (loopback vs remote vs https), `generateEnvForTier` TRUST_PROXY wiring — `apps/cli/test/install/secrets.test.ts`
- `resolveAppUrl` precedence: flag > env > default, empty-env handling, tier-0 pass-through, invalid-URL fail-fast, upgrade inherit + divergent-flag warn — `apps/cli/test/install.test.ts`
- Full CLI suite: 1085 pass. `tsc --noEmit`, eslint, prettier clean; pre-push turbo gate green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,14 @@ curl -fsSL https://get.appstrate.dev | bash -s -- --tier 1 --dir ~/apps/appstrat
 
 `--tier`, `--dir`, and `--port` override the smart defaults. Equivalent env vars: `APPSTRATE_YES=1`, `APPSTRATE_PORT`, `APPSTRATE_BIN_DIR`.
 
+**Deploying on a remote host behind a reverse proxy?** Pass the public URL so OAuth redirects, CORS, and email links point at the right origin (`TRUST_PROXY` is enabled automatically for non-localhost URLs):
+
+```sh
+curl -fsSL https://get.appstrate.dev | bash -s -- --yes --app-url https://appstrate.example.com
+```
+
+Also honored via `APPSTRATE_APP_URL`. The installer does **not** provision the reverse proxy or TLS -- point your proxy (Caddy, nginx, Traefik) at `localhost:<port>` yourself; see [`examples/self-hosting/README.md`](./examples/self-hosting/README.md#production-considerations).
+
 **Want the interactive prompts?** Drop the binary without auto-launching, then run `install` yourself:
 
 ```sh
--- a/apps/cli/src/cli.ts
+++ b/apps/cli/src/cli.ts
@@ -193,6 +193,10 @@ program
     "Host port the MinIO console binds to on Tier 3 (default: 9001). Also honored via APPSTRATE_MINIO_CONSOLE_PORT.",
   )
   .option(
+    "--app-url <url>",
+    "Public URL of the platform for remote deployments behind a reverse proxy, e.g. https://appstrate.example.com (origin only). Drives APP_URL + TRUSTED_ORIGINS and auto-enables TRUST_PROXY for non-localhost/https URLs. Independent of --port (the host bind port the proxy forwards to). Also honored via APPSTRATE_APP_URL. Default: http://localhost:<port>.",
+  )
+  .option(
     "--force",
     "Bypass the 'another Compose project is already running under this name' preflight. Only use if you have already backed up the other install's data.",
   )
@@ -223,6 +227,7 @@ program
       port: typeof opts.port === "string" ? opts.port : undefined,
       minioConsolePort:
         typeof opts.minioConsolePort === "string" ? opts.minioConsolePort : undefined,
+      appUrl: typeof opts.appUrl === "string" ? opts.appUrl : undefined,
       force: opts.force === true,
       autoConfirm,
     });
--- a/apps/cli/src/commands/install.ts
+++ b/apps/cli/src/commands/install.ts
@@ -18,7 +18,9 @@ import { intro, outro, askText, confirm, spinner, exitWithError } from "../lib/u
 import {
   generateBootstrapToken,
   generateEnvForTier,
+  isRemoteAppUrl,
   isValidBootstrapEmail,
+  parseAppUrl,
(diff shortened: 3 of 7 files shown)


pierrecabriere:
Added a fail-fast guard (a8fb4ce): a plain-http loopback `--app-url` whose port disagrees with the bind port (e.g. `--app-url http://localhost:1234` with `--port 3000` or the default) is now rejected at install time with the fix spelled out — previously it produced a green install whose TRUSTED_ORIGINS, /claim banner, and email links pointed at a dead port. `https://localhost:<n>` stays allowed (local TLS-terminating proxy), remote URLs untouched.

Issue #167, opened 2026-04-18
bug(cli): `appstrate install` cannibalizes an already-running stack (hard-coded Compose project name)


pierrecabriere:
## Summary

Running `appstrate install` while another Appstrate stack is already running on the same host silently **adopts and recreates** the first install's containers, even when the new install targets a different directory. The migrate service then fails because the new install generated fresh secrets (random DB password) that don't match the already-running postgres. The user's existing stack is left in a broken half-recreated state with potential data loss.

## Repro

```
# 1. Have an existing running stack in /path/to/install-A (Tier 3)
# 2. In a fresh shell:
APPSTRATE_VERSION=1.0.0-alpha.50 appstrate install
#    → pick Tier 3 → pick /path/to/install-B as install directory
```

Output observed:

```
[+] Running 8/8
 ✔ Container appstrate-redis-1         Healthy                  13.3s
 ✔ Container appstrate-minio-1         Healthy                  12.7s
 ✔ Container appstrate-postgres-1      Healthy                  12.6s
 ✔ Container appstrate-minio-init-1    Exited                   13.0s
 ✘ Container appstrate-migrate-1       service "migrate" di...  12.8s
 ✔ Container appstrate-appstrate-1     Recreated                 1.0s
 ✔ Container appstrate-pi-pull-1       Exited                   11.6s
 ✔ Container appstrate-sidecar-pull-1  Exited                   11.6s
service "migrate" didn't complete successfully: exit 1
└  docker compose up failed with exit code 1
```

Notice `appstrate-appstrate-1` went `Recreated` — this is install A's main container being killed and rebuilt against install B's `.env`. The `Healthy` lines on postgres/redis/minio are misleading: those are install A's existing containers, reported as healthy because Compose's config hash happened to match.

## Root cause

1. `examples/self-hosting/docker-compose.tier{1,2,3}.yml` hard-codes `name: appstrate` at the top of each template.
2. `apps/cli/src/lib/install/tier123.ts::dockerComposeUp()` calls `docker compose up -d` from the new install directory with no `--project-name` override.
3. Docker Compose uses the project name (`appstrate`), not the working directory, to identify a stack. Two installs in different directories → one Compose project → container name collision → Compose \"adopts\" existing containers and recreates on config drift.

## Impact

[...]

PR #830, opened 2026-07-05
storage: proxy uploads via app domain + filesystem default for single-node installs

### LINKED ISSUE #829: storage: single-URL file serving — proxy-download mode for the S3 backend + filesystem default for single-node installs
Follow-up to #822 (see the storage/MinIO survey in the issue comments) and PR #827, which deliberately stayed single-URL (`--app-url` only, no MinIO prompt).

## Problem

Presigned S3 URLs are consumed by the **browser**, so the S3/MinIO endpoint must be publicly reachable. Today that means:

- Tier 3 with bundled MinIO on a remote host: the operator must expose MinIO on a **second public FQDN** (S3 API must sit at the domain root — SigV4 signs the full path, subpath proxying is unsupported by MinIO, minio/minio#7857), publish `:9000`, and set `S3_PUBLIC_ENDPOINT` by hand. Nothing in the installer or docs wires this.
- Worse, out of the box it's broken even on localhost: `docker-compose.tier3.yml` doesn't publish MinIO's S3 port (`:9000`) at all — only the console (`:9001`) — so presigned URLs point at `http://minio:9000` (Docker-internal hostname) and fail in any browser.

`packages/core/src/storage-s3.ts` already has the seam: `presignEndpoint = config.publicEndpoint ?? config.endpoint` (`S3_PUBLIC_ENDPOINT`), but that only relocates the problem to the operator.

### 1. Proxy-download mode for the S3 backend (the real fix — Supabase/GitLab pattern)

Serve all file traffic through the platform on the app domain, regardless of storage backend:

- The **filesystem backend already does exactly this**: uploads/downloads proxy through `APP_URL` with HMAC-signed URLs (`UPLOAD_SIGNING_SECRET`). Generalize that app-domain signed-URL path to the S3 backend — the platform signs its own URL, then streams bytes to/from S3 server-side.
- MinIO stays fully private in the compose network (no published ports, no second FQDN) — the Supabase self-hosted architecture.
- One public URL total: `APP_URL` (wired by `--app-url` since #827).
- Must support `Range` passthrough (media playback) and streamed uploads (no buffering in memory).
- Keep **direct presign** as an opt-out for high-volume multi-node deployments where offloading bytes to S3 matters: `S3_PUBLIC_ENDPOINT` set → presign direct-to-bucket (current behavior), unset → proxy mode. This is exactly GitLab's `proxy_download` boolean, keyed on config presence instead of a new flag.

Trade-off (accepted): in proxy mode every byte transits the Bun process. Irrelevant on single-node (same machine either way); multi-node at scale opts out via `S3_PUBLIC_ENDPOINT`.

### 2. Filesystem storage as the single-node self-hosting default

MinIO adds nothing on a single node once serving is app-domain — the filesystem backend already provides the same UX with one less container and zero extra configuration (matches Supabase/Sentry/n8n/Mattermost defaults from the survey):

- Default self-hosting/single-node install → filesystem storage (`./data/storage/`, persisted volume).
- S3/MinIO becomes the advanced option where the operator brings an endpoint (external MinIO, R2, S3).
- Installer/docs updated accordingly (tier semantics: to be decided whether Tier 3 keeps bundling MinIO or becomes "bring your own S3").

### Explicitly skipped

Publishing `:9000` + defaulting `S3_PUBLIC_ENDPOINT=http://localhost:9000` (stopgap "A" from the discussion) — obsoleted by proxy mode, which keeps MinIO private instead of exposing it.

## References

Survey with product comparison table (Supabase / GitLab / Outline / Coolify / n8n) and links: #822 (comment). Related: #827 (`--app-url`), `docs/ENV.md` → `S3_PUBLIC_ENDPOINT`, `packages/core/src/storage-s3.ts`.

pierrecabriere:
Closes #829. Follow-up to #822 / #827.

## Part 1 — proxy-upload mode for the S3 backend

Presigned S3 upload URLs were the only browser-facing storage URLs (downloads already flow through the platform), and out of the box they were broken: `docker-compose.tier3.yml` never publishes MinIO's S3 port, so browsers PUT to the Docker-internal `http://minio:9000` and fail.

This generalizes the filesystem backend's app-domain signed-URL path to S3 (the Supabase/GitLab pattern):

- **`S3_PUBLIC_ENDPOINT` unset (new default): proxy mode.** `createUploadUrl()` signs an HMAC URL on `APP_URL` (`PUT /api/uploads/_content`) and the existing sink streams the body to S3 server-side — bounded-memory multipart upload, `If-None-Match: *` create-or-fail exclusivity, signed max size enforced *while* streaming (chunked-encoding safe). MinIO stays fully private in the compose network: no published `:9000`, no second public FQDN, one public URL total (`APP_URL`, wired by `--app-url` since #827).
- **`S3_PUBLIC_ENDPOINT` set: direct presign** (exact current behavior — GitLab's `proxy_download` boolean keyed on config presence). Opt-out for multi-node deployments that want upload bytes off the platform's network path.
- The signed-URL builder is extracted to `createProxyUploadDescriptor` (`@appstrate/core/storage-fs`) and shared by both backends. `writeFsUploadContent` → `writeProxyUploadContent` (it was always backend-agnostic — it dispatches through the storage layer).
- Direct SDK consumers of `createS3Storage` are unaffected: proxy mode only engages when the new optional `uploadBaseUrl`/`uploadSecret` config is provided (additive, non-breaking; CHANGELOG entry added).

Accepted trade-off (per the issue): in proxy mode every upload byte transits the Bun process — irrelevant single-node, opt-out via `S3_PUBLIC_ENDPOINT` at scale.

## Part 2 — filesystem storage as the single-node default

With serving app-domain either way, bundled MinIO adds a container without adding capability on a single node (matches the #822 survey: Supabase/Sentry/n8n/Mattermost defaults):

- `appstrate install` now defaults to **Tier 2** (PostgreSQL + Redis, filesystem storage on the persisted `storagedata` volume) — interactive default and `--yes` path.
- **Tier 3 keeps bundling MinIO** but is relabeled the advanced option (this resolves the issue's open "tier semantics" question conservatively — no template shape change, no upgrade migration, existing Tier 3 installs keep working and now have functional browser uploads via proxy mode). Bring-your-own S3 (external MinIO, R2, AWS) stays available on any tier via the `S3_*` env vars.
- Upgrades unaffected: existing installs inherit their recorded env; `mergeEnv` preserves `MINIO_*`/`S3_*`.

### Explicitly skipped (per issue)

Publishing `:9000` + defaulting `S3_PUBLIC_ENDPOINT=http://localhost:9000` — obsoleted by proxy mode.

## Notes

- **AWS-S3 behavior change (documented in `docs/ENV.md`):** deployments on real AWS S3 without `S3_PUBLIC_ENDPOINT` previously got direct presigned AWS URLs; they now proxy uploads through the platform. Set `S3_PUBLIC_ENDPOINT=https://s3.<region>.amazonaws.com` to restore direct presign.
- Range passthrough for downloads (mentioned in the issue) has no work to do today: no storage bytes are served to browsers via presigned GET anywhere — downloads were already platform-side only.

## Tests

- `packages/core/test/storage.test.ts` — new proxy-mode suite (app-domain URL shape, token binds key/size/mime, trailing-slash trim, traversal rejection, `publicEndpoint` wins over proxy config). 698 core tests green.
- `apps/api` — new integration test pinning that `POST /api/uploads` returns an app-domain sink URL (regression guard for presigned URLs leaking through); upload/consume/sink suites green (tier 0).
- `apps/cli` — tier-default tests updated (`--yes` → 2, interactive default → 2, explicit `--tier N` untouched). 1061 CLI tests green.
- `bun run check` (turbo: tsc + eslint + prettier + verify-openapi + api-types) green; `verify-compose-defaults` green.


pierrecabriere:
## Self-review (adversarial pass)

Findings ranked most-severe first; all fixed in 8c9c79da0 unless noted.

### 🔴 Blocker — tier-downgrade hazard on upgrade re-runs (fixed)

With the default flipped to Tier 2, re-running `appstrate install --yes` (or pressing Enter at the tier prompt) on an existing **Tier 3** deployment rewrote `docker-compose.yml` with the Tier 2 template: the MinIO service vanished while `mergeEnv` preserved the S3 env — every stored object (package ZIPs, uploads, run artifacts) became unreachable, silently. The `miniodata` volume survives, but the stack no longer runs anything that can read it.

The reverse direction predated this PR (Tier 2 install + old Tier 3 default → compose gains MinIO, storage flips filesystem→S3, all existing files hidden) — the root cause is that `install` re-runs never pinned the tier.

**Fix**: on upgrade, an unexpressed tier now **inherits the installed stack's tier** — `inferInstalledTier()` reads the compose service set (minio→3, redis→2, postgres→1, unrecognized/absent→null), `reconcileUpgradeTier()` is pure and unit-tested (13 new tests, including against the real tier templates), explicit `--tier N` still wins for scripted tier changes. Same warn-and-inherit contract as ports (#167) and APP_URL (#827).

### 🟡 Branch was based on pre-#827 main (fixed)

#827 merged after this branch was cut and touches the same CLI files. Git reported MERGEABLE, but CI green on the merge commit ≠ reviewed against reality. Rebased; all suites re-run post-rebase.

### 🟡 Doc drift (fixed)

- Root `README.md` still said `--yes` picks Tier 3 — updated.
- `bootstrap.sh` usage example aligned to `--tier 2`.
- My own compose comment claimed the MinIO console is published "on localhost" — it binds all host interfaces (`${MINIO_CONSOLE_PORT:-9001}:9001`). Reworded. (Actually binding it to 127.0.0.1 would be a further hardening, but it's pre-existing behavior and out of scope.)
- OpenAPI upload example showed a presigned S3 URL as the nominal case; now shows the app-domain sink URL. `detect:breaking` passes — descriptions/examples only, no wire change.

### ✅ Claims verified during review (no change needed)

- **Multipart abort on failure**: `@aws-sdk/lib-storage` defaults `leavePartsOnError` to `false` (checked in the resolved package source) — a failed proxy-mode stream auto-aborts the MPU, no orphaned parts, no partial object. The docblock claim is accurate.
- **"Downloads already platform-proxied"**: re-verified — no `GetObject` presigner usage anywhere, no browser-facing storage GET; run documents are fetched by the runtime via HMAC routes. Uploads were genuinely the only presigned browser URLs, so the issue's Range-passthrough requirement has no work to do today.
- **Sink parity in S3 proxy mode**: size ceiling enforced *while* streaming (chunked-safe), `If-None-Match: *` gives the same replay protection as FS `O_EXCL`, first-consume failure rollback (delete object → release claim) is idempotent on S3.
- **False alarm chased**: the root README tier table says Tier 2 executes runs as Bun subprocesses — that describes the **dev** tiers. The self-hosting Tier 2 template keeps `RUN_ADAPTER=docker` + socket mount, so the new default loses zero sandbox isolation.

### ⚠️ Accepted trade-offs (called out, not changed)

- **AWS S3 behavior change**: deployments on real AWS without `S3_PUBLIC_ENDPOINT` silently switch from direct presigned AWS PUTs to proxying through the platform. Working-as-designed per the issue (config-presence keying), documented in `docs/ENV.md` + both `.env.example`s — but worth a release-note line.
- `createProxyUploadDescriptor` lives in `storage-fs.ts` (where the token keyring already lives), so `storage-s3` now imports `node:fs` transitively. Server-side-only modules; splitting a `storage-upload-token` module would churn the published core export map for no consumer benefit today.
- Older S3-compatible stores without conditional-write support lose the sink's replay protection (pre-existing caveat, already documented in core).

Post-fix state: `bun run check` ✅, `detect:breaking` ✅, CLI 1103 ✅, core 698 ✅, API upload suites ✅.

PR #836, opened 2026-07-06
perf(firecracker): cut guest cold start 907→325 ms — baked transpiler cache + host readahead

### LINKED ISSUE #835: firecracker: guest cold start dominates boot latency (~900ms Bun+ESM vs ~200ms docker) — snapshot/resume + rootfs readahead follow-ups
Follow-up to #818. Measured on the preview deployment (`preview-pr818`): platform on Coolify, `appstrate-runner` daemon on the same bare-metal host (Ryzen 7940HS, Ubuntu 24.04), vs local dev (M4 Pro, `RUN_ADAPTER=docker`).

## Measurements

Same agent (`@pierre-cabriere/counter`), timings from the runtime's own breadcrumbs (`performance.now()` from process entry — VM boot ~110 ms is *not* included in these numbers):

| | docker local (M4 Pro) | firecracker (Ryzen 7940HS) |
|---|---|---|
| process cold start (Bun start + ESM imports) | **204 ms** | **907–908 ms** |
| post-start phases (provisioning, bundle, MCP connect) | 419 ms | **~50–90 ms** |
| `runtime ready` total | 623 ms | 954–996 ms |

Reproduced on two runs (`run_0646f27b`, engine claude; `run_f1ded755`, engine pi — cold start identical at 907/908 ms, so it is not engine-dependent).

The guest is actually *faster* than local after process entry (provisioning is near-instant on the flat guest→platform path). The entire gap lives in the Bun process cold start: **~700 ms of pure guest-side startup overhead per run**.

## Causes

1. **Cold page cache per run** — VM-per-run means every boot starts with an empty guest page cache; every ESM import reads `rootfs.ext4` through virtio-blk. Docker keeps host page cache + image layers warm across runs.
2. **Single-thread CPU** — Bun startup + ESM eval are CPU-bound; contributes a constant factor but does not explain 4.4× alone.
3. **2-vCPU contention** — sidecar and agent cold-start concurrently and share 2 vCPUs during exactly this window (the reason for the vCPU floor documented in #818).

## Levers

- **Snapshot/resume** (the real fix, already listed as a #818 follow-up): boot the guest once to the "supervisor up, Bun warm" point, snapshot, resume per run (~10 ms class). Eliminates both kernel boot *and* the Bun/ESM cold start.
- **Host-side rootfs readahead** (cheap intermediate): daemon warms the host page cache for `rootfs.ext4` at boot (e.g. `posix_fadvise`/`readahead` or a one-shot sequential read). Guest virtio-blk reads then hit host RAM instead of disk. Does not help the ESM *eval* cost, only the I/O share.
- **Preloaded/bundled entrypoint**: a single pre-bundled JS artifact for entrypoint+runner would cut module-graph resolution and file count (many small reads → one large sequential read, the best case for virtio-blk).
- Possibly stagger sidecar/agent start by a few ms or bump vCPUs during boot if contention measurably dominates (profile first).

## Acceptance

- `runtime starting (…ms cold start)` on firecracker within ~2× of the docker adapter on comparable hardware, or snapshot/resume lands and the metric becomes moot.

Not a regression — expected trade-off of the VM-per-run design; filing so the ~700 ms/run doesn't get forgotten once #818 merges.

pierrecabriere:
Closes #835.

## Problem

VM-per-run pays the full Bun cold start on every boot: **907–908 ms** measured on the preview host (Ryzen 7940HS) vs **204 ms** for the docker adapter — almost entirely parse/eval of the bundled entrypoint before the first breadcrumb, since the guest page cache is empty on every run.

## Changes

**1. Baked Bun transpiler cache** (`runtime-pi/Dockerfile`, `guest/supervisor.ts`)
- `BUN_RUNTIME_TRANSPILER_CACHE_PATH=/runtime/.transpiler-cache` set in the image, primed at build time: one entrypoint run (env-validation exit, after the full static graph evaluates) + explicit imports of the lazily-loaded SDKs → 19 `.pile` files (6.7 MB). Bun then mmaps instead of re-parsing every >50 KB module on each start.
- Build fails if priming yields an empty cache (guards against a silent regression if the entrypoint ever exits before the graph loads).
- Docker containers inherit the ENV; the firecracker guest gets the variable injected by the supervisor (agent env comes from the config drive, not image ENV). Config-drive env can still override.

**2. Host page-cache readahead** (`runner/readahead.ts`, `runner/daemon.ts`)
- The daemon sequentially reads `rootfs.ext4` + `vmlinux` once at boot (background, non-blocking): after a host reboot the first run's guest virtio-blk reads hit host RAM instead of cold disk. Best-effort by contract — logs, never throws. Unit-tested (multi-chunk read + missing-file continue).

## Measured (preview deployment, same agent/host as #835)

| | before (#835) | after |
|---|---|---|
| cold start | 907–908 ms | **325 / 325 / 329 ms** (−64%) |
| runtime ready | 954–996 ms | **788 ms** |
| vs docker local (204 ms) | 4.4× | **1.6×** — acceptance (≤2×) met |

Readahead observed live: rootfs 1850 MiB warmed in 407 ms at daemon boot.

## Rejected approaches (evaluated empirically)

- `bun build --bytecode`: requires `--format=cjs`, which forbids the entrypoint's top-level await AND cannot `require()` the Pi SDK (ESM-only exports map).
- `bun build --compile`: externals resolve from `/$bunfs`, breaking on-disk node_modules resolution entirely.
- The transpiler cache achieves the same parse-skip with zero module-format changes.

## Degradation model

Every failure path degrades to performance, never correctness: stale/corrupt/absent cache → Bun re-parses (cache keyed by content hash + Bun version, per-arch primed in each arch's own image build); unwritable cache dir → writes skipped; readahead failure → warn + continue.

## Remaining (out of scope)

- `sdkImportMs` ~400 ms residual: the external Pi SDK dist files are <50 KB each, under Bun's cache threshold. Runs concurrently with provisioning, so not the critical path today.
- Snapshot/resume stays the endgame fix (phase 3, post-jailer hardening).

### DIFF
--- a/apps/api/src/modules/firecracker/guest/supervisor.ts
+++ b/apps/api/src/modules/firecracker/guest/supervisor.ts
@@ -39,6 +39,15 @@ const SIDECAR_BIN = "/usr/local/bin/sidecar";
 const RUNNER_EXEC_WRAPPER = "/usr/local/bin/appstrate-runner-exec";
 const AGENT_ENTRY = "/runtime/dist/entrypoint.js";
 const CONFIG_PATH = "/config/config.json";
+/**
+ * Pre-warmed Bun transpiler cache baked into the rootfs at image build (see
+ * runtime-pi/Dockerfile). The guest page cache is empty on every boot
+ * (VM-per-run), so skipping the re-parse of the bundled entrypoint + Pi SDK
+ * is a direct cold-start win. Docker containers inherit this from the image
+ * ENV; here the agent env comes from the config drive, so the supervisor
+ * injects it. New entries written at runtime land in the tmpfs overlay.
+ */
+const TRANSPILER_CACHE_PATH = "/runtime/.transpiler-cache";
 
 function log(msg: string): void {
   process.stdout.write(`[supervisor] ${msg}\n`);
@@ -254,7 +263,7 @@ async function main(): Promise<void> {
   const agent = spawnAs(
     GUEST_AGENT_USER,
     cfg.agent.argv ?? ["/usr/local/bin/bun", "run", AGENT_ENTRY],
-    cfg.agent.env,
+    { BUN_RUNTIME_TRANSPILER_CACHE_PATH: TRANSPILER_CACHE_PATH, ...cfg.agent.env },
     "/workspace",
   );
   log(`agent pid ${agent.pid}`);
--- a/apps/api/src/modules/firecracker/runner/daemon.ts
+++ b/apps/api/src/modules/firecracker/runner/daemon.ts
@@ -32,6 +32,7 @@ import { getFirecrackerEnv } from "./host-env.ts";
 import { getRunnerEnv } from "./env.ts";
 import { createRunnerApp } from "./server.ts";
 import { ensureGuestArtifacts } from "./artifacts.ts";
+import { warmHostPageCache } from "./readahead.ts";
 import { verifyGuestPath, type GuestPathResult } from "./net-probe.ts";
 import { getErrorMessage } from "@appstrate/core/errors";
 import { logger } from "./logger.ts";
@@ -73,6 +74,15 @@ try {
 } catch (err) {
   fatal("guest artifacts", err);
 }
+
+// Warm the host page cache for the boot artifacts in the background
+// (issue #835): after a host reboot the first run otherwise pays cold-disk
+// virtio-blk latency for the whole guest boot read set. Deliberately not
+// awaited — a slow disk must not delay daemon readiness, and the warm is
+// pure best-effort (it logs, never throws).
+void warmHostPageCache([fcEnv.FIRECRACKER_ROOTFS_PATH, fcEnv.FIRECRACKER_KERNEL_PATH], {
+  logger,
+});
 
 // The guest-visible platform URL comes from the daemon's env, not from
 // the platform process env — on this host there IS no platform process.
--- a/apps/api/src/modules/firecracker/runner/readahead.ts
+++ b/apps/api/src/modules/firecracker/runner/readahead.ts
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Host page-cache warmer for the guest boot artifacts (issue #835).
+ *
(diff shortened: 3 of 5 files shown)


pierrecabriere:
## Robustness verification — `BUN_RUNTIME_TRANSPILER_CACHE_PATH` (source + empirical)

**Is it a standard?** It is Bun's [documented public env var](https://bun.com/docs/runtime/environment-variables) (not an internal/undocumented hook): "For files larger than 50 KB, Bun caches transpiled output into `$BUN_RUNTIME_TRANSPILER_CACHE_PATH` … content-addressable … safe to delete at any time, even while a Bun process is running."

**Cache validation** (verified in Bun's implementation, [`src/jsc/RuntimeTranspilerCache.rs`](https://github.com/oven-sh/bun/blob/main/src/jsc/RuntimeTranspilerCache.rs)):
- Key = Wyhash of **file contents**; entries additionally validate a format version constant (`EXPECTED_VERSION`), input byte length, a features hash, and output/sourcemap hashes.
- Any mismatch/corruption → entry **deleted + silent re-transpile**. Degradation is strictly perf, never correctness — matches the PR's degradation model.
- **No architecture coupling**: the cache stores transpiled JS text + sourcemap, not native code. Multi-arch images are safe (and each arch primes its own cache in its own build anyway).
- Bun upgrade → format-version/feature mismatch → clean miss. Non-issue here: prime and runtime use the same `bun` binary baked in the image.

**Docker default caveat (important):** official `oven/bun` images ship the cache **disabled** ("Bun's Docker images disable it automatically" — rationale: ephemeral FS makes a *runtime-written* cache useless). This PR's ENV overrides that deliberately: a **build-time-baked** cache is exactly the case where it pays. Verified live in the built image (env override effective, 19 `.pile` files, measurable start-time delta).

**Threshold check:** shipped Bun 1.3.x caches only files >50 KB (empirically confirmed: 10 KB file not cached, 60 KB cached). Only 1 of the 44 Pi SDK dist files exceeds it — which is the measured `sdkImportMs` ~400 ms residual. Bun `main` has lowered the constant to 4 KiB (`MINIMUM_CACHE_SIZE = 4 * 1024`), which would make 21 of those files cacheable — a free improvement on a future Bun bump, no code change needed.

PR #852, opened 2026-07-07
fix(oidc): survive cross-context resume of the PKCE flow (invite signup state mismatch)

pierrecabriere:
## Problem

Invited **new** users on SMTP-enabled instances frequently hit **"State mismatch — possible CSRF attack"** when joining an org through signup.

Deterministic chain:

1. Invite page picks the `signup` starter → `initPkceFlow` stores the OAuth `state` + PKCE verifier in `sessionStorage`, redirects to `/api/oauth/register`.
2. SMTP configured (necessarily — the invite itself arrived by email) → Better Auth `requireEmailVerification: true` → no session cookie after sign-up → "check your email" interstitial. The OAuth dance pauses.
3. The verification link opens in a **new tab or device**. `sessionStorage` is per-tab → empty there. BA auto-signs-in and resumes the flow via the pinned `callbackURL` → `/auth/callback?code&state=S1`.
4. `storedState === null` ≠ `S1` → throw. (Even past that, the PKCE verifier is equally absent, so the token exchange was doomed too.)

## Fix

Two complementary changes in `apps/web/src/modules/oidc/lib/oidc.ts`:

- **Encode the post-login redirect inside the OAuth `state`** — `base64url({ n: nonce, r: redirectTo })`. OAuth guarantees the authorize endpoint echoes `state` verbatim, so the target (e.g. `/invite/:token`) survives the server-side detour that no client-side storage can.
- **Cross-context resume branch in `handleOidcCallback`** — distinguish two cases previously conflated:
  - `storedState` present but different → **hard failure, unchanged** (real CSRF guard).
  - `storedState` never stored in this context → skip the token exchange (impossible without the verifier, unnecessary since the SPA discards tokens — the BA session cookie is the real auth) and recover the redirect from the echoed state, constrained to a same-origin relative path.

## Security analysis of the resume branch

- The `code` is **never exchanged** there → no code injection / session fixation surface. The unconsumed code simply expires.
- The caller (`auth-callback.tsx`) already runs `refreshAuth()`, which throws when no BA session exists — that session check is what authenticates the path.
- Decoded redirect is validated: relative, starts with `/`, rejects `//host` and `\` forms → no open redirect. A crafted callback URL yields at most an in-app navigation for an already-authenticated user.
- Legacy opaque states decode to `undefined` → fallback `/`.

## Coverage

| Scenario | Before | After |
|---|---|---|
| Same-tab login/signup (no SMTP pause) | OK | OK (strict path unchanged) |
| Invite signup + SMTP, link in new tab | ❌ State mismatch | ✅ lands on `/invite/:token`, accept button |
| Invite signup + SMTP, link on another device | ❌ | ✅ (BA cookie set on that device) |
| Forged `state` in same tab | rejected | rejected (unchanged) |
| Forged callback URL in fresh tab | error page | in-app navigation at most, code never used |

## Tests

New `apps/web/src/modules/oidc/lib/test/oidc.test.ts` (12 tests, DOM-less fakes following the `pairing-store.test.ts` pattern): state round-trip, nonce uniqueness, legacy-state tolerance, open-redirect guard, strict same-context exchange, CSRF rejection, missing-verifier rejection, resume-without-exchange, crafted-redirect fallback, provider-error precedence.

`bun test` ✅ · `tsc --noEmit` ✅ · eslint ✅ · prettier ✅

### DIFF
--- a/apps/web/src/modules/oidc/lib/oidc.ts
+++ b/apps/web/src/modules/oidc/lib/oidc.ts
@@ -40,8 +40,47 @@ export async function calculateCodeChallenge(verifier: string): Promise<string> 
   return base64url(new Uint8Array(digest));
 }
 
-export function generateState(): string {
-  return base64url(crypto.getRandomValues(new Uint8Array(16)));
+/**
+ * OAuth `state` — CSRF nonce + post-login redirect carrier.
+ *
+ * The state is a base64url-encoded JSON payload `{ n, r? }`: `n` is a random
+ * nonce (the CSRF property), `r` the optional post-login redirect path. The
+ * authorize endpoint echoes `state` verbatim, so `r` survives server-side
+ * detours that a new browsing context cannot: the email-verification flow
+ * (`POST /api/oauth/register` with SMTP → "check your email" → link opens in
+ * a NEW tab/device where `sessionStorage` is empty) resumes the OAuth flow
+ * via the pinned `callbackURL` and still lands on `/auth/callback` with this
+ * state — the only surviving carrier of where the user was headed.
+ */
+export function generateState(redirectTo?: string): string {
+  const payload: { n: string; r?: string } = {
+    n: base64url(crypto.getRandomValues(new Uint8Array(16))),
+  };
+  if (redirectTo) payload.r = redirectTo;
+  return base64url(new TextEncoder().encode(JSON.stringify(payload)));
+}
+
+/**
+ * Extract the post-login redirect from an echoed `state` parameter.
+ *
+ * Returns `undefined` for anything that is not a same-origin relative path —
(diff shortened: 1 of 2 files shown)


pierrecabriere:
Also fixes the same "State mismatch" error reported on **social login/signup**: the social buttons resume the flow via `callbackURL=/api/auth/oauth2/authorize${search}`, and on mobile the webview→system-browser handoff (Google rejects OAuth in webviews) lands `/auth/callback` in a context with empty `sessionStorage` — exactly the cross-context resume case. The recovery branch is cause-agnostic: any context break recovers as long as the BA session cookie exists in the landing context, which the social flow guarantees (the BA callback runs in that same context).

PR #886, opened 2026-07-10
fix(runtime-pi): run provision e2e on the engine's native platform

### LINKED ISSUE #882: provision-container.e2e: hasImage() ignores platform, turning a legitimate skip into an opaque exit-125 failure on arm64
## Symptom

On an arm64 host (Apple Silicon), `bun test runtime-pi/test/provision-container.e2e.test.ts` with `TEST_DOCKER=1` fails deterministically:

```
error: docker run failed: Unable to find image 'appstrate-pi:latest' locally
docker: Error response from daemon: pull access denied for appstrate-pi, repository does not exist
        or may require 'docker login': denied: requested access to the resource is denied
Expected: 0
Received: 125
```

It fails **whether or not** `appstrate-pi:latest` exists locally, which is what makes it confusing: `docker image inspect appstrate-pi:latest` succeeds, `docker images` lists it, and the test still claims the image is missing.

## Cause

The test launches the container pinned to amd64 (`runtime-pi/test/provision-container.e2e.test.ts:152-154`):

```ts
"--platform",
"linux/amd64",
```

but its gate does not check the platform (`:44-49`):

```ts
function hasImage(): boolean {
  try {
    return spawnSync("docker", ["image", "inspect", IMAGE], { stdio: "ignore" }).status === 0;
  } catch {
    return false;
  }
}
```

`docker image inspect` resolves by tag and is **architecture-blind**. A locally-built arm64 `appstrate-pi:latest` satisfies it, so `RUN` is true and the suite runs. Then `docker run --platform linux/amd64` cannot match the local arm64 image, tries to pull an amd64 variant from Docker Hub, and gets denied — exit 125.

The gate exists precisely to skip this test when the image is unusable. It just measures the wrong thing.

## Impact

- `bun run test:docker` and the full-suite Tier-3 run are red on any arm64 dev machine, by default, on `main`.
- The failure surfaces as `pull access denied` for a non-existent public repository, which reads like a registry/auth problem and sends you looking in the wrong place.
- CI is unaffected: `ubuntu-latest` is amd64 native, and `.github/workflows/test.yml:174` builds with `--platform linux/amd64`.

## Reproduction

```sh
docker build -t appstrate-pi -f runtime-pi/Dockerfile .   # arm64 on Apple Silicon
docker image inspect appstrate-pi:latest --format '{{.Os}}/{{.Architecture}}'   # linux/arm64
TEST_DOCKER=1 bun test runtime-pi/test/provision-container.e2e.test.ts          # exit 125
```

Workaround, and confirmation of the diagnosis — the test passes:

```sh
docker build --platform linux/amd64 -t appstrate-pi:latest -f runtime-pi/Dockerfile .
TEST_DOCKER=1 bun test runtime-pi/test/provision-container.e2e.test.ts          # 1 pass
```

(The cross-arch build goes through QEMU and is slow, but it works.)

## Suggested fix

Make `hasImage()` assert the architecture the run pins, e.g.

```ts
function hasImage(): boolean {
  try {
    const out = spawnSync("docker", ["image", "inspect", IMAGE, "--format", "{{.Os}}/{{.Architecture}}"], {
      encoding: "utf8",
    });
    return out.status === 0 && out.stdout.trim() === "linux/amd64";
  } catch {
    return false;
  }
}
```

so the existing skip path fires with an honest reason instead of the test dying inside `docker run`. The warning it already prints (`:57-61`) would then tell the developer exactly what to build.

Worth considering separately: whether the amd64 pin is needed at all, or whether the test could run the host's native architecture. The pin is justified in-comment only for CI portability (`--add-host` on Linux engines), not for the platform itself.

pierrecabriere:
Closes #882

## Problem

The provision-container e2e pinned `docker run --platform linux/amd64` while its gate (`hasImage()`) only checked image presence via `docker image inspect` — which is architecture-blind. On arm64 hosts (Apple Silicon), a natively built `appstrate-pi:latest` satisfied the gate, then `docker run` could not match the local arm64 image, tried to pull an amd64 variant of a non-existent public Docker Hub repo, and failed with a misleading `pull access denied` (exit 125). `bun run test:docker` was red by default on every arm64 dev machine.

## Why remove the pin instead of just fixing the gate

The amd64 pin had no justification:

- The platform's Docker orchestrator (`apps/api/src/services/orchestrator/docker-orchestrator.ts`) runs this image on the engine's **native** platform in production — the pin made the test diverge from real usage.
- The release pipeline (`release.yml`) publishes `appstrate-pi` as multi-arch (`linux/amd64,linux/arm64`), so there is no amd64-only dependency.
- The in-code comment about CI portability applies to `--add-host`, not `--platform`.

Fixing only the gate (the issue's suggested patch) would have turned the failure into a skip but still forced arm64 devs through a slow QEMU cross-build to actually run the test. Removing the pin lets them run natively (~5s).

## Changes

- **`runtime-pi/test/provision-container.e2e.test.ts`**
  - Drop the `--platform linux/amd64` pin from `docker run`.
  - Replace `hasImage()` with a platform-aware gate: compare the local image's `{{.Os}}/{{.Architecture}}` against the daemon's `{{.Server.Os}}/{{.Server.Arch}}` (both emit Go's GOOS/GOARCH vocabulary, so they are directly comparable). A stale cross-arch image now skips honestly, with a rebuild hint in the warning, instead of dying inside `docker run`. Output is validated against `^[a-z0-9]+/[a-z0-9]+$` and every failure mode (daemon down, image absent, unexpected output) degrades to a skip.
- **`.github/workflows/test.yml`**
  - Build the image without the platform pin (native on `ubuntu-latest`, identical result today; also correct if arm64 runners are ever used).

## Verification (arm64 host, Apple Silicon)

| Scenario | Before | After |
| --- | --- | --- |
| Stale amd64 image + arm64 daemon | exit 125, `pull access denied` | honest skip: `image(appstrate-pi:latest)=linux/amd64 daemon=linux/arm64 — rebuild natively: docker build -t appstrate-pi:latest -f runtime-pi/Dockerfile .` |
| Natively built arm64 image | required slow QEMU cross-build | **1 pass** in ~5s |

- Full `runtime-pi/` suite with `TEST_DOCKER=1`: **779 pass, 0 fail** (56 files).
- `tsc --noEmit`, ESLint, Prettier: clean.
- CI (`ubuntu-latest`, amd64): behavior unchanged — native build matches native daemon, e2e runs as before.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -167,11 +167,11 @@ jobs:
       - name: Install dependencies
         run: bun install
 
-      # Build the agent runtime image the e2e launches. linux/amd64 is native on
-      # ubuntu-latest, so no QEMU. Tag `appstrate-pi:latest` — the test's default
-      # `PI_IMAGE`, and the trigger for its `hasImage()` gate.
+      # Build the agent runtime image the e2e launches, on the runner's native
+      # platform — the test gates on the image matching the engine's platform
+      # (#882). Tag `appstrate-pi:latest` — the test's default `PI_IMAGE`.
       - name: Build appstrate-pi image
-        run: docker build --platform linux/amd64 -t appstrate-pi:latest -f runtime-pi/Dockerfile .
+        run: docker build -t appstrate-pi:latest -f runtime-pi/Dockerfile .
 
       # The root bunfig preload boots the test docker-compose stack
       # unconditionally; start it so the preload is satisfied (the e2e itself
--- a/runtime-pi/test/provision-container.e2e.test.ts
+++ b/runtime-pi/test/provision-container.e2e.test.ts
@@ -83,7 +83,7 @@ const RUN = daemon !== null && image === daemon;
 if (dockerEnabled && !RUN) {
   const hint =
     image !== null && daemon !== null && image !== daemon
-      ? ` — rebuild natively: docker build -t ${IMAGE} -f runtime-pi/Dockerfile .`
+      ? ` — rebuild natively: docker build --platform ${daemon} -t ${IMAGE} -f runtime-pi/Dockerfile .`
       : "";
   console.warn(
     `[provision-container.e2e] skipped — docker=${hasDocker()} image(${IMAGE})=${image ?? "absent"} daemon=${daemon ?? "unknown"}${hint}`,
@@ -180,9 +180,13 @@ describe.skipIf(!RUN)("runtime-pi container provisions documents without spinnin
           "-d",
           "--name",
           containerName,
-          // No --platform pin: the gate above guarantees the local image
-          // matches the engine's native platform, mirroring how the platform's
-          // Docker orchestrator launches this image in production.
+          // Pin to the engine's native platform (which the gate above
+          // guarantees the local image matches, mirroring how the platform's
+          // Docker orchestrator launches this image in production). Explicit
+          // rather than omitted: a DOCKER_DEFAULT_PLATFORM env override would
+          // otherwise re-route the run to a foreign platform behind the
+          // gate's back (#882). `daemon` is non-null whenever RUN is true.
+          ...(daemon !== null ? ["--platform", daemon] : []),
           // Linux portability — Docker Desktop adds this automatically, but CI
           // engines need it explicit.
           "--add-host",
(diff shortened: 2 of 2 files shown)


pierrecabriere:
Addressed a review finding in 752352ef0: the gate compared image↔daemon platforms but `docker run` omitted `--platform`, so a `DOCKER_DEFAULT_PLATFORM` env override could still re-route the run to a foreign platform behind the gate's back (reproduced: `DOCKER_DEFAULT_PLATFORM=linux/amd64` on an arm64 daemon+image → same exit-125 `pull access denied`). The run now pins `--platform ${daemon}` explicitly, and the rebuild hint includes `--platform ${daemon}` too (a bare `docker build` under the same override would rebuild the wrong platform again).

Verified on arm64: passes both with and without `DOCKER_DEFAULT_PLATFORM=linux/amd64`.

PR #888, opened 2026-07-10
fix(security): close 17 authorization, tenant, credential and SSRF boundaries

pierrecabriere:
Remediates the 2026-07 prioritized security review. Every claim was re-verified against the code before being fixed — the report's own verdicts were not taken on trust, and four of them needed correction.

> **Update (round 2)** — a reviewer rejected `a2e71025` with five blockers. All five were real; all five are fixed. Details in the "Round 2" section below.

## Verification outcome

The report claimed 18 critical causes, "17 confirmed". Independent re-verification found **14 correct as stated, 3 overstated, 1 false in its mechanism**. All are addressed, each at its true scope.

| Report claim | Verdict | Fixed as |
|---|---|---|
| CRIT-11 private MCP resolved cross-tenant by name collision | **False mechanism** | `packages.id` is the PK and equals the fully-qualified `@scope/name`, so a cross-org name collision is impossible. The real gap — `resolvePublishedManifest` resolving with no `orgId` filter — is now scoped. |
| CRIT-05 `PUT packages` exposes a foreign manifest | **Overstated** | `getPackage`'s `orgOrSystemFilter` already blocks the run and read paths. The genuine write-side gap (blind upsert, existence oracle) is closed. |
| CRIT-12 any late event resurrects a terminal run | **Overstated** | `assertSinkOpen` already 410s upstream. The residual TOCTOU and the non-monotone status transition are closed in the CAS. |
| CRIT-14 actor-less history reads other actors' results | **Overstated** | Unreachable today (`getActor` never returns null; a check constraint exists). The missing predicate is added as defense in depth. |
| CRIT-17 SMTP SSRF | **True, re-prioritized** | Real, but reachable only by an authenticated org admin acting on their own app — not unauthenticated. |
| The other 13 | **Confirmed** | See below. |

CRIT-02 is the most severe and was moved to the front of the remediation order: the auth pipeline computes an API key's scopes correctly, but the org-administration routes gated on `requireOrgRole`, which reads the **creator's live DB role** and never consults them. A key scoped to `runs:read` could `DELETE /api/orgs/:orgId`.

## Round 2 — the reviewer's five blockers

**[P0] The LLM proxy still had the check-then-fetch SSRF.** `llm-proxy/core.ts` validated DNS, discarded the pin, then called `fetchImpl(upstreamUrl)` — the host was re-resolved at connect time and redirects were followed. An org-configurable `openai-compatible` provider could rebind or redirect into the internal network, carrying the provider API key. CRIT-16 was closed on one inference path, not both. Now routed through `egressGuardedFetch` with `maxRedirects: 0`.

Auditing for the same pattern found **three more unguarded callers**, all with org-controllable hosts, none of them in the original report:
- `org-models.ts::testModelConfig` — same class, and it carries the provider API key.
- `connect/dcr.ts::registerDynamicClient` — the endpoint comes from an org-configured remote MCP server's discovery metadata; only a literal IP check guarded it. (This is the report's P1-11.)
- the sidecar's remote-MCP fetch — injects a manifest-named credential header with no `sensitiveHeaders`.

All three are fixed. `org-proxies.ts::testProxyConnection` is deliberately **not** fixed: `guardedFetch` cannot pin Bun's `proxy:` option, and pinning an https proxy's host would break its TLS identity. It carries no platform credential and targets a fixed host. Documented as an accepted residual rather than given a fake pin.

**[P0] Redirects could still exfiltrate a credential from the credential proxy.** The `authorized_uris` allowlist was enforced only on the initial URL, and cross-origin header stripping was a hardcoded three-name list — but the proxy injects arbitrary vendor headers (`X-Api-Key`, …), which followed a redirect to any host. A same-origin redirect could also walk off the authorized path with headers and body intact.

`guardedFetch` gains `validateHop` (called on **every** hop including hop 0; throwing aborts the exchange) and `sensitiveHeaders` (unioned with the builtin set, never replacing it). The credential proxy passes its existing `authorized_uris` matcher as `validateHop` — reusing the same function so the two cannot drift — and the header names it actually injected, collected at injection time rather than guessed. An off-allowlist hop now **aborts** instead of strip-and-continue. `maxRedirects: 0` was considered and rejected: legitimate vendor APIs 302 inside their own allowlisted host, and `validateHop` gives the property that actually matters — a request can never be sent to a URL outside `authorized_uris`, on any hop.

**[P1] CRIT-15 was only closed for email/password signup.** The social-callback and magic-link verify legs are driven by Better Auth itself and still read the ambient global cookie. The realm resolver now derives the pending client from the OAuth transaction: Better Auth publishes its single-use state via `getOAuthState()` on the social callback (`dist/api/state/oauth.d.mts`, pinned 1.7.0-beta.4), and the magic-link token is bound at issuance. Both are server-side and unforgeable by the browser. An OIDC-initiated flow whose binding is missing or inconsistent now **fails closed**. No new table — the binding reuses Better Auth's `verification` table, so it works on Tier 0 and across instances. Both legs are closed; two-tab flows no longer interfere in either direction.

**[P1] The `check` job was genuinely broken.** `RUN_TOKEN_SECRET` became required but was missing from `globalPassThroughEnv`, so Turbo stripped it from every subprocess and `detect:breaking` died on `expected string, received undefined`. Added to `turbo.json`. Reproduced locally with no `.env` present and confirmed fixed.

**[P1] CRIT-07 was incomplete for historical data.** `run-metric-broadcaster.ts` still summed `llm_usage` by `run_id` alone on the mid-run broadcast path — now scoped by `(run_id, org_id)`. Migration `0021` detaches poisoned legacy rows with `run_id = NULL` and then **validates** the composite FK. The detach is the correct repair: the poisoned row's `org_id` is truthful (it came from the authenticated principal); only the run attribution was forged. Deleting would erase real spend; rewriting `org_id` would move a victim org's ledger.

**Regression coverage** — the reviewer was right that the new contracts had none, which is why `codecov/patch` was red. 43 new tests, each written to fail if its fix is reverted.

## Round 1 — what changed

**Authorization** — org-administration routes reject API keys outright (`org:*`/`members:*` are not in `API_KEY_ALLOWED_SCOPES`, so no key can legitimately hold them). SSE requires `runs:read` and derives `isAdmin` from the resolved role instead of hardcoding `"admin"`. `/me/connections` is scoped to the key's org+application when the caller is an API key; the cross-org view stays an interactive-session feature, and the scope is a required service parameter that lands in the SQL `WHERE`.

**Tenant isolation** — `llm-proxy` validates `X-Run-Id` against the principal's org and application before the upstream call; `computeRunCost` filters `llm_usage` on `(run_id, org_id)`; a composite FK makes the invariant structural. Bundle import claims and validates ownership in one advisory-locked transaction, so the loser of a concurrent same-id import 409s instead of grafting a version and storage bytes onto the winner's foreign-org package. `updateInstalledPackage` requires a pre-existing association and checks visibility atomically with the write.

**Credentials** — `/internal/oauth-token` fails closed on a null pin. `loadAccessibleConnectionById` binds `integrationId` and `authKey` in the `WHERE` and re-asserts before decryption. The credential proxy separates the real target from a redacted rendering. OAuth model providers are classified by `authMode` first — a system model lost its `credentialId`, downgrading an OAuth provider to API-key handling and putting the raw subscription bearer in `MODEL_API_KEY` inside the agent container.

**Egress** — `guardedFetch` really pins the connection to the validated address while preserving `Host` and TLS SNI (verified empirically on Bun 1.3.11: a wrong `serverName` fails certificate validation, so the pin does not bypass TLS). SMTP resolves fail-closed and connects to the pinned address at connect time.

**State machine** — the ingestion CAS includes `sink_closed_at IS NULL`, and `updateRun` enforces monotone status in the `WHERE`. A removed member's schedules are disabled transactionally with the membership delete, and every fire revalidates the frozen actor.

## Breaking changes — read before deploying

1. **`RUN_TOKEN_SECRET` is now required** (≥16 chars per keyring entry). When unset, the keyring degraded to an unkeyed SHA-256 of the runId — anyone knowing a runId could forge its bearer. **This is the one item that needs an ops step before rollout.**
2. **Boot crashes** if `SYSTEM_PROVIDER_KEYS` binds a provider whose `authMode` is `oauth2`.
3. API keys receive **403 on org-administration routes** and on SSE without `runs:read`.
4. `/internal/oauth-token` returns **403 for any run without a pinned credential**.
5. `PUT /applications/:id/packages/:pkgId` on a not-installed package now **404s**.
6. Removing an org member **disables their schedules** in that org.
7. Event POSTs racing finalize now get **410** (`run_sink_closed`).
8. Deployments egressing through an HTTP(S) proxy with **hostname-based ACLs** will now CONNECT to IPs. Per-call opt-out: `pinToResolvedAddress: false`.

## Migrations

`0020` adds `uq_runs_id_org_id` (cannot fail — `id` alone is the PK) and the composite FK **`NOT VALID`**, so existing rows are never scanned at apply time. `0021` detaches poisoned rows and **validates** it. `0022` deletes historical stray cross-org `application_packages` rows (creatable by the pre-fix upsert PUT). (Renumbered from 0019/0020 after merging main, where #911 took the 0019 slot; the boot migrator applies pending migrations in one transaction, so on a large `llm_usage` apply 0020 and 0021 in separate releases — see the comment in 0021.) No `DROP CONSTRAINT` — the prod `_fkey`/`_fk` name-drift trap does not apply.

Drizzle's generated migration ordered the FK *before* its referenced unique index (it would have failed to apply) and smuggled in an unrelated pre-existing `chat_*` timestamptz drift. The SQL is hand-ordered, and the snapshot was de-drifted so that drift stays visible to whoever owns `schema/chat.ts`.

## Verification

- `bun run check` — 33/33, **run with `--force`** after CI caught a typecheck error that Turbo's cache had masked locally
- `bun test` — **8217 pass, 0 fail**, 631 files
- All 23 migrations replayed against a fresh PGlite database; the composite FK ends `convalidated = true`
- `detect:breaking` passing with no `.env` present, reproducing the CI environment

## Related issues (tracking — deliberately NOT closed by this PR)

None of these should auto-close at merge; they are the follow-up trackers spun out of this remediation:

- #891 / #892 / #893 — the 16 P1 findings of the audit, kept out of this PR (public-contract breaks, Firecracker e2e harness required, decoupled rollback)
- #894 — AAD credential-envelope v2 (discovered during CRIT-10 remediation, separate project)
- #895 — accepted residuals + audit coverage gaps (must survive the merge by definition)
- #897 — LOW defensive hardenings found during merge-readiness review

## Known follow-ups (deliberately out of scope)

- Credential decryption is still not cryptographically bound to the integration (no AAD). Needs a versioned envelope in `@appstrate/connect`.
- `GET /api/orgs/:orgId` still returns raw invitation tokens to any same-org API key (report P1-08).
- `org-proxies.ts::testProxyConnection` — accepted residual, documented in code.
- If an end user is hard-deleted, a frozen schedule job firing afterwards cannot record a visible failed run (dangling FK); that sub-case still logs and skips.
- The remaining 15 P1 findings of the original report are untouched.

> **Update (round 3)** — merged `main` (migrations renumbered 0020/0021 after #911 took 0019), and closed a reviewer-found CRIT-05 residual: `listInstalledPackages` missed `orgOrSystemFilter`, so a historical stray association row could leak a foreign org's `draft_manifest` through the package list. Filter added + regression test + migration `0022` prunes the stray rows. The `pg_constraint` guards in 0020/0021 now match on `(conname, conrelid, contype)` — `conname` alone is not unique in Postgres.


pierrecabriere:
Les cinq bloqueurs étaient réels — vérifiés un par un contre le code avant correction. Tout est fixé sur `ec89b8811`.

**[P0] SSRF check-then-fetch du LLM proxy.** Corrigé : `llm-proxy/core.ts` passe par `egressGuardedFetch` avec `maxRedirects: 0`. En auditant le même motif, j'ai trouvé **trois autres appelants non gardés** que le rapport initial ne mentionnait pas, tous avec un host contrôlable par l'organisation : `org-models.ts::testModelConfig` (qui transporte la clé API du provider), `connect/dcr.ts::registerDynamicClient` (endpoint issu des métadonnées de découverte d'un MCP distant, protégé par un simple contrôle d'IP littérale — c'est le P1-11 du rapport), et le fetch remote-MCP du sidecar. Les trois sont corrigés. `org-proxies.ts::testProxyConnection` est **laissé tel quel** : `guardedFetch` ne peut pas épingler l'option `proxy:` de Bun, et épingler l'host d'un proxy https casserait son identité TLS. Documenté comme résidu accepté plutôt que doté d'un faux pin.

**[P0] Exfiltration de credential par redirect.** `guardedFetch` gagne `validateHop` (appelé à **chaque** hop, hop 0 inclus ; lever une exception avorte l'échange) et `sensitiveHeaders` (union avec la liste intégrée, jamais un remplacement). Le credential proxy passe son propre matcher `authorized_uris` comme `validateHop` — la même fonction, pour qu'elles ne puissent pas diverger — et les noms des headers qu'il a réellement injectés, collectés au moment de l'injection. Un hop hors allowlist **avorte** désormais, au lieu de strip-and-continue. J'ai écarté `maxRedirects: 0` : des APIs vendeur 302 légitimement à l'intérieur de leur propre host autorisé, et `validateHop` donne la propriété qui compte — aucune requête ne peut partir vers une URL hors `authorized_uris`, sur aucun hop.

**[P1] CRIT-15.** Fermé sur les trois legs. Le resolver dérive maintenant le client depuis la transaction OAuth : Better Auth publie son `state` à usage unique via `getOAuthState()` sur le callback social (`dist/api/state/oauth.d.mts`, 1.7.0-beta.4), et le token magic-link est lié à l'émission. Les deux sont côté serveur, non forgeables par le navigateur. Un flux OIDC dont le binding manque ou diverge **échoue fermé**. Pas de nouvelle table : le binding réutilise la table `verification` de Better Auth, donc ça marche en Tier 0 et multi-instance.

**[P1] CI.** Diagnostic exact. `RUN_TOKEN_SECRET` manquait dans `globalPassThroughEnv` de `turbo.json`. Reproduit en local en déplaçant `.env` et en n'exportant les secrets que dans le shell. Au passage, une seconde leçon : le cache Turbo masquait une erreur de typecheck sur un fichier de test — `bun run check` vert en local, rouge en CI. Je termine désormais par `bun run check -- --force`.

**[P1] CRIT-07 historique.** `run-metric-broadcaster.ts` est scopé sur `(run_id, org_id)`. Migration `0020` : détache les lignes empoisonnées avec `run_id = NULL`, puis `VALIDATE CONSTRAINT`. Le détachement est la bonne réparation — l'`org_id` de la ligne est véridique (il vient du principal authentifié), seule l'attribution au run était forgée. Supprimer effacerait une dépense réelle ; réécrire `org_id` déplacerait le ledger d'une victime.

**Couverture.** 43 régressions, chacune écrite pour échouer si son correctif est annulé. Le `codecov/patch` rouge n'était pas un manque de tests mais une conséquence de la config du repo : seul le job `unit` alimente Codecov, et `.codecov.yml` dit explicitement que les PR dont le code est exercé par la suite d'intégration doivent porter le label `integration`. Label ajouté → job d'intégration exécuté → `codecov/patch` passe à **59.74% du diff** (cible 50%).

État : `check`, `Integration tests`, `Unit tests`, `smoke`, `CodeQL`, `Analyze`, `dnat-regression`, `Secret Scanning`, `codecov/patch` — tous verts. `mergeStateStatus: CLEAN`.

Deux étapes ops restent obligatoires avant le déploiement, détaillées dans la description : `RUN_TOKEN_SECRET` devient requis, et le boot crashe si `SYSTEM_PROVIDER_KEYS` contient un provider `oauth2`.

pierrecabriere:
## Revue de merge-readiness + merge de `main`

Relu la PR en détail (deux reviewers adversariaux sur des périmètres disjoints, plus vérification manuelle des points remontés). `main` a avancé de 3 commits pendant ce temps — mergé dans la branche, aucun conflit, tout re-vérifié sur l'arbre fusionné.

### Verdict : mergeable. Deux corrections apportées, le reste tracé en issues.

**Corrigé dans la PR** (findings dans nos propres fichiers) :
- **Migrations 0019/0020 non idempotentes** → auraient pu crash-looper le boot sur un rejeu. Vu l'historique de réparations manuelles du watermark `__drizzle_migrations` de ce repo, c'était un vrai risque. `CREATE UNIQUE INDEX IF NOT EXISTS` + garde `pg_constraint` sur l'`ADD CONSTRAINT` + `VALIDATE` gardé. 0020 pré-vérifie les lignes `source='runner'` org-mismatchées et **échoue avec un message actionnable** au lieu d'une violation de CHECK opaque. Prouvé sur PGlite : apply à froid (FK `convalidated=true`), rejeu complet de 0019+0020 sans crash, et une ligne runner corrompue plantée → 0020 échoue avec le message nommant la requête de diagnostic exacte.

**Réfuté en revue** (pas des bugs) :
- Le run OAuth aliasé qui aurait donné un 403 sur `/internal/oauth-token` : `pi.ts:184` (`assertOauthRunNotAliased`) rejette ce run **avant** tout provisioning, il n'atteint jamais l'endpoint.
- La divergence de normalisation hop 0 du credential proxy : les 95 patterns `authorized_uris` réels des packages système passent identiquement en brut et normalisé — fail-closed sans impact, tracé S1.

**Tracé en issues** (hors périmètre, découplé du rollback des P0) :
- #891 — concurrence OAuth, idempotency, frontières de permissions (P1-01/02/03/05/08/09/10 ; cassent des contrats publics, demandent un plan de dépréciation)
- #892 — isolation Firecracker/Docker (P1-12/13/14 ; non vérifiables par la CI de cette PR, nécessitent le harness Lima)
- #893 — durcissement léger + CI (P1-04/06/07/15/16)
- #894 — AAD `(integrationId, authKey)` sur le chiffrement des credentials (chantier envelope v2)
- #895 — résidus acceptés + lacunes de couverture de l'audit (36,25 % de la codebase)
- #897 — durcissements défensifs LOW relevés en revue (S1, S2, O1, R1, F1, F2, F3, F5)

### Rappel des préconditions ops avant déploiement (inchangées)
1. Poser `RUN_TOKEN_SECRET` (≥16 caractères) sur les instances existantes — sinon crash au boot.
2. Vérifier qu'aucun `SYSTEM_PROVIDER_KEYS` ne référence un provider `oauth2` — sinon crash au boot (volontaire).
3. **Réparer/vérifier le watermark `__drizzle_migrations` en prod** avant de shipper, sinon 0019+0020 sont silencieusement skippées. Vérifier après deploy que `llm_usage_run_id_org_id_fk` existe avec `convalidated=true`.

Issue #881, opened 2026-07-09
@appstrate/google-drive: api_upload never exposed — upload_protocols misplaced in manifest, forcing brittle hand-built multipart via api_call


otarbes:
## Summary

The `@appstrate/google-drive` system integration never exposes the `api_upload` tool, even though the platform supports it, because the manifest declares `upload_protocols` in the wrong place. As a result, any agent that needs to **write a file** (create a Doc, upload content) is forced to hand-build a `multipart/related` body and POST it through `api_call` — brittle, base64-prone, and it returns a giant response that spills to the blob store and derails the agent. Tested end-to-end on `beta.38`.

## Root cause

`api_upload` is exposed only when the auth declares `upload_protocols` at `auths.<key>.upload_protocols`. In `@appstrate/google-drive` it's nested one level too deep, under `_meta`:

```jsonc
// current (api_upload NOT exposed)
"_meta": {
  "dev.appstrate/api": {
    "auths": { "primary": { "upload_protocols": ["google-resumable"] } }
  }
},
"auths": {
  "primary": { "type": "oauth2", /* no upload_protocols here */ }
}
```

Runtime confirmation on `beta.38`:
- `GET /api/integrations/@appstrate/google-drive` → `tool_catalog: [{ "name": "api_call" }]` (no `api_upload`).
- Selecting `api_upload` in an agent's `integrations_configuration` → import rejected: `unknown_tool: Tool "api_upload" is not exposed by integration @appstrate/google-drive`.

Expected: `upload_protocols` at `auths.primary.upload_protocols` so the resolver surfaces `api_upload`.

## Impact

With only `api_call` available, creating a Google Doc from Markdown requires the agent to emit a raw `multipart/related` body inline (metadata part + media part). Observed consequences:
- Agents frequently never complete the upload (the model stalls building the large body; on large upstream context the 120s no-activity watchdog then kills the run **before** the Drive step is even reached).
- When the upload does happen, the Drive response measured ~40k tokens and **spilled to the blob store**, so the agent couldn't read back `webViewLink` and returned `document_url: null` even though the file was created.
- Only after rewriting the prompt to (a) request `?fields=id,webViewLink` (tiny response) and (b) forbid base64 and pin an explicit `boundary` did it become reliable.

`api_upload` would remove all of this — it's the tool designed for exactly this (chunked/resumable, agent-side, handles the body/encoding).

## Fix

[...]

PR #899, opened 2026-07-13
fix(runtime-pi): api_upload conversion — derive X-Upload-Content-Type from source MIME, not target

otarbes:
## Problem

`api_upload` over `google-resumable` cannot convert an uploaded file to a Google-native type (Markdown/HTML/CSV → Google Doc/Sheet). The resumable init succeeds but Drive rejects the **first chunk** with a `400 Bad Request`, so no file is created.

Follow-up to #881 (which exposed `api_upload` in the tool catalog). This is a **separate** bug — the conversion path — that surfaced once the tool became usable.

## Root cause

`google-resumable.ts` sets `X-Upload-Content-Type` — which declares the MIME of the **bytes being uploaded** (the source) — from `metadata.mimeType`. But for a conversion, `metadata.mimeType` is the **target** type (e.g. `application/vnd.google-apps.document`). Declaring a Google-native type as the source is nonsensical (you can't upload native-type bytes), so Drive accepts the init but rejects the chunk.

## Validation (live Drive API)

Full resumable flow (init + chunk) against the real Drive API, varying only `X-Upload-Content-Type`:

| `X-Upload-Content-Type` | body `mimeType` | init | chunk | result |
|---|---|---|---|---|
| `application/vnd.google-apps.document` (target) | `…document` | 200 | **400** | fails (current behavior) |
| `text/markdown` (source) | `…document` | 200 | **200** | ✅ converted Google Doc created |

## Fix

Introduce an explicit **`sourceMimeType`** on `api_upload`, threaded through to the adapter:

- `sidecar/mcp.ts` — authoritative `api_upload` input schema (what the agent sees) + a description spelling out source-vs-target
- `api-upload-extension.ts`, `api-upload-resolver.ts`, `upload-adapters/types.ts` — plumb `sourceMimeType` from tool args → `IntegrationUploadRequest` → `AdapterContext`
- `google-resumable.ts` — `X-Upload-Content-Type` now prefers `ctx.sourceMimeType`; falls back to `metadata.mimeType` **only when it is not a `application/vnd.google-apps.*` native type** (so the 400 can no longer happen and plain uploads are unchanged)

Usage for conversion becomes:

```jsonc
{
  "target": "https://www.googleapis.com/upload/drive/v3/files?uploadType=resumable",
  "fromFile": "report.md",
  "uploadProtocol": "google-resumable",
  "sourceMimeType": "text/markdown",
  "metadata": { "name": "Report", "parents": ["…"], "mimeType": "application/vnd.google-apps.document" }
}
```

## Backward compatibility

- No `sourceMimeType`, `metadata.mimeType` a real (non-native) type → unchanged (still used as the source declaration).
- No `sourceMimeType`, `metadata.mimeType` a Google-native type → header now **omitted** instead of sent (previously guaranteed a 400).
- Plain uploads (no `mimeType`) → unchanged.

## Tests

- New `runtime-pi/test/google-resumable-source-mime.test.ts` — 4 cases (explicit source; native-target skipped; non-native fallback; none). **4/4 pass.**
- Existing `api-upload-extension.test.ts` (5/5) and `mcp-api-upload-resolver.test.ts` (20/20) green.
- `runtime-pi` typecheck clean.

Based on latest `main` (`af4fd939`), which is ahead of `v1.0.0-beta.39`.

cc @pierrecabriere

### DIFF
--- a/runtime-pi/mcp/api-upload-extension.ts
+++ b/runtime-pi/mcp/api-upload-extension.ts
@@ -111,6 +111,7 @@ function makeExtension(
             fromFile: { type: "string" },
             uploadProtocol: { type: "string", enum: protocols },
             metadata: { type: "object", additionalProperties: true },
+            sourceMimeType: { type: "string" },
             partSizeBytes: { type: "integer", minimum: 1 },
           },
         },
@@ -121,6 +122,7 @@ function makeExtension(
           fromFile?: string;
           uploadProtocol?: string;
           metadata?: Record<string, unknown>;
+          sourceMimeType?: string;
           partSizeBytes?: number;
         };
 
@@ -171,6 +173,7 @@ function makeExtension(
             fromFile: args.fromFile,
             uploadProtocol: protocol as UploadProtocol,
             metadata: args.metadata,
+            ...(args.sourceMimeType !== undefined ? { sourceMimeType: args.sourceMimeType } : {}),
             ...(args.partSizeBytes !== undefined ? { partSizeBytes: args.partSizeBytes } : {}),
           },
           {
--- a/runtime-pi/mcp/api-upload-resolver.ts
+++ b/runtime-pi/mcp/api-upload-resolver.ts
@@ -89,6 +89,9 @@ export interface IntegrationUploadRequest {
   fromFile: string;
   uploadProtocol: UploadProtocol;
   metadata?: Record<string, unknown>;
+  /** MIME type of the source bytes (→ `X-Upload-Content-Type`). Distinct
+   *  from `metadata.mimeType`, which is the desired/target file type. */
+  sourceMimeType?: string;
   partSizeBytes?: number;
 }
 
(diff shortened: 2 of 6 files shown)


pierrecabriere (approved):
Solid root-cause fix — `X-Upload-Content-Type` semantics confirmed against Google's docs (header declares the SOURCE bytes' MIME, `metadata.mimeType` is the conversion target), and the live-API validation table settles it. The explicit `sourceMimeType` param is the right call over the minimal "omit native types" fix: relying on Drive sniffing would degrade Markdown conversions to plain-text imports.

Two non-blocking notes for potential follow-ups:
- `sourceMimeType` is silently ignored by the other adapters (tus, MS Graph, S3 multipart) — an agent passing it there gets no effect and no warning. A debug log or doc note in the adapter contract could help.
- The plumbing extension → resolver → adapter for the new field isn't covered by a dedicated test (the new suite exercises the adapter directly). Existing resolver tests stay green, so low risk.

Merging.

PR #922, opened 2026-07-19
feat(runtime): add Browser Use marketplace capability

pierrecabriere:
## Summary

- model Chromium as an MCP-server execution capability instead of a language runtime
- add first-party Browser Use drivers for Leboncoin and Vinted, including authenticated session acquisition and restoration
- add a hybrid local-companion flow: login happens in a dedicated Chrome on the user's machine, then a bounded portable session is re-proved in the operator-selected cloud or in-process backend
- allocate one provider profile per integration connection, with stable encrypted proxy routing and no cross-tenant browser-state sharing
- add trusted browser connection acquisition for link and run-start flows, with exact system-package grants and authenticated proof before persistence
- add a first-party, versioned browser worker and matching process, Docker, Firecracker, and Browser Use Cloud providers
- add a scoped blind CONNECT gateway that preserves Chromium TLS, pins public DNS results, enforces exact origins, chains through the selected upstream proxy, and never falls back to direct egress
- propagate browser requirements before isolation-boundary creation and account for worker resources in Docker and Firecracker

## Hybrid local companion

- the hosted Connect page mints a 256-bit, short-lived companion capability; only its SHA-256 digest is stored, and companion endpoints are `no-store`
- the deep link contains only an exact attempt endpoint and bearer; the macOS shim passes it to the compiled worker over stdin, not argv
- the companion validates every server-controlled URL before opening Chrome or the OS: HTTPS in production, loopback HTTP only for local development, exact endpoint path, exact HTTPS origins, and no wildcard hostnames
- login runs in a dedicated temporary Chrome profile on the user's residential IP and remains human-controlled
- only cookies and localStorage for manifest-authorized exact domains are captured; state shape, entry counts, field sizes, duplicates, and the 900 KiB total envelope are validated again by the API
- the handoff and provider live-view URL are encrypted at rest; proxy credentials are encrypted and never returned to the browser client
- the provider is operator-controlled. A hosted client cannot switch from `process` to `browser-use-cloud` or consume a different routing policy
- the target backend restores the state and must return an authenticated proof before the integration connection and provider binding become ready
- Browser Use Cloud receives a connection-scoped profile UUID and connection-scoped proxy snapshot; process mode receives the same portable state but no cloud routing material

## Durable state, concurrency, and cleanup

- companion attempts use a durable state machine: `pending → claimed → state_received → provisioning → interaction_required/complete`, with explicit failed/expired terminal states
- API restart recovery reclaims stale provisioning work, resumes queued handoffs, expires abandoned attempts, and purges terminal metadata after bounded retention
- browser profile use is serialized for the complete agent run through connection-owned leases, monotonic state versions, and fencing tokens
- reconnect cannot replace a provider profile while a run holds its lease; stale run plans fail with `BROWSER_STATE_CONFLICT`
- deleting a connection remains immediate, but remote profile deletion waits for an active run lease to expire
- provider deletion uses a durable idempotent outbox with compare-and-swap claims and bounded exponential retry
- PostgreSQL cascade triggers enqueue cleanup for profiles deleted through application/user/connection cascades, while completed attempt metadata never schedules deletion of its live bound profile

### Leboncoin

- local-companion login uses the user's real browser and network path, then re-proves the exported session in the selected target backend
- restored state is accepted only after the browser reaches the account area with the exact-domain login cookie; cookie presence alone is not proof
- read-only `search_listings`, `get_listing`, and `session_status` tools
- DataDome/account verification is surfaced as interaction-required while unrelated tools remain available; no unconditional DataDome bypass is claimed

### Vinted

- local-companion login with the same encrypted exportable-session contract
- `search_items`, `get_item`, and `browser_status` tools
- two-step publication: `prepare_item_draft` fills a draft without submitting; `publish_item` requires an exact one-time token and `confirm_publish: true`
- 1–20 workspace-relative JPEG/PNG/WebP images, bounded to 20 MiB each and 100 MiB total, with traversal and symlink escape rejected
- the draft token is consumed before the publish click, so a failed or ambiguous submission cannot be replayed

## Browser Use runtime and cloud provider

- dedicated pinned Python 3.12 Browser Use runtime and bridge; Chromium remains in the isolated browser worker
- strict authenticated CDP and context APIs, exact HTTPS origins, bounded state, structured JavaScript arguments, safe error canonicalization, and multi-file upload
- cloud CDP is terminated behind a per-run local bearer broker; the runner never receives the provider API key or raw remote CDP URL
- cloud CDP endpoints are host-validated, message/inflight/page-bounded, and deny nested tunnels, browser lifecycle commands, foreign contexts, and off-origin navigation
- cloud mode is restricted to trusted system connection-acquisition drivers because its subresources do not traverse Appstrate's local egress gateway
- Vinted publication requires the local Docker, process, or Firecracker provider because remote Chromium cannot address Appstrate's local upload paths

## Session semantics

- link-time marketplace sessions keep the `exportable` contract: the persisted credential is a portable cookie/localStorage snapshot
- Browser Use Cloud may additionally retain continuity in the connection-owned provider profile, but the target driver still receives and re-proves the portable state
- snapshots are encrypted at rest and delivered only over the private sidecar-to-driver channel; they are never placed in env, argv, or model-visible tool output
- true `browser-bound` link persistence remains fail-closed outside this flow
- process-mode cookie rotations made during a run are not currently written back; an expired session requires reconnecting the integration

## Security and isolation

- browser capability and secret-aware acquisition are disabled by default behind separate feature gates
- secret-aware drivers require an exact package/version grant, optional exact-origin ceiling, and system-package resolution provenance; org-owned packages cannot receive bootstrap secrets
- connect-login and browser-connect tools are always removed from the agent allowlist and added to the sidecar hidden-tools set, including wildcard selections
- the agent never receives the browser endpoint, bearer token, bootstrap inputs, cookies, provider profile UUID, proxy credentials, or exported session values
- Docker workers are non-root, read-only, `cap-drop=ALL`, `no-new-privileges`, seccomp-constrained, and resource-bounded
- Firecracker assigns distinct driver/browser UIDs and fixed loopback ports; the browser UID is denied general loopback egress, including sidecar ports 8080/8081
- Firecracker runner ingress and the orchestrator recompute platform-owned browser resources and reject client-inflated requests

## Validation

- `TMPDIR=/private/tmp bun run check`: **35/35 tasks pass**; OpenAPI valid, generated API types current, type coverage **99.01%**, no breaking API changes
- companion/API/profile/lease/outbox/cascade suite: **16 pass**
- Sidecar provider selection, Browser Use Cloud binding/proxy/broker, process and Docker provider suite: **26 pass**
- companion protocol/CDP bounds: **7 pass**; real local Chrome CDP extraction smoke: **1 pass**
- Drizzle journal and snapshot: `drizzle-kit check` passes; PostgreSQL migration and cascade triggers exercised by integration tests
- system-package validation: **70/70**; MCP parity: GitHub **7**, Leboncoin **4**, Vinted **6** tools
- Browser Use runner, browser worker, and final Sidecar Docker images build successfully
- final Sidecar image: `sha256:0406ec5e2dfc24b7e41da3d5c4b3eae2996ca4d4ed0bec33ed29c2a66f6931e6`
- macOS companion app builds, its plist validates, and deep ad-hoc signature verification passes
- live Vinted smoke loaded the real catalog with **221** interactive elements, restored storage, and no DataDome challenge
- live Leboncoin smoke reached the real route and restored storage, but the automated target worker still received a DataDome challenge

## Rollout and limits

- production remains opt-in through `BROWSER_ENABLED`, `BROWSER_CONNECT_ENABLED`, `BROWSER_DRIVER_GRANTS`, `BROWSER_PROVIDER`, and provider configuration
- local companion improves first login by using the user's browser fingerprint and residential IP; it does not promise that a later cloud IP will be accepted
- for cloud continuity, operators should configure a stable trusted residential proxy per routing policy; country proxying or stealth alone is not a DataDome guarantee
- CAPTCHA, MFA, and account verification remain explicit human interaction gates
- process mode remains explicitly non-isolated/best-effort
- the macOS app produced here is suitable for local/dev validation and is ad-hoc signed; public production distribution still requires Developer ID signing, notarization, packaging, and an update channel

## Documentation and delivery

- environment variables, rollout gates, integration trust boundary, companion protocol, durable attempt lifecycle, leases, cleanup, provider selection, worker/gateway lifecycle, Docker/Firecracker provisioning, file-upload limits, and resource accounting are documented in code and package README files
- preview/release workflows publish the Browser Use runner and rebuild affected browser images when shared policy changes


pierrecabriere:
## Built-in Leboncoin integration added

Commit `2dd0d270d` adds:

- system integration `@appstrate/leboncoin@1.0.0` with run-start browser acquisition;
- granted system driver `@appstrate/leboncoin-browser@1.0.0`;
- private hidden `acquire_session` plus read-only `search_listings`, `get_listing`, and `session_status`;
- exact Leboncoin/DataDome origin allowlist, encrypted cookie-session export, and operator configuration docs;
- 7 focused unit tests and deterministic AFPS archives.

Local validation: full `bun run check` passed (34/34), conformance passed (4/4 tools), and the real Docker topology reached both Leboncoin search and auth. On the current direct IP, both returned the expected fail-closed `BROWSER_INTERACTION_REQUIRED` DataDome result; no anti-bot bypass was attempted.

PR #924, opened 2026-07-20
fix(module-chat): eliminate layout shifts in the chat thread

pierrecabriere:
## Problem

During inference, the copy button disappeared from **every** message, collapsing each assistant message by ~28px — the whole transcript visibly jumped at inference start and finish. A broader audit found the same class of bug (state transitions changing block heights) in several other places.

## Invariant introduced

The transcript only changes height when new content appends at the bottom. A mounted element never changes geometry from a state transition — visibility is opacity-only, variable-content slots have fixed heights, and a tool call renders ONE component for its whole life (no mid-stream component swaps).

## Changes

- **Action bar (the reported bug)** — dropped `hideWhenRunning` (it unmounted the bar on all messages while the thread ran) and reserved its space with a fixed `h-7` wrapper. Copy stays usable mid-stream.
- **Viewport** — removed `scroll-smooth`: it turned every streaming append into a visible glide and amplified residual shifts.
- **Thinking indicator** — pinned to `h-6` (exactly one `prose-sm` line) so the dots→first-text swap is a 0px change.
- **Tool cards** — fixed `h-9` header row: the HTTP badge (taller than the text line) and duration no longer grow the row when the result lands.
- **Run launches** (`runAgent` / `runInline` / `run_and_wait`) — the run progress panel is now mounted for the call's whole life. A launch failure renders inside the panel (error glyph + message on the fixed line 2) instead of swapping to the generic card. `shouldRenderRunLaunchPanel` removed.
- **Connect card** (`initiateIntegrationConnect`) — mounted from the first frame with a "Préparation…" state instead of swapping from a generic row when the auth url arrives. All states (preparing / idle / pending / error / connected) share the same two-row geometry; an initiate failure renders in place.
- **Sidebar** — fixed-width right slot (`w-14`) so spinner/dot/timestamp transitions no longer reflow truncated titles.

## Notes

- The connected state of the connect card is now the same two-row card as the offer (was a compact one-line chip) — deliberate: constant geometry wins over compactness.
- Known residual (documented, not fixed here): markdown streaming can still reflow when a code fence closes (paragraph re-renders as `<pre>`); and the run-launch panel appears only once `operation_id` has streamed in (first few tokens of args). Both are content-arrival transitions, not state flips.
- Follow-up candidate: a Playwright "layout invariants" test that streams a mocked turn and asserts message heights are monotonically non-decreasing.

## Validation

- `turbo check` green across all 17 affected tasks (module-chat + web + api typecheck, eslint, prettier).
- `bun test` in module-chat: 197 pass.
- `detect:breaking`: 0 breaking.

### DIFF
--- a/apps/web/src/hooks/use-global-run-sync.ts
+++ b/apps/web/src/hooks/use-global-run-sync.ts
@@ -234,6 +234,11 @@ export function useGlobalRunSync() {
         throw new Error(`realtime stream unavailable (${res.status})`);
       }
 
+      // The protocol is signal-only: frames emitted while the stream was down
+      // are lost forever. Reconcile the chat list on every (re)connect instead
+      // of leaving a missed `chat_session_update` to the 60s safety net.
+      handleChatSessionUpdate(qcRef.current);
+
       const reader = res.body.getReader();
       const decoder = new TextDecoder();
       let buf = "";
--- a/packages/core/src/chat-turn-metadata.ts
+++ b/packages/core/src/chat-turn-metadata.ts
@@ -14,6 +14,12 @@ export const CHAT_TOOL_STEP_BUDGET_DENIAL =
 export interface AppstrateTurnMetadata {
   engine: ChatTurnEngine;
   finishReason?: ChatTurnFinishReason;
+  /**
+   * Client-safe failure message when `finishReason` is "error". Persisted with
+   * the turn (unlike the transient UI-stream `error` chunk) so a reloaded
+   * conversation can still show why the turn failed.
+   */
+  errorText?: string;
   stepCount: number;
   maxSteps: number;
   toolStepBudget?: number;
--- a/packages/module-chat/src/pi-chat/engine.ts
+++ b/packages/module-chat/src/pi-chat/engine.ts
@@ -205,8 +205,18 @@ export function runPiSubscriptionChat(input: PiSubscriptionChatInput): Response 
           if (!turnAbort.signal.aborted) throw err;
         }
 
+        // Invariant: an errored turn ALWAYS surfaces a visible error. The
+        // `error` chunk covers the live client; `errorText` in the persisted
+        // turn metadata covers reloads (error chunks are transient — they never
+        // become message parts). The fallback text guards any capture gap in
+        // the mapper — a silent empty turn is the one unacceptable outcome.
         const meta = mapper.result();
-        if (meta.errorText) write({ type: "error", errorText: meta.errorText });
+        const errorText =
+          meta.errorText ??
+          (meta.finishReason === "error"
+            ? "La génération a échoué (erreur du modèle)."
+            : undefined);
+        if (errorText) write({ type: "error", errorText });
 
         const stepCount = mapper.stepCount();
         write({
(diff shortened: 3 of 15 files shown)


pierrecabriere:
Added `496e608`: sidebar status/timestamp freshness (reported while testing this branch, but pre-existing — dates from the #911 switch to SSE + 60s safety net, this PR's sidebar change was CSS-only):

- **Frozen relative-time labels** — `relativeTime()` was computed at render only; React Query structural sharing keeps `data` referentially stable on identical refetches → no re-render → "à l'instant" forever on a quiet list. A 30s `useNowTick` clock now re-renders the rows.
- **Sticky generating spinner** — the `generating` flip rides a fire-and-forget NOTIFY; a lost SSE frame left the spinner up for the full 60s safety net. `refetchInterval` is now conditional: 3s while any session is `generating`, 60s otherwise.

pierrecabriere:
`3f41c09` — robustness rework of the sidebar status after field testing showed it only worked on a conversation's first message:

**Root cause**: only the first message had an optimistic cache entry; every later turn depended on the full server round-trip (NOTIFY → SSE → refetch), and the server sets `generating` only *after* the multi-second inference preamble. The 3s backstop couldn't cover the start edge by construction (it conditions on the cache already showing `generating`).

**A — local-first**: `ConversationInner` now mirrors `chat.status` into the sessions cache on every turn — spinner on + fresh `updatedAt` + row moved to top on send (also creates the row for new conversations), spinner off + one reconciling invalidate on settle. Subsumes the old one-shot optimistic prepend and the `onFinish` invalidate (both removed).

**B — SSE reconnect reconcile**: the realtime protocol is signal-only, so frames emitted during a disconnect are lost forever; the chat list is now invalidated on every (re)connect of the global stream.

Cross-device start-edge latency (another device sees the spinner only after the preamble) is a known residual — would need a server-side pending-turn registry, deliberately left out (YAGNI) since the reported symptom is same-tab.

pierrecabriere:
`c770806` — provider errors no longer render as an eternally-pending chat (diagnosed from a real failed conversation: 6 blind provider retries, turn persisted as invisible `step-start` parts only, thinking dots animating forever).

Four gaps, each fixed at its own boundary (~60 lines + 3 mapper tests):

1. **Mapper** — pi-agent-core's `handleRunFailure` path (tool exception, context overflow) emits no `message_end`; the errored message only rides `turn_end`/`agent_end`, which the mapper discarded → silent empty turn with `finishReason: "stop"`. Both events now feed failure capture. Tightened along the way: `stopReason: "aborted"` (explicit user stop) is no longer captured as an error.
2. **Engine invariant** — `finishReason: "error"` ⇒ an `error` chunk is always emitted (fallback text if capture missed), and the client-safe `errorText` is persisted in the turn metadata (error chunks are transient — they never become message parts).
3. **Metadata** — optional `errorText` on `AppstrateTurnMetadata`.
4. **UI** — `ThinkingIndicator` only animates while the message is actually running (a dead contentless message no longer reads as hung), and `TurnErrorNotice` renders the persisted `errorText` after reload, suppressed while the live `MessageError` is visible.

Known residuals (deliberate): no retry-progress feedback during the provider retry window, and the ai-sdk path's hard-error case skips the finish chunk (no metadata → no reload notice; the live error and the dots fix still apply).

pierrecabriere:
`88a93f7` — DRY pass on the error display: live failures rendered as the red `ErrorPrimitive` box while reloaded ones rendered as a bare notice line — two components for the same fact. Now a single `MessageError` (one visual, the red box): persisted turn metadata is the preferred source, transient assistant-ui error status is the fallback for failures that never reach a finish chunk. Net −7 lines, `ErrorPrimitive` dropped.

Issue #944, opened 2026-07-23
Free-plan org has credit_quota: 0 / status: "none" — all agent runs fail with 402 quota_exceeded, making the Free tier unusable


OtterHunter:
_Suggested labels: `bug`, `billing`_

## Summary

A freshly created organization sits on the **Free** plan, but its billing account reports an effective `credit_quota` of **0** with `status: "none"` (no subscription attached). The plans catalog advertises Free as granting **5000** credits per cycle. Because the effective quota is 0, **every** agent run — including inline/test runs — is rejected at pre-flight with `402 quota_exceeded`, even though `credits_used` is 0. The Free tier is therefore unusable: no agent can run without attaching a paid subscription, which contradicts offering a Free tier with a 5000-credit quota.

## Environment

- Org on the `free` plan, caller role `owner`
- Org id: `c337ecec-4920-4f5a-ba64-de9f44543dd8`
- Access: Appstrate REST API

## Steps to reproduce

1. Create a new organization (or use one that has never had a billing subscription attached).
2. `GET /api/billing` for that org → note `credit_quota: 0`, `status: "none"`.
3. Create an agent and trigger any run (`POST /api/agents/{scope}/{name}/run`, or an inline run).
4. The run is rejected with `402 quota_exceeded` before it starts.

## Expected behavior

An org on the `free` plan should have `credit_quota` provisioned to the plan's granted amount (5000) with an active status, so runs can execute up to that quota. If Free instead requires an explicit activation step, that step should be discoverable and the run-blocking error should say so.

## Actual behavior

- `GET /api/billing` returns `plan.id: "free"` but `credit_quota: 0`, `credits_used: 0`, `usage_percent: 0`, `status: "none"`, `period_end: null`.
- Every run returns `402 quota_exceeded` — "Credit quota exceeded for org …" — even though nothing has been consumed (`credits_used: 0`). The quota itself is 0.

## Evidence

`GET /api/billing`:

```json
{
  "plan": { "id": "free", "name": "Free" },
  "plans": [
    { "id": "free",    "name": "Free",    "price": 0,  "credit_quota": 5000 },
    { "id": "starter", "name": "Starter", "price": 29, "credit_quota": 20000 },
    { "id": "pro",     "name": "Pro",     "price": 99, "credit_quota": 80000 }
  ],
  "usage_percent": 0,
  "credits_used": 0,
  "credit_quota": 0,
  "period_end": null,
  "status": "none",
  "upgrades": [
[...]

PR #967, opened 2026-07-24
feat(admission): quote-based usage admission for zero-cost BYOK and future compute accounting (deploy cloud module first)

### LINKED ISSUE #961: Make usage admission quote-based for zero-cost BYOK and future compute billing
## Summary

Usage admission should be based on the **estimated total credits billable to the organization**, not on `modelSource === "system"` alone.

#951 fixed the immediate BYOK regression by dispatching `beforeUsage` only for platform-provided system models. That matches today's pricing because:

- platform-funded model usage is billable;
- BYOK model usage is not billable by Appstrate;
- compute is currently free.

However, the filter hard-codes the assumption that a BYOK run is always free. Once platform compute is billed, a platform-hosted BYOK run will have:

```text
model credits   = 0
compute credits > 0
total credits   > 0
```

Conversely, **today** a platform-hosted BYOK run has a total billable cost of zero and must be allowed even when the organization has `credit_quota = 0`.

## Required semantics

| Execution | Model credits | Compute credits | Admission |
|---|---:|---:|---|
| Platform + system model, today | `> 0` | `0` | Check available credits |
| Platform + BYOK, today | `0` | `0` | Allow, including with zero available credits |
| Platform + BYOK, future compute billing | `0` | `> 0` | Check available credits |
| Remote + system model | `> 0`, known at proxy time | `0` | Gate at the system proxy |
| Remote + BYOK | `0` | `0` | Allow |

`modelSource` / `credentialSource` remains necessary for attribution and for calculating the model component. It should not be the top-level decision for whether usage admission runs.

## Proposed direction

Introduce an admission quote representing the estimated billable components:

```ts
interface UsageQuote {
  modelCredits: number;
  computeCredits: number;
  totalCredits: number;
}
```

Conceptually:

```ts
const modelCredits =
  credentialSource === "system"
    ? estimateModelCredits(modelRates, expectedUsage)
    : 0;

const computeCredits =
  executionPlane === "platform"
    ? estimateComputeCredits(computeProfile, timeout)
    : 0;

const totalCredits = modelCredits + computeCredits;
```

For the first phase, `computeCredits` is explicitly `0`; this ticket does **not** need to implement compute charging. It should establish the contract so enabling a non-zero compute rate later does not require changing admission topology again.

The OSS/core boundary should expose neutral execution facts or an estimated provider/compute cost rather than importing Cloud plan definitions. The private Cloud module remains responsible for converting the quote into credits and enforcing plan policy.

## Quota comparison

The current check rejects when:

```ts
creditsUsed + estimatedCredits >= creditQuota
```

That rejects a genuinely free operation on an empty account because `0 + 0 >= 0`.

The balance check should follow:

```ts
const remainingCredits = Math.max(0, creditQuota - creditsUsed);

if (quote.totalCredits > remainingCredits) {
  throw quotaExceeded();
}
```

A separate entitlement/status gate may still reject suspended or otherwise blocked accounts. A zero balance by itself must not act as a generic entitlement flag.

## Important invariants

- `cost: null` means **unknown**, not free. Only a known zero billable estimate may bypass the balance check.
- BYOK may retain a non-zero equivalent catalog cost in the usage ledger for observability; its **billable model component** is nevertheless zero because the organization paid the provider directly.
- Each usage path is admitted exactly once:
  - platform run at run preflight;
  - first-party chat at chat admission;
  - remote system-model usage at the system proxy.
- The quote is an admission estimate, not the final debit. Actual usage remains settled from durable ledgers; future hard caps may add reserve/settle semantics.

## Acceptance criteria

- [ ] Platform BYOK run + zero compute rate + zero available credits is allowed.
- [ ] Platform system-model run with a positive estimate + zero available credits is rejected.
- [ ] A quote exactly equal to the remaining balance is allowed.
- [ ] A quote greater than the remaining balance is rejected.
- [ ] A future-test fixture with non-zero compute credits rejects a zero-balance platform BYOK run without changing the admission call graph.
- [ ] Remote BYOK remains ungated by Cloud credits.
- [ ] Remote system-model usage remains gated at the proxy, without double-gating platform runs.
- [ ] Unknown model pricing is not silently treated as a zero-cost system model.
- [ ] Commercial account-status/entitlement rejection remains independent from the arithmetic balance check.

## Related

- #944 — report that exposed the zero-quota/BYOK interaction.
- #951 — tactical fix that skips `beforeUsage` for non-system models under today's pricing.

pierrecabriere:
## What

Usage admission stops being a boolean and becomes a quote.

Today the `beforeUsage` module hook is dispatched only when the platform has already concluded the operation is billable — in practice, only when the resolved model comes from a platform-supplied credential:

```ts
if (input.modelSource === "system" && hasHook("beforeUsage")) { … }   // run preflight
if (!isSystemModel(args.presetId)) return null;                       // chat
```

That hard-codes **"the org brings its own key ⇒ nothing is consumed"**. It is true today only because platform compute is not accounted for. The moment it is, a platform-hosted BYOK run has `model = 0`, `compute > 0`, `total > 0` — and the platform would be skipping admission for it entirely. Fixing that later would mean changing the admission topology again, in exactly the place where a mistake means either spurious 402s or unmetered usage.

So the platform stops classifying and starts reporting. The hook now fires for **every** run and **every** chat turn, carrying neutral execution facts; the module quotes them and decides.

Closes #961. Follows #951 (which fixed the spurious-402 tactically by skipping the hook for non-system models) and #944 (the report that surfaced the zero-quota/BYOK interaction).

## The contract (`@appstrate/core` 4.0.0 → 4.1.0)

`BeforeUsageParams` gains three facts:

| Fact | Values | Meaning |
|---|---|---|
| `credentialSource` | `"system"` / `"org"` / `null` (runs) | Whose credential is spent on inference. `null` = a remote-origin run resolves its model off-platform. |
| `executionPlane` | `"platform"` / `"remote"` | Whose compute runs the work. |
| `timeoutSeconds` | `number` / `null` (runs) | Effective **post-ceiling** upper bound on platform compute occupancy. `null` = "this seam does not own the run's compute, contribute nothing for it". |

Two properties this shape is chosen for:

- **The two facts are independent axes.** An org can bring its own credential and still occupy a platform sandbox; a platform credential can be used on a caller-supplied host. Anything that collapses them into one signal mis-admits one of those combinations — which is precisely the shape of the pre-filter being removed.
- **No billing vocabulary enters core.** Apache-2.0 core reports what is *consumed*, never what it is *worth*. Turning facts into a policy decision stays entirely in the private module. `timeoutSeconds` is a ceiling the platform already computes, not a price input.

`timeoutSeconds` is nullable rather than using `0` as a sentinel: on a published type, a legitimate value that silently doubles as "ignore me" is a footgun, and widening `number` → `number | null` later would be breaking.

**Minor, not major:** an implementor that only *reads* the params is unaffected. Code that *constructs* `BeforeUsageParams` (module test fakes) must supply the new fields.

## Where the facts come from

| Seam | Reports |
|---|---|
| run preflight (platform origin) | resolved `credentialSource`, `platform`, effective post-ceiling timeout |
| run creation (remote origin) | `null`, `remote` |
| chat admission | resolved preset source, `platform` — a turn always runs in our own process |
| system-proxy seam | `"system"` (it only runs for a resolved system preset), plane from run origin, `timeoutSeconds: null` |

The timeout is now computed **before** the ceiling clone so the hook sees the effective bound. An agent declaring a timeout above the platform ceiling can never occupy compute for longer than the ceiling, so quoting the declared value would over-state it.

## The bypass this preserves

`system-proxy-admission.ts` keeps `runOrigin === "platform" && credentialSource === "system"` as its short-circuit — the same expression, a different meaning: from *"did the hook already run"* to *"was the MODEL component already accounted for"*.

It defends a real attack. A platform BYOK run is now admitted reporting no platform credential, so it is waved through. Attaching that active run id to a raw system-preset `/api/llm-proxy` request would launder platform-funded inference through an operation that was quoted at zero, defeating a rejection. Pinned by a regression test named for the bypass, plus a new test asserting the first-party chat loopback is *not* double-dispatched.

## Observable behaviour changes

Dispatching unconditionally is not free, and both consequences are intended rather than incidental:

- **One extra admission read on previously-skipped paths.** A BYOK platform run and a BYOK chat turn now reach the module, which reads one indexed billing-account row before admitting. Previously they did nothing at all. A *remote* BYOK run still costs nothing — the module short-circuits on the facts before touching its store.
- **A suspended account can no longer run BYOK work on platform compute.** The entitlement gate is independent of the quote, so a blocked subscription is rejected even when the operation quotes zero. That is the point of separating the two gates: a suspended org occupying platform sandboxes for free was the gap. Remote BYOK is unaffected (see the cloud PR for the deliberate divergence there).

## Verification

- `bun run check` — 33/33 tasks green (the 4 lint warnings are pre-existing `module-chat` UI `react-refresh` noise).
- Full tier-0 suite — **8726 pass, 89 skip, 0 fail** across 675 files.
- Tests updated where they asserted the old topology (that reversal is intentional, not a regression), plus new coverage: dispatch for `credentialSource: "org"` and `null`, post-ceiling timeout reported, `runningCount` still projected-inclusive, the proxy bypass, and the chat-loopback non-dispatch.

## Deploy lockstep (read before releasing)

The private billing module must land **first**. Ordering:

1. publish `@appstrate/core` **4.1.0**;
2. deploy the cloud module version that reads the quote;
3. deploy this platform change.

Deploying this platform change against an *older* cloud module would re-introduce #944: the platform stops pre-filtering, and a module still gating on a flat estimate would 402 every BYOK run. In OSS mode nothing changes — with no module loaded the hook is a no-op regardless.

## Prior art / SOTA checks

The design was checked against primary sources rather than assumed:

- Admitting a zero quote instead of gating it has direct precedent — RFC 4006 result code **4011** ("credit control not applicable, e.g. the service is free of charge"), and LiteLLM short-circuits its budget gate on a non-positive reservation. The *opposite* behaviour is filed as a bug against LiteLLM.
- Reporting execution facts at the boundary and keeping plan/price conversion in the metering layer is the RFC 8506 / OpenMeter shape.
- Using a declared ceiling as an *admission* bound while settling on measured actuals is the AWS Bedrock pattern (reserve `max_tokens`, refund the unused remainder, bill actual usage).

**Known divergence, carried deliberately:** the per-run estimate is multiplied by the in-flight count, which has no primary-source precedent — the SOTA answer to concurrent overshoot is an atomic reserve-on-admit / release-on-terminate counter (Bedrock, LiteLLM, RFC 8506). That multiplication is pre-existing behaviour, reserve/settle is explicitly out of scope for this ticket, and it remains a lossy stand-in for a reservation: an orphaned run inflates the count until the platform's orphan sweep reclaims it.

Follow-up worth filing: bound the model estimate by `max_tokens` / the model's output ceiling instead of a flat per-operation figure (also closes a `max_tokens: 10^9` estimate-inflation surface).

## Out of scope

No migration, no schema change, no DTO change. `runs.model_source` keeps its name — it is the same concept under an older persisted name, and the mapping is documented at the seam that reads it. Compute is still not charged. Reserve/settle hard caps remain deferred.

---

### DIFF
--- a/apps/api/src/modules/README.md
+++ b/apps/api/src/modules/README.md
@@ -206,10 +206,22 @@ Full design: `docs/architecture/OBSERVABILITY.md`.
 
 ## Hooks and events
 
-- **Hooks** (`callHook`, first-match-wins): `beforeUsage`, `afterRun`, `beforeSignup`. The first module that provides a hook is called, subsequent modules are skipped. `beforeUsage` gates metered LLM usage on a surface — a discriminated union over `run` (agent run) and `chat` (chat turn); `afterRun` returns a metadata patch persisted on the final run record, `beforeSignup` gates signup.
+- **Hooks** (`callHook`, first-match-wins): `beforeUsage`, `afterRun`, `beforeSignup`. The first module that provides a hook is called, subsequent modules are skipped. `beforeUsage` is the admission gate over LLM usage — a discriminated union over `run` (agent run) and `chat` (chat turn), detailed below; `afterRun` returns a metadata patch persisted on the final run record, `beforeSignup` gates signup.
 - **Events** (`emitEvent`, broadcast-to-all): `onRunStatusChange`, `onOrgCreate`, `onOrgDelete`. Handlers run for side effects only; errors in one handler are isolated and do not block others.
 
 Names are defined in `packages/core/src/module.ts` (`ModuleHooks`, `ModuleEvents`). To add a new hook or event, update that file first so both platform and modules see the same contract.
+
+### `beforeUsage` — admission (core 4.1.0+)
+
+The hook is dispatched for **every** run and **every** chat turn, not for a subset the platform pre-selected. Each dispatch carries neutral execution facts, and the module turns them into a decision:
+
+- **`credentialSource`** — whose credential is spent on inference: `"system"` (platform-supplied credential or system model preset), `"org"` (the organization's own BYOK key or OAuth subscription), or — `run` only — `null` when a remote-origin run resolves its model later on its own host (any inference it then routes through the system model proxy is admitted at that seam instead).
+- **`executionPlane`** — whose compute runs the work: `"platform"` (a sandbox the platform operates, or its own chat process — always the case for `chat`) or `"remote"` (caller-supplied host).
+- **`timeoutSeconds`** (`run` only) — the effective post-ceiling upper bound on platform compute occupancy, or `null` at a seam that does not own the run's compute and must therefore contribute nothing for it (NOT "unknown, assume the worst" — that double-counts).
+
+**Deciding that an operation consumes nothing is the module's job**, not the platform's. The platform used to skip the hook whenever the organization brought its own credential; that assumption breaks as soon as platform compute is accounted for, since a BYOK run still occupies a platform-operated sandbox. A module that only cares about platform-supplied inference reproduces the old outcome by returning `null` when `credentialSource !== "system"`.
+
+An operation the organization supplies entirely by itself — `credentialSource !== "system"` **and** `executionPlane !== "platform"`, i.e. a remote BYOK run — should be short-circuited with `null` before the handler reads any of its own state (no DB round-trip, no account lookup): there is nothing for the platform to account for.
 
 ## Auth strategies
 
--- a/apps/api/src/routes/llm-proxy.ts
+++ b/apps/api/src/routes/llm-proxy.ts
@@ -209,7 +209,11 @@ async function handleProxy(
         context: "run",
         packageId: runAttribution.packageId!,
         runOrigin: runAttribution.runOrigin,
-        modelSource: runAttribution.modelSource,
+        // Persisted `runs.model_source` — the run's credential source under its
+        // older column name. The admission seam reads it to tell a run whose
+        // model component was already quoted at preflight from one that was
+        // quoted at zero (and could otherwise launder system inference).
+        credentialSource: runAttribution.modelSource,
       } as const)
     : c.get("firstPartyLoopback")
       ? ({ context: "chat", sessionId: chatSessionId } as const)
--- a/apps/api/src/services/chat-subscription.ts
+++ b/apps/api/src/services/chat-subscription.ts
@@ -167,11 +167,20 @@ export async function recordChatUsage(record: ChatUsageRecord): Promise<void> {
  * Chat admission gate — the chat-surface entry into the `beforeUsage` hook.
  *
  * The chat module calls this for its non-subscription (built-in / API-key)
- * branch before starting a turn. The gate decides system-provided vs. org-owned
- * SERVER-SIDE (`isSystemModel` on the chosen preset) so the module stays dumb:
- *   - org's own API-key model → never gated (returns null immediately);
- *   - system-provided model → dispatch `beforeUsage` (chat context). A rejection
- *     flows back for the module to surface as an RFC 9457 problem response.
+ * branch before starting a turn. The gate resolves system-provided vs. org-owned
+ * SERVER-SIDE (`isSystemModel` on the chosen preset) so the chat module stays
+ * dumb — it has no model-registry access — but that resolution is REPORTED as
+ * the `credentialSource` fact, not used to pre-filter:
+ *
+ *   - every turn dispatches `beforeUsage` (chat context) with
+ *     `credentialSource` + `executionPlane`; a rejection flows back for the
+ *     module to surface as an RFC 9457 problem response.
+ *   - a turn on the org's own credential reports `credentialSource: "org"`. The
+ *     platform no longer declares it free and skips the hook: a chat turn always
+ *     runs inside the platform's own process, so the platform funds its compute
+ *     even when it funds no inference. A module that meters only
+ *     platform-supplied inference quotes that turn at zero and admits it — same
+ *     outcome as the old early return, but decided by the module.
  *
  * Returns null when no module provides the hook (OSS mode allows everything).
  */
@@ -180,13 +189,18 @@ export async function checkUsageAllowed(args: {
   presetId: string;
   sessionId: string | null;
 }): Promise<UsageRejection | null> {
-  // An org's own model spends the org's own credential — never platform-metered.
-  if (!isSystemModel(args.presetId)) return null;
   if (!hasHook("beforeUsage")) return null;
   const rejection = await callHook("beforeUsage", {
     orgId: args.orgId,
     context: "chat",
     sessionId: args.sessionId,
+    // A chat turn resolves its model on the platform before admission, so the
+    // credential source is always determinable here (never `null`, unlike a
+    // remote-origin run).
+    credentialSource: isSystemModel(args.presetId) ? "system" : "org",
+    // A turn executes in the platform's own process — never on a
+    // caller-supplied host.
+    executionPlane: "platform",
   });
   return rejection ?? null;
 }
--- a/apps/api/src/services/run-creation.ts
+++ b/apps/api/src/services/run-creation.ts
@@ -139,19 +139,29 @@ export async function createRun(input: CreateRunInput): Promise<CreateRunResult>
   // --- Shared preflight: rate, concurrency, timeout cap, beforeUsage hook.
   //     Single source of truth across platform / remote / scheduled origins.
   //
-  //     `modelSource: null` — a remote-origin run resolves NO platform model at
-  //     creation (the runner executes on its own host with its own model +
-  //     credentials; `runs.model_source` stays NULL — see the createRunRow
-  //     comment below). So the run-surface `beforeUsage` hook is skipped: it is
-  //     not cheaply determinable here whether the run will route inference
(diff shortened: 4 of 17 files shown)


pierrecabriere:
Re-running with the `integration` label so the DinD integration suite covers the new admission paths.

pierrecabriere:
### On the red `codecov/patch`

The patch number is an artifact of a comment-heavy diff, not a coverage gap. Measured locally with `TEST_TIER=0 bun test --coverage apps/api/test`:

| File | Codecov patch % | Local line coverage |
|---|---|---|
| `system-proxy-admission.ts` | 15.62% ("27 missing") | **100.00%** |
| `run-pipeline.ts` | 18.75% | 89.79% |
| `run-creation.ts` | 10.00% | 86.18% |
| `run-preflight-gates.ts` | 16.12% | 79.75% |

`system-proxy-admission.ts` is the decisive case: codecov reports 27 changed lines missing coverage on a file whose every executable line is covered. This diff is roughly 70% comments — the seams carry dense load-bearing rationale, and rewriting the stale pre-filter reasoning was part of the change — so the added lines are overwhelmingly non-executable.

The `integration` label is applied and the integration job passes, so this is not the carryforward cold-start case `codecov.yml` documents; the flag reconciliation happened and the number did not move.

I have deliberately **not** touched `codecov.yml` to make this green — the `patch` gate is blocking by design and relaxing it to clear a single PR is a maintainer decision, not a drive-by. Flagging for a call: either an informational override here, or a look at whether the coverage reporter is emitting zero-hit records for comment lines (which would deflate patch % on every comment-heavy PR in this repo).

Everything else is green: `check` 33/33, unit, integration, smoke, CodeQL, Analyze.

pierrecabriere:
### Correction + root cause of the red `codecov/patch`

My previous comment blamed comment lines. **That was wrong** — bun's lcov emits records only for executable lines, verified directly:

```
$ bun test --coverage --coverage-reporter=lcov   # full suite, tier 0
SF:apps/api/src/services/system-proxy-admission.ts
LF:37   LH:37     # 37 executable lines, 37 hit, zero uncovered
```

The actual mechanism is **carryforward merged against shifted line numbers**. Codecov's merged report for that file:

```
system-proxy-admission.ts: 42.04% (37/88)
```

The **37 hits match our LH:37 exactly** — our upload landed fine. The discrepancy is the denominator: codecov thinks the file has **88** lines where the report has **37**. The surplus comes from a carried-forward `integration` session recorded against the *base* version of the file. This diff shifts line numbers substantially (the seams gained ~50 lines of rewritten rationale), so the old session's hit map no longer aligns; the union of the two line sets inflates the denominator and the unmatched lines read as missing.

Same pattern on the others — codecov 45.32% vs 79.75% locally on `run-preflight-gates.ts`, 42.04% vs 100% on `system-proxy-admission.ts`.

**Why this is repo-wide, not PR-specific:** any PR that shifts line numbers in a file whose coverage arrives via a carried-forward flag will show deflated patch coverage. `.codecov.yml` already documents the cold-start half of this (`main` has no `integration` baseline, which is why `project` is `informational: true`); the patch gate has the same root cause but is still blocking.

**Likely resolution without touching policy:** once `main` carries a real post-merge `integration` upload, the carried-forward stale session stops being the baseline and this self-corrects — exactly the condition the `.codecov.yml` comment says to wait for before flipping `project` to blocking.

Still not touching `.codecov.yml` — the call on whether to override here or wait for the baseline is yours.

pierrecabriere:
Folded into #968, which now ships both changes as a single `@appstrate/core@5.0.0`.

Rationale for combining rather than sequencing: this PR adds three **required** fields to `BeforeUsageParams` and makes `beforeUsage` fire for operations it never fired for. That is breaking for anything constructing those params, and it only survives a modules-before-platform deploy — so folding it into an already-major release is free, whereas shipping it afterwards as 4.1.0/5.1.0 would not be honest about what it requires. The 4.1.0 version is never published.

The one substantive divergence was resolved against this PR: the `runOrigin === "platform" && credentialSource === "system"` short-circuit in `system-proxy-admission.ts` stays **deleted**. The preflight gate admits a run *launch*, whose inference flows through the sidecar and never reaches `/api/llm-proxy`; this seam admits one raw proxy call, a distinct billable unit minting its own `llm_usage` row. One quote per launch cannot cover an unbounded number of proxy calls, and `assertRunAttributable` binds an API-key principal only to org + application — so any key in the application can stamp `X-Run-Id` of any live platform system run onto its own calls and keep spending past its quota. The argument that settles it is the one this work exists for: once platform compute is billed, an organization's balance moves *during* the run, so admitting once at launch gates every later call against a stale balance.

Everything else here is preserved in full — the removal of the run/chat pre-filters, the three execution facts, and the timeout computed before the ceiling clamp. At the proxy seam the facts are `credentialSource: "system"` (the seam runs only for a resolved system preset), `executionPlane` derived from the run origin, and `timeoutSeconds: null`.

Known consequence, deliberately deferred and now commented in both repos: with the short-circuit gone, the model component is quoted at a per-run rate on every proxy call, because `BeforeUsageParams` has no discriminant separating a run launch from one proxy call. It over-gates rather than under-gates.

Closing in favour of #968. The billing half is likewise folded into appstrate/cloud#41.

Issue #928, opened 2026-07-21
System packages: 63 packages stuck at 1.0.0 with drifted content — prod serves stale bytes


pierrecabriere:
## Symptom

Every boot logs 63 `level:50` errors (observed live on prod at v1.0.0-beta.40, and previously at beta.39 with 65):

```
System package content changed without a version bump — refusing to overwrite a
published, immutable version. Bump the version in the source manifest; the
previously-loaded bytes remain authoritative.
  packageId: "@appstrate/asana"  version: "1.0.0"
  dbIntegrity:     sha256-BbcsCxoeCFXuwOBKox6npzRJYUnKdnodP1KCQJsO7oc=
  sourceIntegrity: sha256-BZIJUbaTWusUqA88utld4MGxlOeogDr12jupaUFl4ks=
```

## Impact — this is not just log noise

The guard (`apps/api/src/services/system-packages.ts:183`) is working as designed: it refuses to overwrite a published immutable version, so **production keeps serving the previously-published bytes**. Any source change to these packages is inert in prod.

Concretely, two changes shipped in v1.0.0-beta.40 are NOT live:

- #907 — `clickup`/`gmail`/`github` MCP tool-policy sync
- #927 — `clickup_download_task_attachment` declaration

## Root cause

Real content drift, not rebuild noise. Ruled out non-determinism: `zipArtifact` (`packages/core/src/zip.ts:58`) sorts keys and pins `DETERMINISTIC_MTIME`, so identical source yields identical integrity.

The ratio confirms it — 63 of 66 system packages drift, and the 3 clean ones are exactly the three that were ever version-bumped:

- `@appstrate/clickup-mcp` 1.2.0
- `@appstrate/gmail-mcp` 2.3.0
- `@appstrate/github-mcp` 1.1.0

Every package still at `1.0.0` has accumulated uncommitted-to-a-version manifest edits since first publish.

## Why CI does not catch it

`build:system-packages:check` validates source/archive consistency inside the repo. It has no view of what integrity is already published in a database, so drift only surfaces at boot against prod state. Worth considering a check that compares against the deployed registry.

## Fix

For each drifted package, bump the version — source dirs are `scripts/system-packages/{type}-{name}-{version}/`, so this is a directory rename + manifest `version` field + rebuild:

```sh
bun run scripts/build-system-packages.ts
```

Consumers pin `^1.0.0`, so a `1.0.1` bump resolves without breaking dependents.

## Affected packages (63)

[...]

Issue #896, opened 2026-07-10
bundle_integrity_mismatch: published skill artifacts fail integrity check at run time, oscillating and surviving republish (blocks all agents in org)


OtterHunter:
## Summary

Published package artifacts in one org are failing server-side integrity validation at run time with `500 bundle_integrity_mismatch` — *"A stored package artifact no longer matches the integrity hash recorded when it was published, and was refused."* The failure is **not** caused by anything in the manifests or content: it affects multiple skills that publish cleanly (`201`), it **oscillates** between dependencies across consecutive runs, it **survives republishing**, and it occurs in **both `published` and `?version=draft` modes**. This strongly points to **artifact corruption at rest** in the package store (or a mismatch between the stored bytes and the recorded SRI hash), rather than a caller-side problem.

This is the precise diagnosis of what was reported as "Issue 2" (the unhelpful `500 internal_error` post-publish) in #878 — the error now surfaces with a specific code (`bundle_integrity_mismatch`) instead of a generic 500.

**Impact: total.** All agents in the affected org that depend on the corrupted shared skills cannot run at all — the run is refused before execution. In our case four separate domain agents share four common skills, so none of the four can execute.

Environment: Appstrate Cloud, org `<redacted-org-id>`, application `<redacted-app-id>`, model `Claude Opus 4 8` (org-configured), via the REST API through the MCP-exposed operations. Package scope anonymized as `@acme`. One agent (`@acme/agent-a`) depends on five skills: `skill-A`, `skill-B`, `skill-C`, `skill-D` (four shared "common" skills) and `skill-E` (its domain skill).

---

## Symptom

`POST /api/agents/@acme/agent-a/run` (both with and without `?version=draft`) fails with:

```json
{
  "type": "https://docs.appstrate.dev/errors/bundle-integrity-mismatch",
  "title": "Bundle Integrity Mismatch",
  "status": 500,
  "detail": "A stored package artifact no longer matches the integrity hash recorded when it was published, and was refused. Republish the affected package. (Integrity check failed for @acme/skill-A@1.0.0)",
  "code": "bundle_integrity_mismatch"
}
```

The named package in the `detail` **changes from run to run**.

## What makes this a server-side corruption bug (not a caller error)

We attempted the remediation the error message itself suggests ("Republish the affected package"), and observed a **whack-a-mole that never converges**:

[...]

PR #983, opened 2026-07-27
Normalise the manifest a fork mints; gate the module↔core version contract (#974, #973)

### LINKED ISSUE #973: verify-module-contract cannot detect a stale module: it checks ownership, not signatures
## What is wrong

`scripts/verify-module-contract.ts:224-228,370-383` is the only automated guard on the platform↔module contract. Its entire policy is:

```ts
for (const [name, entry] of Object.entries(ledger)) {
  if (entry.owners.length === 0) { problems.push(…) }
}
```

It verifies that every hook and service has at least one owner in a **hand-written ledger**. It checks no arity, no signature, no consumer version. A module compiled against an older `@appstrate/core` therefore passes it in green.

## Why it matters now

PR #972 hardens three signatures in `@appstrate/core@6.0.0`: `checkUsageAllowed` gains a required `subscription` argument, and `beforeSignup` / `afterSignup` gain a required `ctx`. None of those hardenings has an automated guard against an out-of-tree module.

The concrete failure is not a crash — it is silent mispricing. `apps/api/src/services/chat-subscription.ts:234`:

```ts
credentialSource: args.subscription || !isSystemModel(args.presetId) ? "org" : "system"
```

A 5.0.0-era caller omitting `subscription` makes every subscription turn read as platform-funded. No error, no log, wrong billing.

Reproduction: publish a module with `beforeSignup: async (email) => {}` and a `services.checkUsageAllowed({orgId, presetId, sessionId})` call, then run `bun run verify:module-contract` → `✅ module contract clean`.

## Possible directions

- Assert arity and shape against the ledger, not just ownership.
- Have modules declare the `@appstrate/core` major they were built against and refuse to load on a mismatch — the loader already fails hard on a declared-but-missing module, so the seam exists.
- At minimum, make `callHook` log when a hook is invoked with fewer arguments than the contract declares.

Found during the PR #972 adversarial review.

### LINKED ISSUE #974: package-fork and post-install-package publish artifacts from an unvalidated manifest
## What is wrong

`apps/api/src/services/package-fork.ts` and `apps/api/src/services/post-install-package.ts` both call `createVersionAndUpload` with a manifest they never pass through `validateManifest`.

PR #972 fixed the same class of defect one layer down: `PUT /api/packages/…` was persisting `body.manifest` raw while validating a separate copy, and `createVersionFromDraft` was snapshotting the raw draft into `package_versions.manifest` and into the immutable ZIP. Both now normalise. These two paths were not touched — no regression was introduced, but the hole is the same shape.

## Concrete consequence

Forking copies an existing package's manifest verbatim. Fork a legacy agent that still names a retired `runtime_tools` id and the id is carried into a **new** artifact, published after its removal, frozen in an integrity-checked ZIP.

More generally: any invariant enforced at validation time is bypassed on these two paths, so the guarantee "a stored manifest has been validated at least once" does not hold platform-wide.

## Suggested fix

Route both through the same normalisation the version-creation path now uses. Note the helper is deliberately structural rather than a Zod re-parse — key order and unknown fields must survive byte-for-byte so publish dedup (#896) keeps working, and it returns the same reference when there is nothing to drop.

The read/write direction matters here: a fork reads something already stored, so `{ retiredRuntimeTools: "drop" }` is the right policy, not `reject` — a legacy agent must stay forkable.

Found during the PR #972 adversarial review.

pierrecabriere:
Closes #974. Closes #973.

Both issues came out of the PR #972 adversarial review, and both are about a contract that is enforced in one place and bypassed in another.

## #974 — a fork minted an artifact from an un-normalised manifest

A fork is a **read that mints**: it copies an already-published manifest — immutable by construction, so a `runtime_tools` id retired *after* that publish can never be repaired at the source — into a brand new draft row, draft storage files, version row and ZIP. `forkWithConfig` copied it verbatim, so forking a legacy agent regraved the retired id into an artifact minted today.

The manifest is now normalised **once, before `createOrgItem`**, with the same structural dropper `createVersionFromDraft` uses. Placement matters: normalising only the version row would leave the retired id in the fork's draft, and the fork's first re-publish would mint it right back. The drop is structural (no Zod re-parse), so key order, unknown fields and absent defaults survive it untouched and publish dedup (#896) is unaffected.

A source manifest invalid for any *other* reason is **logged, never rejected**: drifted catalog entries exist in production (#928) and a gate here would make them permanently un-forkable.

`postInstallPackage` got the matching guarantee on the read direction — it re-reads `manifest.json` out of caller-supplied ZIP bytes and now refuses to persist a manifest no schema ever saw. **Gate, not rewrite**: that `zipBuffer` is the version's identity, so normalising the DB column alone would make the row diverge from the ZIP, and pinned runs read the manifest from the ZIP. Both of today's callers validate the same bytes upstream, so this is a backstop for a future caller that skips the parser.

## #973 — nothing guarded the contract's *version* dimension

`verify-module-contract` audits the contract's **shape** (every hook, event and service has an owner). It checks no arity, no signature, no version — so a module compiled against an older `@appstrate/core` passed it green. Live proof: the platform ships core 6.0.0, the out-of-tree cloud module pins `^5.0.0`, and the script printed `✅ module contract clean`.

The blind spot is dangerous in **one direction only**. Platform → module (a hook call) tolerates a stale handler — an extra argument is ignored. Module → platform does not: `PlatformServices.checkUsageAllowed` gained a required `subscription` flag in 6.0.0, and a 5.x-era caller that omits it does not error, it reports every subscription turn as platform-funded. Silent mispricing.

Three guards, cheapest first:

- **`checkUsageAllowed` fails closed** when `subscription` is not a boolean (below the `hasHook` early return — in OSS mode nothing prices a turn, so a stale caller keeps getting the `null` it always got).
- **A boot gate**: `@appstrate/core` exports `CORE_VERSION`, `ModuleManifest` gains an optional `core_version` range, and the loader checks a module's declared range (manifest field, else the `@appstrate/core` range in the module's own `package.json`) against it. In-tree modules resolve core from the workspace and are gated by `tsc`, so they skip silently. An unknown range warns; it is a blind spot, not a fault.
- **`scripts/check-consumer-versions.ts`** already covers publish time and is left alone (see below).

### ⚠️ Operator-facing: `MODULE_CONTRACT_ENFORCE` ships as `warn`

`fail` would crash the first deploy that loads `@appstrate/cloud`: this build ships core 6.0.0, npm still serves 5.0.0, and cloud pins `^5.0.0` — it cannot bump until core is published. So the default is `warn` (log the mismatch, boot anyway), and the intended end state is `fail`, documented in `docs/ENV.md`, `.env.example` and the loader.

Flip it once the chain has run: **`afps-shared` → `core@6.0.0` on npm → cloud bump + republish → platform deploy**. An operator who has already done that lockstep should set `MODULE_CONTRACT_ENFORCE=fail` today. Worth filing as a follow-up so the temporary default does not become permanent.

## What the review changed, and one thing it deleted

An adversarial pass over the first two commits produced the third. The notable outcome: the core-version check that had been added to `verify-module-contract` was **deleted rather than fixed**. `scripts/check-consumer-versions.ts` already does the job better — it reads every consumer repo over the GitHub API (so it works where the sibling repos are not checked out) and runs in `publish-core.yml` before `npm publish`, the moment drift matters. The new check was a strict subset of it, a guaranteed no-op in CI (every `DECLARER_ROOTS` entry resolves `workspace:*`; only `cloud` carries a real range and `cloud/` is never checked out there), it re-implemented range parsing with a regex that mis-read `>=5 <7` and `<7.0.0`, and it turned `bun run check` red on any dev machine with the sibling repo cloned — overriding a drift deferral `packages/core/CHANGELOG.md` documents deliberately. The script keeps a pointer to where the version dimension *is* covered.

Also from the review: the built-in short-circuit now wins over `manifest.core_version` (a `core_version` copy-pasted from the authoring README must not fail the boot of a module `tsc` already gates); the loader resolves `<specifier>/package.json` before walking up from the entry file (the walk silently found a `dist/` `{"type":"module"}` marker instead of the real manifest); and three comments that asserted something false were corrected.

## Deliberately not built

- **Arity assertion in the ledger** — it is offline, cannot see an absent module's code, and the platform's own arity is `tsc`-guaranteed.
- **A `handler.length` warning in `callHook`** — false-positives forever on an up-to-date module that legitimately ignores a trailing parameter (`async (email) => {}`). The version gate covers the real case.
- **Prerelease handling in the gate** — documented as a known limitation instead: were `CORE_VERSION` ever `7.0.0-beta.1`, every module declaring `^7.0.0` would fail at once. Core has never published a prerelease and the fix would widen a core helper's API for a case that does not exist.

## Verification

- `bun run check` — 33/33 tasks green.
- `TEST_TIER=0 bun test` — **8916 pass / 89 skip / 1 fail**. The single failure (`llm_usage durable retry … the worker PERSISTS it`, a 5 s timeout under full-suite load) is **pre-existing**: the same run on `origin/main` gives 8897 pass and the identical failure, and the file passes 9/9 in isolation. Untouched by this branch.
- Every new test was checked against pre-fix code: the two fork normalisation tests, the post-install gate test and the three loader gate tests fail without their fix. Two of the new tests are forward guards rather than regressions and say so in their comments (the byte-stability one passes pre-fix by construction).


pierrecabriere:
Suivi du flip `MODULE_CONTRACT_ENFORCE` `warn` → `fail` (après `afps-shared` → `core@6.0.0` sur npm → bump + republish cloud) : #984.

pierrecabriere:
Audit de sur-ingénierie passé sur le diff de cette PR (4 chasseurs + 5 vérificateurs adversariaux). Résultat dans le commit `49171ca9f` — net −59 lignes :

- **walk-up de `_coreRangeFromPackageJson` supprimé** — code mort (jamais exécuté par la suite) **et** nuisible : pour un module désigné par chemin, il remontait au `package.json` ancêtre déclarant `workspace:*`, ce qui **sautait silencieusement le gate**.
- **`ModuleManifest.core_version` supprimé** — 0 déclarant, et placé *au-dessus* de la range résolue, donc il ne pouvait que dégrader le verdict. Fenêtre saisie : core npm est encore en 5.0.0. `CORE_VERSION` reste.
- **gate `validateManifest` de `postInstallPackage` supprimé** — doublon de `parsePackageZip` sur les mêmes octets ; matrice de 10 manifests drifted : aucune régression possible.
- **citation `#928` corrigée** (code + CHANGELOG) : les 66 manifests figés par #928 sont tous valides — c'est de la dérive d'octets, pas de schéma. Vraies preuves : migration #481 et le chemin d'import ZIP, désormais suivi en #987.
- 5 tests du gate déplacés de `loadModulesFromInstances` (aucun appelant de prod) vers `loadModules()` sur des fixtures disque : ils épinglent maintenant le chemin réellement emprunté en production.

Gardés après tentative de sauvetage : le gate de boot + `MODULE_CONTRACT_ENFORCE` (`beforeSignup`/`afterSignup` ont aussi gagné un `ctx` requis en 6.0.0 sans aucune garde runtime → `checkUsageAllowed` ne couvre qu'1 signature sur 3) et le warn du fork (seul chemin de l'API qui mint un manifest sans contrôle de schéma).

Bug hors périmètre trouvé au passage : #987.

Issue #989, opened 2026-07-27
Core test harness mounts module routers without init(): chat billing gate and document teardown are inert in tests


pierrecabriere:
The core `apps/api` integration suite mounts the chat module's router **without running its `init()`**, so every request it serves runs against a degraded dependency baseline in which two post-incident guards are silently inert.

## The defect

`apps/api/test/helpers/app.ts:274-280` calls `mod.createRouter?.()` directly. `getDiscoveredModules()` auto-discovers `packages/module-*/src/` by default, so `@appstrate/module-chat` is included.

Because `init(ctx)` never runs, `createChatRouter` falls through to `buildChatPlatformDeps()` with no context (`packages/module-chat/src/index.ts:65`), and the fallback ladder at `packages/module-chat/src/platform-services.ts:116-149` supplies:

| Dependency | Fallback | What it disables |
| --- | --- | --- |
| `checkUsageAllowed` | `Promise.resolve(null)` (`:146-147`) | the #968/#971 admission / quota gate — **fail-open** |
| `cleanupSessionDocuments` | `Promise.resolve()` (`:137-143`) | the #965 atomic document teardown — **silent no-op** |
| `rateLimit` | `passThroughRateLimit` (`:114`) | per-route rate limiting |

## Why it matters

Both disabled guards were written in response to incidents. Any test in the core suite that believes it exercises the billing admission gate or chat-session document teardown is exercising nothing, and would keep passing if either guard were deleted outright. This is a test-fidelity defect, not a production breach — the production path is unaffected (see below).

## Production is NOT affected

`apps/api/src/lib/modules/module-loader.ts:329` -> `:348`: `_modules.set(...)` runs only after `await mod.init(ctx)` returns, and a throw at `:331` is fatal to boot. `registerModuleRoutes` (`:490-491`) iterates `_modules` exclusively. So any module whose `createRouter()` the platform calls has provably run `init()`.

## Secondary finding: the justifying comment is false

`packages/module-chat/src/platform-services.ts:109` and `packages/module-chat/src/index.ts:64` both justify the fallback ladder with "OSS standalone wiring may skip init". `createChatRouter` has exactly **one** caller in the entire workspace — `packages/module-chat/src/index.ts:65`. There is no OSS standalone wiring. That claim is the sole stated justification for a fail-open branch on a billing gate and should be removed.

## Proposed fix

[...]

Issue #988, opened 2026-07-27
normalizeAuthScheme silently corrupts raw credentials that start with a scheme name


pierrecabriere:
`normalizeAuthScheme` rewrites the whole `Authorization` value, so a legitimate raw credential whose first bytes spell a scheme name is silently corrupted before it reaches the upstream.

## The defect

`packages/connect/src/proxy-primitives.ts:155-180`

```ts
/^(Bearer|Basic|Token)(?=[^\s])/i
```

The regex is correctly scoped to the `authorization` / `proxy-authorization` header **names** (`:165`, `:176`), and well-formed values (`Bearer abc`, `Token abc`) are untouched — the `(?=[^\s])` lookahead fails on them. The problem is that it matches the **resolved value**, so a raw secret that happens to start with `bearer` / `basic` / `token` gets a space injected mid-token:

```
"basically_a_key_123"  ->  "basic ally_a_key_123"
"tokenlive_sk_123"     ->  "token live_sk_123"
"bearerXYZ"            ->  "bearer XYZ"
```

The upstream then receives a mangled credential and answers 401. Nothing logs the rewrite, so the failure looks like an invalid credential rather than a platform-side mutation.

## How it is reached

The normalisation runs **after** credential injection on both entrypoints:

- `apps/api/src/services/credential-proxy/core.ts:344` -> `:346`
- `runtime-pi/sidecar/credential-proxy.ts:454` -> `:455`

so the injected value is in scope. Two live paths put a raw secret in `Authorization`:

1. An integration declaring `credential_header_name: "Authorization"` with **no** `credential_header_prefix` — `buildInjectedCredentialHeader` (`proxy-primitives.ts:112`) emits the bare token.
2. The model writing `Authorization: "{{api_key}}"` through the free-form headers surface (`CREDENTIAL_PROXY_INPUT_SCHEMA`, `runtime-pi/sidecar/mcp.ts:691-698`), substituted at `credential-proxy.ts:453-454`.

## Why the function should not simply be deleted

It covers a real authoring defect. The MCP `api_call` surface exposes free-form `headers` to the model, and a model writing `Authorization: "Bearer{{access_token}}"` produces `Bearerghp_...` on the wire and a hard 401. The declarative `credential_header_prefix` path cannot emit that shape, but it is not the only producer.

## Proposed fix

Repair the **template**, before substitution, instead of the **resolved value** after it:

```ts
[...]

Issue #984, opened 2026-07-27
Flip MODULE_CONTRACT_ENFORCE to fail once core 6.0.0 is published


pierrecabriere:
## What

`MODULE_CONTRACT_ENFORCE` ships as `warn` in #983. Flip it to `fail` once the core 6.0.0 release chain has run.

## Why it is `warn` today

The boot gate added in #983 compares the `@appstrate/core` range a module declares (`manifest.core_version`, else the range in its own `package.json`) against the platform's `CORE_VERSION`. With `fail` as the default, the first deploy that loads `@appstrate/cloud` would crash at boot:

- the platform builds `packages/core@6.0.0`;
- npm still serves `@appstrate/core@5.0.0`;
- `cloud/package.json` pins `"@appstrate/core": "^5.0.0"` and **cannot** bump before core is published.

So `warn` is a temporary state, not the intended design. The mispricing the gate exists for is already blocked independently: `checkUsageAllowed` throws fail-closed when `subscription` is not a boolean, whatever this var says.

## Preconditions for the flip

1. `@appstrate/afps-shared` published (if its range moved).
2. `@appstrate/core@6.0.0` published to npm (tag `core@6.0.0`).
3. `cloud` bumped to `^6.0.0`, adapted to the 6.0.0 signatures, republished — plus any other out-of-tree module in the deployed `MODULES` list (`@appstrate/module-claude-code` is recorded at `^2.19.0`).
4. Prod `.env` reviewed: an operator who has completed the lockstep can set `MODULE_CONTRACT_ENFORCE=fail` before the code default changes.

## The change itself

- `packages/env/src/index.ts` — `MODULE_CONTRACT_ENFORCE: z.enum(["fail", "warn"]).default("fail")`
- `apps/api/src/lib/modules/module-loader.ts` — drop the "warn ships as the default only because the release chain has not run" paragraph from `enforceCoreVersionContract`'s JSDoc.
- `docs/ENV.md`, `.env.example`, `apps/api/src/modules/README.md` — same trim.
- `apps/api/test/unit/modules/module-loader.test.ts` — the tests that opt into `fail` via a local `enforceFail()` helper become default-path tests; the "boots under the shipped default" test flips to asserting a throw.

## Do not

[...]

PR #994, opened 2026-07-27
Pre-beta.42 hardening: credential corruption, inert test guards, 63 drifted system packages

pierrecabriere:
Three pre-release fixes, one per commit. No migrations, no new required
environment variable.

## `fix(connect)` — a raw credential starting with a scheme name was corrupted (#988)

`normalizeAuthScheme` ran on the **resolved** `Authorization` value, after
credential injection. Its regex matched any secret whose first bytes spell a
scheme name and injected a space mid-token:

| sent by the integration | reached the upstream |
| --- | --- |
| `basically_a_key_123` | `basic ally_a_key_123` |
| `tokenlive_sk_123` | `token live_sk_123` |
| `bearerXYZ` | `bearer XYZ` |

The upstream answered 401 and nothing logged the rewrite, so a platform-side
mutation looked like the user's credential being invalid.

The repair now runs on the caller **template**, before substitution, anchored on
`{{`. That covers 100% of the authoring defect it exists for — a model writing
`Authorization: "Bearer{{access_token}}"` through the free-form `api_call`
headers surface — with zero false positives, because a template is never a
secret.

**The post-injection pass could never have repaired anything.**
`buildInjectedCredentialHeader` builds `${prefix} ${token}` itself, so the
declarative path always had its space. On that path the pass could only ever
corrupt. It is removed from both entrypoints.

Coverage: the primitive is unit-tested against all three corrupting values plus
the defect it must still repair; `executeApiCall` carries the same cases
end-to-end. The API entrypoint shares the primitive and its remaining change is
a deletion, so it has no new test of its own.

## `test(harness)` — two post-incident guards were inert in every test (#989)

The test harness called `mod.createRouter?.()` straight off each discovered
module, so `init(ctx)` never ran. Modules carrying a no-context fallback then
served every test request against a degraded baseline. For chat that meant:

| dependency | fallback | what it disabled |
| --- | --- | --- |
| `checkUsageAllowed` | `Promise.resolve(null)` | the #968/#971 admission gate — **fail-open** |
| `cleanupSessionDocuments` | `Promise.resolve()` | the #965 document teardown — **silent no-op** |

Any test that believed it exercised either guard exercised nothing, and would
have stayed green if the guard were deleted outright.

Production was never affected: `_modules.set(...)` runs only after
`await mod.init(ctx)` returns, and `registerModuleRoutes` iterates `_modules`
exclusively.

The preload now calls `loadModulesFromInstances()` — an entry point that already
existed *"for tests"* and runs the same topo-sorted init as production. It lives
in the preload, not the app helper, because `getTestApp()` is synchronous and
hundreds of call sites depend on that.

With the harness at parity the chat fallback has no remaining consumer, so
`buildChatPlatformDeps(ctx)` now takes a **required** context and
`createRouter()` throws rather than serving a baseline that looks like it works.

`test-harness-module-init.test.ts` pins the parity. **Verified by
falsification** — disabling the preload's init step fails it. The whole suite
passes unchanged (9086 pass / 0 fail), which is itself the finding: nothing
depended on the degraded baseline.

## `fix(system-packages)` — 63 archives drifted from their published version (#928)

Production has logged 63 `level:50` errors at every boot since beta.39, and the
guard is right to refuse: it keeps serving the previously-published bytes. The
consequence is that source changes to those packages were **inert in
production**, including #907 (clickup/gmail/github tool-policy sync) and #927
(`clickup_download_task_attachment`) — both shipped, neither ever live.

Which packages drifted was **measured, not inferred**. `zipArtifact` is
deterministic (sorted keys, fixed mtime), so for each committed archive the
bytes at the commit that *added* it — the revision that minted that version —
must still equal the bytes on disk. Byte-comparing all 66 gives **63 drifted, 3
clean**, and the 3 clean ones (`gmail-mcp-2.3.0`, `github-mcp-1.1.0`,
`clickup-mcp-1.2.0`) are exactly the ones production does not report. Both
sides agree on the count, so this is the whole set and nothing else.

Each of the 63 gets a patch bump: manifest version rewritten textually (key
order and formatting preserved — a JSON round-trip would have reformatted 63
files for nothing), source directory renamed, archives rebuilt.

`syncCanonical` picks the highest semver, so the corrected content becomes
canonical on the next boot. Nothing is destroyed: the immutable `1.0.0` rows
stay in the database for anything pinned to them, and the sync prunes no
versions.

`github-git-mcp.test.ts` imported its subject through a version-pinned path,
which this bump broke; it now resolves the directory, so the next bump will not
break it again.

## Verification

- `bun run check` — 33/33 tasks green
- `bun test` — 9086 pass, 46 skip, **0 fail** (697 files), against a freshly
  reset test stack
- `bun run build:system-packages:check` — 66/66 archives match their sources

## Deploy notes

No migration. `MODULE_CONTRACT_ENFORCE` stays `warn` (#984 is still gated on the
core 6.0.0 release chain). On the first boot after deploy, the 63 boot-time
`level:50` errors should disappear — that is the observable acceptance test.


pierrecabriere:
Precision on the #928 blast radius, so the deploy expectation is not overstated.

`application_packages.version_id` is nullable and is only ever set through an explicit `PATCH /api/applications/:id/packages/:packageId` carrying `version_id`. Installs that never pinned carry NULL, resolve through the canonical (highest-semver) entry, and therefore pick up `1.0.1` on the first boot after deploy — that is the majority path and the one that makes #907/#927 live.

An install that **did** pin `1.0.0` keeps resolving the immutable `1.0.0` row, by design. Those stay on the old bytes until re-pinned. Nothing breaks for them; they simply do not receive the fix yet.

So the acceptance test after deploy is the boot log (63 `level:50` errors → 0), not "every org now has the fixed tool policy".

PR #1008, opened 2026-07-27
fix(models): stale subscription model lists (claude-code, codex)

pierrecabriere:
## Summary

Connecting a Claude Code or Codex subscription reported stale models — and reported *two different* stale lists. The CLI helper printed `claude-opus-4-7, claude-sonnet-4-6, claude-haiku-4-5` while the dashboard showed `claude-fable-5, claude-opus-4-8, …`, at a moment when the vendored catalog already carried `claude-opus-5` and `claude-sonnet-5`.

Three independent causes, fixed in order:

**1. Hand-curated lists with no correction mechanism.** Each subscription provider declared two hardcoded arrays (`featuredModels`, `modelDiscoveryCandidates`). `SUBSCRIPTION_COMPLIANCE.md` forbids probing these providers to enumerate models, so nothing could ever correct a stale snapshot — and nothing did, for two model generations. `claude-code` now declares a `CatalogModelSelector` and derives both lists from the vendored Anthropic catalog, so the weekly refresh carries a new generation through on its own.

`codex` deliberately stays an explicit array: the ChatGPT sign-in set is defined by [OpenAI's documentation](https://learn.chatgpt.com/docs/models) and is narrower than the OpenAI API catalog, so deriving would over-list badly. Its list is bumped to the documented set and the ids deprecated for sign-in are dropped.

**2. A persisted copy of a pure function.** For `modelDiscovery: { mode: "static" }` providers, `available_model_ids` held no per-credential information — verified in production, where all five non-null `claude-code` rows carried an identical array. What it did hold was a snapshot written once and never refreshed, which is why a corrected provider definition still surfaced the old list. It is now resolved at read time through one accessor. Probe providers are untouched: there the column is genuinely per-credential and stays authoritative.

**3. A third source for the redeem response.** The pairing redeem returned `featuredModels` — a 3-id subset — while the dashboard rendered the credential's servable set. The two surfaces disagreed *by construction*, on every connection, even when both lists were current. Both now read the same accessor.

Plus the reason none of this was caught: the weekly refresh regenerated `subscription-watch/chatgpt.json` but tested drift only against `apps/api/src/data/pricing`, so a week where only the subscription snapshot moved opened no PR and discarded the snapshot on runner teardown. The review signal that both the refresh script and the codex provider point at had never been able to fire.

## What users see after this

| Surface | Before | After |
| --- | --- | --- |
| CLI helper | `claude-opus-4-7, claude-sonnet-4-6, claude-haiku-4-5` | same list as the dashboard |
| claude-code featured | opus-4-7, sonnet-4-6, haiku-4-5 | opus-5, sonnet-5, haiku-4-5, fable-5 |
| claude-code selectable | 7 ids topping out at opus-4-8 | 8 ids topping out at opus-5 |
| codex featured | gpt-5.5, gpt-5.4, gpt-5.4-mini | gpt-5.6-sol, gpt-5.6-terra, gpt-5.6-luna |

## Guardrails added

- **Blocking test gate** (`curated-model-drift.test.ts`) — fails CI when the vendored catalog *or* the subscription-watch snapshot holds an id newer than a curated static provider's list. Generic over "static provider with an explicit array"; selector-backed lists re-derive themselves and need no gate. Ids a subscription legitimately does not serve go in `subscription-watch/reviewed.json`, a decision record with a documented reason per id — not a mute button. Known blind spot, stated in the test header: the floor is per-stem, so an entirely new stem (`o5`) has no floor and is not compared.
- **Loud boot check** — a selector that resolves to no model now throws at registration, naming the provider, catalog and families. Previously an unknown catalog or a mistyped family resolved to `[]` and registered silently with an empty picker.
- **Weekly drift scope** widened to every generated tree under `apps/api/src/data`, with the PR body naming which one moved.

## Review notes

- **Migration 0030** clears `available_model_ids` for `claude-code` / `codex` rows, which are inert once derived. Idempotent, scoped to two provider ids, no schema change. Rollback with the migration applied is self-healing: the old code reads `null`, the seed gate falls back to the old hardcoded ids, and the next "Refresh models" repopulates.
- **`@appstrate/core` contract**: `featuredModels` widens from `readonly string[]` to a union. Non-breaking for module authors, breaking for any *reader* of `ModelProviderDefinition`. `packages/core` is already on an unpublished `6.0.0` (`git tag core@*` tops out at `5.0.0`), and no consumer repo reads the field. No version bump in this PR.
- **`connect-helper`** (separate repo) reads `availableModelIds` as a `string[]`. Field name, type and requiredness are unchanged — only the value's source. It keeps working untouched and simply prints the right list.
- One deliberate reduction: `claude-opus-4-6` and several deprecated `gpt-5.x` ids leave the *selectable* set. The seed gate is create-time only, so existing `org_models` rows and in-flight runs are unaffected — verified against production, which already has an `org_models` row (`gpt-5.4-nano`) outside the candidate list and working.
- Two adversarial review passes were run against this branch; the second is reflected in the last commit.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactoring

## Checklist

- [x] `bun run check` passes (33/33)
- [x] `bun test` passes — `apps/api/test/unit` 1292/0, module suites 817/0, touched integration files 66/0
- [x] I have read the Contributing Guide


pierrecabriere:
Deferred follow-up filed as #1009 — resolving the model provider credential at run time instead of binding it in `org_models`. That is the change that removes the N×M duplication users actually feel; it is breaking, touches billing, and its premise (does collapsing the static/probe seam pay for itself?) needs settling before code. Deliberately not mixed into this PR.

pierrecabriere:
### Two adversarial review passes — what they changed, and two corrections to this PR's own claims

**Pass 2 corrected the severity of a guard added in pass 1.** The loud boot check threw inside `bootCritical()`, which `index.ts` awaits with no `try` — so a selector resolving to nothing exited the process. One of its inputs is the weekly bot-refreshed pricing catalog, meaning an upstream family rename would have taken the whole API down for a model-picker problem. The guard is now split by cause: a selector over a catalog that does not exist still throws (that is source code, no refresh can produce it), while a selector over an existing catalog that resolves to nothing logs at error level and registers empty.

**It also caught the branch contradicting its own thesis.** Four tests pinned catalog *contents*, so adding `claude-opus-6` to the vendored catalog — exactly what the weekly refresh PR does — turned them red, in a PR whose whole argument is that a new generation should flow through with no code change. They now pin invariants against the real catalog and exact output against a synthetic one.

Two claims in the description above deserve tightening:

1. **The parser divergence was latent, not live.** At the previous commit the only declared selector was `claude-code` over the Anthropic catalog, whose ids all use the 8-digit date form the old parser did handle. Running the old parser against today's real catalogs produces byte-identical output for every shipped list. The fix is correct and the second parser had to go, but no user-visible list was ever affected by it.
2. **Dropping `gpt-5.2` and `gpt-5.3-chat-latest` reverses a deliberate policy.** The comment on `main` kept recently-deprecated ids "so plans that still serve them keep them selectable"; this PR drops them on the authority of the vendor doc, which states both are deprecated for ChatGPT sign-in. Production has no `org_models` row on either id (its Codex rows are `gpt-5.4-mini`, `gpt-5.4-nano`, `gpt-5.5`), so there is no live regression — but it is a policy change, not a cleanup, and `gpt-5.2` sits below the drift gate's floor, so nothing will resurface it automatically if that reading turns out to be wrong.

Residual risk, named rather than fixed: the error-log arm of the guard is the only alarm for a selector that stops resolving. The invariant test ("one id per declared family") does fail in CI on the refresh PR that would cause it, so the bot path is covered; a rename landing by another route is not.

Issue #1021, opened 2026-07-27
DX papercuts: manifest $schema URL 404s, skills-list contract wording, api passthrough drops 307 body, dead `dependencies.tools` key accepted


pierrecabriere:
Four small, independent papercuts collected in one session of authoring and publishing an agent package. Each is cheap to fix on its own; splitting them into four issues felt like more noise than signal, happy to split if you'd rather track them separately.

---

### 1. The `$schema` URL every manifest declares returns 404

Every package manifest carries:

```json
"$schema": "https://afps.appstrate.dev/schema/v1/agent.schema.json"
```

That URL is a **404**. The real schema lives at `packages/core/schema/agent.schema.json` and is never served. Any editor or tool that resolves `$schema` (VS Code's JSON language server, `ajv` in remote-ref mode, a CI manifest linter) gets nothing — so authors lose completion and validation on the one file where a typo silently changes behaviour.

Either serve the schema at that URL (the docs domain already exists), or stop advertising a URL that doesn't resolve. Same applies to `https://schemas.afps.dev/v0/mcp-server.schema.json`, declared by the system mcp-server packages.

---

### 2. `GET /api/packages/skills` documents itself as org-wide but returns a catalogue view

The OpenAPI summary says *"List all skills (system + org) in the organization"*, and documents `?active=true` as an opt-in narrowing. Observed:

```
/api/packages/skills               → 1 skill
/api/packages/skills?active=false  → 1 skill
/api/packages/skills?active=true   → 1 skill
/api/library → packages.skill      → 7 skills  (same org)
```

The code is doing what it intends — `listOrgItems` defaults to `or(source = 'system', application_packages.packageId IS NOT NULL)` (`apps/api/src/services/package-items/crud.ts`), i.e. system packages plus **org packages installed in the current application**. The six missing skills exist and are org-owned; they are simply not installed in that app, which `/api/library` shows plainly (`installed_in: []`).

So this is a contract/wording bug, not a filtering bug — but it is genuinely misleading: it made me conclude a skill had been deleted from the instance. `/api/packages/agents` returns all 7 agents, which reinforces the wrong mental model (agents happen to all be installed).

Ask: say what it does in the summary ("system packages + org packages installed in this application"), and consider `?scope=org|app` for the org-wide listing that the current wording promises.

---

### 3. `appstrate api GET /api/documents/{id}/content` silently yields 0 bytes

[...]

PR #1030, opened 2026-07-27
fix: DX papercuts — package list contract wording, CLI 307 passthrough, retired `dependencies.tools` (#1021)

pierrecabriere:
Fixes items 2, 3 and 4 of #1021. Item 1 (the `$schema` 404) is deliberately left out — it is being handled separately.

All three turned out to be real, but **two of the issue's own proposed remedies would have caused more damage than the papercut**, so this PR resolves the underlying complaints a different way. Details below.

---

## 2 — `GET /api/packages/{type}` documented itself as org-wide

**Verdict: valid, and documentation-only.**

`listOrgItems` has always returned the *application-scoped* catalogue — system packages, plus organization packages with an `application_packages` row for the current `X-Application-Id`. That is intentional and consistent: the same filter backs `GET /api/agents`, and the package **detail** route gates on it through `hasPackageAccess`, so a package the list hides is one the detail route 404s. It is commented in `listOrgItems` itself and locked by tests. Only the four OpenAPI descriptions were wrong.

Why agents looked complete while skills did not: auto-install fires on create / import / fork, and bundle import installs **only the bundle root** (an agent). Skills that arrive as bundle dependencies never get an install row.

**Fix:** say what the endpoints do, and point at `GET /api/library` for the organization-wide catalogue — which already returns these rows *plus* per-application `installed_in` state. Also corrects the `active` parameter description (it drops system packages with no install row, and integrations resolve through effective activation so environment-provided system integrations stay listed).

**`?scope=org` was considered and rejected.** It would duplicate `/api/library` while returning strictly less information; adding it to the list without the detail route would return ids whose detail endpoint 404s, and making detail follow means changing an access-control boundary; it multiplies the `active` matrix with a contradictory cell (`active` is application-relative); and no caller asks for it. If per-type organization-wide listing is ever wanted, a `?type=` filter on `/api/library` is the cheaper move.

No behaviour change, no new query parameter.

## 3 — `appstrate api` wrote 0 bytes on a 307

**Verdict: valid, with one correction to the framing — `-L` already exists and already works.**

The route answers `307` with a presigned `Location` when object storage has a public endpoint (filesystem storage, and S3 without one, proxy-stream a normal `200` — which is why this only reproduces on deployments configured that way). `redirect: "manual"` is a deliberate, test-pinned default, and Hono's `c.redirect()` sends no body, so the passthrough faithfully streamed zero bytes and reported success. The gap was that the default path said **nothing**.

**Fix:** print a hint on stderr naming the status and the `Location`. Everything else is unchanged — default stays `manual`, exit code stays 0, stderr only so piped and `-o` output stay byte-exact, silenced by `-s` like the 401 hint, and skipped for `-I`/HEAD where a bodyless response is expected rather than surprising. Scoped to the fetch redirect statuses rather than all of 3xx: a `304` is bodyless by design and a `300` does carry a body.

**Following redirects by default was considered and rejected**, on measured evidence rather than assumption. On Bun 1.3.11 a cross-origin hop drops `Authorization` and `Cookie` — but **not** custom headers. `buildHeaders` merges arbitrary `-H` values alongside `X-Org-Id`/`X-Application-Id`, and only the *initial* URL is origin-validated, never the `Location`. Making `-L` implicit would turn every `appstrate api` call into a server-directed fetch that forwards user `-H` secrets to a host the server chooses.

That same measurement exposed a documentation bug worth its own mention: the `-L` help text and README said cross-origin hops strip `Authorization`. True, but incomplete in a way that invites a mistake — a reader would reasonably conclude `-L` is safe while passing a second credential via `-H`. Both are corrected. A stale comment claiming `apiFetchRaw` hard-codes `redirect` is also fixed.

## 4 — `dependencies.tools` accepted silently

**Verdict: valid, and it is a regression.**

AFPS 1.0 defined `dependencies.tools` and `dependencies.providers`; AFPS 2.0 renamed them to `mcp_servers` and `integrations`. The platform *did* reject the old spellings at publish time via `assertNoLegacyDepKeys` — until fb318b3bf (#528) deleted that guard along with the 1.x read-fallbacks, on the reasoning that all 1.x back-compat was dead. The read-fallbacks were. The guard was not: it was the author-input rejection.

**The issue's proposed fix — `additionalProperties: false` on `dependencies` — is not implemented, for two independent reasons:**

1. **It breaks published data.** Stored manifests are re-validated on *every run*; a published agent carrying `dependencies.tools` would become permanently unrunnable with a 500 and no repair path, since the artifact is immutable and integrity-checked. That is precisely the failure mode `RetiredRuntimeToolsPolicy` was created to prevent.
2. **It contradicts AFPS.** The spec mandates extensibility for objects it does not explicitly close (§10). `dependencies` is not closed, so `dependencies._meta` is legal today — and `@afps-spec/schema` is a separately published, externally consumed spec. Closing that object is a spec revision, not a schema tweak.

A non-blocking warning alone was also rejected as too weak for author input — that repeats the silent-drop mistake the existing policy doc argues against.

**Fix:** mirror `RetiredRuntimeToolsPolicy` exactly. Reject on author input naming the replacement key; tolerate on already-persisted manifests, unchanged; warn on bundle import through the existing `warnings` channel. The schema is untouched and stays loose — this is a policy on author input, not a shape constraint.

The retired vocabulary lives in one table both the rejection and the warning read, so they cannot enumerate a different set. The direction stays a **single** flag rather than a second one that must co-vary — a call site setting one and forgetting the other is exactly how this regressed the first time. The flag keeps its historical `retiredRuntimeTools` name (documented as governing both) because renaming published `@appstrate/core` surface is a breaking release for a naming nicety.

---

## Heads-up for whoever merges this

The issue notes that one of the reporter's older agents still carries `dependencies.tools`. After this change, **submitting that manifest through the author path returns a 400** — the intended signal, and the parity the issue asks for.

It is mostly self-healing in practice: the agent editor now normalises a loaded draft the same way it already did for retired runtime tools, so opening and saving such an agent silently drops the key. Without that, the agent would be *uneditable* — a 400 on a field the form cannot display. Only the agent branch needs it; the skill and integration editors expose a raw-JSON tab where the key is visible and fixable.

A raw API `PUT` carrying the key still 400s by design. Worth a line in the release notes. Nothing in this repo carries a retired key — all 66 shipped `system-packages/*.afps` manifests were checked structurally (the `tools` field on the mcp-server package is the unrelated MCPB root advisory list).

## Verification

- `bun run check` — 33/33 tasks green (the single `react-refresh` warning in `module-chat` is pre-existing and untouched here).
- `verify:openapi` all checks passed; `detect:breaking` reports no changes against the baseline.
- Targeted suites re-run after review edits: 130 CLI, 855 core + web unit, 972 API integration — 0 failures.
- New coverage: author-direction rejection for both retired keys (unit, route `PUT`, and `/import` ZIP); a **stored published manifest carrying `dependencies.tools` still runs**, with a non-vacuity guard asserting the author direction still refuses the same bytes; bundle import succeeds, returns the warning, and leaves the stored draft byte-identical; `dependencies._meta` and the three canonical maps still accepted, proving the object was not closed; and for the CLI, the 307 papercut end-to-end plus the `-s`, `-I` and `-L` paths and stdout purity.

### DIFF
--- a/apps/api/src/openapi/parameters.ts
+++ b/apps/api/src/openapi/parameters.ts
@@ -108,7 +108,10 @@ export const parameters = {
     in: "query" as const,
     required: false,
     description:
-      "When `true`, narrows the list to packages installed and enabled in the current application.",
+      "When `true`, narrows the list to packages installed and enabled in the current " +
+      "application — system packages with no install row drop out. Integrations are the one " +
+      "exception: they are filtered on effective activation, so an environment-provided " +
+      "system integration stays listed even though it has no install row.",
     schema: { type: "string", enum: ["true"] as const },
   },
 } as const;
--- a/apps/api/src/openapi/paths/packages.ts
+++ b/apps/api/src/openapi/paths/packages.ts
@@ -422,7 +422,11 @@ export const packagesPaths = {
       operationId: "listSkills",
       tags: ["Packages"],
       summary: "List skills",
-      description: "List all skills (system + org) in the organization.",
+      description:
+        "List the skills available to the current application (`X-Application-Id`): system " +
+        "packages, plus organization packages installed in this application. Organization " +
+        "packages that exist but are not installed here are NOT returned — for the " +
+        "organization-wide catalogue with per-application install state, use `GET /api/library`.",
       parameters: [
         { $ref: "#/components/parameters/XOrgId" },
         { $ref: "#/components/parameters/XAppId" },
@@ -887,7 +891,12 @@ export const packagesPaths = {
       operationId: "listAgentPackages",
       tags: ["Packages"],
       summary: "List agent packages",
-      description: "List all agent packages (system + org) in the organization.",
+      description:
+        "List the agent packages available to the current application (`X-Application-Id`): " +
+        "system packages, plus organization packages installed in this application. " +
+        "Organization packages that exist but are not installed here are NOT returned — for " +
+        "the organization-wide catalogue with per-application install state, use " +
+        "`GET /api/library`.",
       parameters: [
         { $ref: "#/components/parameters/XOrgId" },
         { $ref: "#/components/parameters/XAppId" },
@@ -1665,7 +1674,12 @@ export const packagesPaths = {
       operationId: "listIntegrationPackages",
       tags: ["Packages"],
       summary: "List integration packages",
-      description: "List all integration packages (system + org) in the organization.",
+      description:
+        "List the integration packages available to the current application " +
+        "(`X-Application-Id`): system packages, plus organization packages installed in this " +
+        "application. Organization packages that exist but are not installed here are NOT " +
+        "returned — for the organization-wide catalogue with per-application install state, " +
+        "use `GET /api/library`.",
       parameters: [
         { $ref: "#/components/parameters/XOrgId" },
         { $ref: "#/components/parameters/XAppId" },
@@ -2245,7 +2259,12 @@ export const packagesPaths = {
       operationId: "listMcpServerPackages",
       tags: ["Packages"],
       summary: "List MCP-server packages",
-      description: "List all MCP-server packages (system + org) in the organization.",
+      description:
+        "List the MCP-server packages available to the current application " +
+        "(`X-Application-Id`): system packages, plus organization packages installed in this " +
+        "application. Organization packages that exist but are not installed here are NOT " +
+        "returned — for the organization-wide catalogue with per-application install state, " +
+        "use `GET /api/library`.",
       parameters: [
         { $ref: "#/components/parameters/XOrgId" },
         { $ref: "#/components/parameters/XAppId" },
--- a/apps/api/src/openapi/paths/packages.ts
+++ b/apps/api/src/openapi/paths/packages.ts
@@ -1,5 +1,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
+/**
+ * Shared tail of the four per-type list descriptions (skills / agents /
+ * integrations / mcp-servers) — only the leading noun differs.
+ */
+const listPackagesSharedDescription =
+  "system packages, plus organization packages installed in this application. " +
+  "Organization packages that exist but are not installed here are NOT returned — for " +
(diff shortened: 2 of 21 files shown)


pierrecabriere:
Ran an adversarial minimality audit on this branch (DRY / KISS / YAGNI / indirection / debt). Verdict was that ~95% of the diff earns its place — of 906 lines, ~480 are tests and the actual behaviour change is 30 lines of logic — but it found ~30 collapsible lines, now removed in a9147450d. No behaviour change; `detect:breaking` still reports no changes and `verify:api-types` still passes, so no baseline or generated-type churn.

What was cut:

- **`dropRetiredDependencyKeys` returned a `dropped` field nobody read.** It copied the shape of `dropRetiredRuntimeTools` without copying the reason — that one's `dropped` *is* consumed (`package-fork.ts`, `package-versions.ts`). Returns the manifest directly now, removing a speculative field from published `@appstrate/core` surface.
- **`RETIRED_DEPENDENCY_KEYS` was exported with zero consumers** outside its own module, plus a test that just restated the source literal while the two tests below it already pinned the same mapping behaviourally.
- **The four list-endpoint descriptions repeated the same tail verbatim.** Extracted, following the existing `proxySharedDescription` precedent in `credential-proxy.ts`.
- **Rationale restated four times inside one block** in `dependencies.ts`, and twice in `validation.ts` forty lines apart. Cross-file repetition kept — different audiences.

Probed and deliberately kept: the single direction flag (two co-varying flags is exactly how this regressed the first time), `REDIRECT_STATUS` and the CLI comment (each documents a tested decision), `collectRetiredDependencyKeyWarnings` (third instance of an established sibling pattern), and the `runs-remote-registry` test (its non-vacuity guard is what makes it non-redundant).

**One piece of debt knowingly deferred:** the option is still named `retiredRuntimeTools` while now governing both retired kinds. My original estimate of the rename cost was wrong — it is six call sites, all in `apps/api`, zero in `cloud/` or `registry/`. Still not worth cutting a core major for on its own, but it should ride the next core major rather than becoming permanent. The JSDoc says plainly that the name is historical.

PR #1092, opened 2026-08-04
fix(api): make Docker orchestration safe for rolling deploys

pierrecabriere:
## Summary

- preserve running managed containers during boot cleanup so a new API replica cannot remove workloads owned by the live replica
- keep reaping terminal containers after stale runs are stopped
- prefer the configured stable `PLATFORM_API_URL` before ephemeral Docker replica discovery
- expose a monotonic `starting -> ready -> draining` readiness lifecycle
- withdraw `/health` with `SIGURG` before graceful shutdown while continuing requests already routed during load-balancer propagation
- document all emitted readiness states in OpenAPI without breaking the existing 200-response enum, and regenerate the web API types
- add unit, integration, container, and rolling-deployment regression coverage

## Why

A Compose rolling deployment temporarily runs two API replicas against the same Docker daemon. The previous generic orphan sweep force-removed every managed container, including workloads still served by the old replica. Sidecars could also bind to a short-lived container alias instead of the stable Compose service name.

The old replica must also leave the proxy healthy set before it receives `SIGTERM`. `SIGURG` is used for that explicit drain transition because older Linux images ignore it by default, which keeps the first rollout backward compatible.

## Validation

- cleanup and platform URL unit regressions: 11 passed
- Docker cleanup integration regressions: 3 passed
- health/readiness/OpenAPI targeted suites: 18 passed
- OpenAPI structural validation and generated API type verification passed
- breaking-change detection: 0 breaking changes
- production platform container E2E: healthy orchestrator reports Docker healthy; unavailable orchestrator reports Docker unhealthy
- rolling E2E under continuous load: no request errors, maximum 2 application replicas
- deliberately unhealthy replacement rejected; previous healthy replica remained active

### DIFF
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -35,7 +35,7 @@ import { createUploadsRouter, createUploadContentRouter } from "./routes/uploads
 import { createDocumentsRouter, createDocumentPreviewRouter } from "./routes/documents.ts";
 import { createAdminStorageDeletionRouter } from "./routes/admin-storage-deletion.ts";
 import { createSpaFallbackHandler } from "./routes/spa.ts";
-import healthRouter, { bootGate, markServerReady } from "./routes/health.ts";
+import healthRouter, { bootGate, markServerDraining, markServerReady } from "./routes/health.ts";
 import { createIntegrationsRouter } from "./routes/integrations.ts";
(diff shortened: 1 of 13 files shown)


pierrecabriere:
Closing by design: we are dropping the bespoke single-stack rolling-deployment approach in favor of a simpler operational model with a short maintenance window. No code from this PR should be merged.

PR #1114, opened 2026-08-05
test(auth): measure the session cookie cache + add the prod evidence query (audit P3)

pierrecabriere:
The P3 items of the 2026-08-05 performance audit were deferred out of #1113 because none of them could be decided without evidence. This produces the evidence for all three — it does **not** flip any behaviour.

## 1. Better Auth `cookieCache` — measured, not assumed

The cache was hardcoded off to work around upstream issue #7607 (an expired `session_data` reportedly failing to regenerate from a still-valid `session_token`, logging users out). That issue is now closed **"not planned"** — no fix was shipped, so its closure is not evidence that anything changed. Both the old "keep it off, the bug may exist" and the audit's "the issue is closed, so enable it" rest on the same missing measurement.

`AUTH_SESSION_COOKIE_CACHE_SECONDS` (default `0` = off) turns the cache into a deployment knob and lets the test drive both modes through the existing env-then-`_rebuildAuthForTesting()` seam. A TTL rather than a boolean, because the number *is* the security exposure.

`apps/api/test/integration/auth/session-cookie-cache.test.ts` — 10 tests, real sign-in, full cookie jar (the fast-path helper crafts a `session_token` directly, which would bypass the cache and make every assertion vacuous):

| Scenario | Result |
|---|---|
| Flag off → no `session_data` cookie; on → cookie present with the configured `Max-Age` | as configured |
| Authenticated request from the cached session | 200 |
| **`session_data` expired, `session_token` still valid** (the #7607 scenario) | **200 — does not reproduce** |
| Session `realm` (platform-realm guard) across a cache round-trip | preserved |
| Session `createdAt` (the `freshAge` gate) after cache regeneration | preserved, stale stays stale |
| Out-of-band revocation (row deleted) inside the TTL | **still 200** |
| Same, after the TTL | 401 |
| **Replayed pre-logout cookie jar** inside the TTL | **still 200** |
| Same, cache off | 401 immediately |

Two findings worth reading twice:

- **#7607 does not reproduce** on the installed version with this repo's plugin set. The stated reason for keeping the cache off no longer holds.
- **Sign-out does not revoke a captured jar.** `session_data` is a self-contained signed assertion; sign-out clears it for the client that made the request but cannot reach a copy someone else holds. So with the cache on, "log out" and "log out everywhere" stop being immediate — bounded by the TTL. I had written this test expecting a 401; the measurement said 200, and the assertion now records what actually happens.

**The default stays `0`.** Enabling it is a deployment decision with a measured price, and this PR does not make it. The benefit side (one session SELECT per authenticated request) still needs prod DB metrics to weigh against that window.

## 2 & 3. `scripts/perf-evidence.sql`

Read-only (catalog reads + a bounded 10k-row sample), safe on a live primary. Run it on prod and the other two items become decidable:

- **Run-list DTO** — `pg_column_size` per JSONB column at p50/p95/max, what one 20-row page carries, how much of it is droppable, and the >100 kB tail. A slimmer list DTO breaks the public contract, so the payload has to justify a versioned DTO first.
- **Prefix indexes** — candidates derived from `pg_index` rather than a hand-written list, with usage, size, the covering index, per-table write amplification, and the `stats_reset` window (an `idx_scan` of 0 proves nothing without it). Unique and partial indexes are excluded: they are never redundant on column-prefix grounds.

Two notes on the detector: the array cast of `indkey` (an `int2vector`) is **zero-based**, so the obvious 1-based slice silently compares the wrong columns and reports a clean bill of health — it goes through the text form instead. Validated locally: 22 candidates, including `idx_run_logs_run_id` covered by `idx_run_logs_lookup` — the very index Postgres was observed *preferring* over its wider composite. Which is the point: covered is not useless, and nothing should be dropped without `EXPLAIN (ANALYZE, BUFFERS)` on the real queries.

## Verification

`bun run check` green. Full API suite: 3761 pass / 0 fail.


pierrecabriere:
## Production run — the two deferred items are now decidable

Ran `scripts/perf-evidence.sql` against prod (read-only). The script needed two fixes first, both of which only showed up against a real database (`b28eced`): a `relname` join that Postgres rejected as ambiguous, and a `stats_reset` that is NULL whenever the counters were never explicitly reset — i.e. usually, and precisely when the window matters.

**Observation window**: `pg_stat_bgwriter.stats_reset = 2026-03-21` → **4.5 months** of counters. Not the 1.6 days of postmaster uptime — the counters survived the restarts. That is what makes a `0` meaningful below.

**Scale for context**: 3,953 runs, 108,583 run_logs, 3,121 llm_usage rows; largest index 10 MB, largest heap 52 MB.

### 1. Slimmer run-list DTO — NO

Wire-shape measurement (`row_to_json(...)::text`, uncompressed, so this is what serialization actually produces):

| | avg | p50 | p95 |
|---|---:|---:|---:|
| whole run row | 2,468 B | 1,669 B | 7,057 B |
| droppable (`input` + `result` + `checkpoint` + `context_snapshot`) | 788 B | — | 4,548 B |

That is **32% of an average row**, ~15 kB out of a ~48 kB 20-row page. After HTTP gzip (JSON compresses 4–6×) the real saving is a couple of kB per page request.

Against that: a public contract break, a versioned list DTO, and two shapes to keep in sync forever. **Not worth it.** Revisit if p95 `result` grows an order of magnitude or list pages get much larger — the query is in the repo, rerun it then.

### 2. Prefix indexes — NO, and the original recommendation is refuted

The three largest "redundant prefix" indexes are **the most-used indexes in the database**:

| index | size | its scans | covering index | covering scans |
|---|---:|---:|---|---:|
| `idx_runs_package_id` | 120 kB | **153,838** | `idx_runs_package_run_number` (328 kB) | 55,089 |
| `idx_runs_application_id` | 112 kB | **56,955** | `idx_runs_app_status_started` (544 kB) | 58,377 |
| `idx_run_logs_run_id` | 1,344 kB | **2,745** | `idx_run_logs_lookup` (10 MB) | 1,718 |

Dropping these pushes those lookups onto indexes 3–8× larger. "Functionally covered" was never the same claim as "unused", and the planner agrees — this is the measured version of the local `EXPLAIN` that first flagged `idx_run_logs_run_id`.

Genuinely unused and covered, over the full 4.5 months:

- `idx_notifications_run` — 160 kB, **0 scans**, while its covering index took 74,186
- `idx_llm_usage_user_id` — 72 kB, 0 scans
- `idx_llm_usage_api_key_id` — 64 kB, 0 scans
- `idx_runs_api_key_id` — 16 kB, 0 scans

Total recoverable: **~310 kB**, on a database whose largest single index is 10 MB. Write amplification saved: those tables took 2,083 and 27,589 writes over 4.5 months — roughly 0.002 index insertions per second. Below the noise floor.

**Recommendation: no migration.** The candidates are real but the gain is kilobytes; revisit when `llm_usage` is two orders of magnitude bigger, at which point `idx_llm_usage_user_id` / `_api_key_id` are the first two to go.

### Net

All three P3 items resolve to "measured, not worth doing now" — but each is now a number rather than a hunch, and the query that produced it is in the repo. The `cookieCache` half of this PR stands as-is: the flag exists, defaults to off, and the matrix documents what turning it on costs.

PR #1118, opened 2026-08-05
feat: read-only package file explorer + rendered AFPS manifest view

pierrecabriere:
Implements `claudedocs/file-explorer-plan-2026-08-05.md`, then follows where it led.

A package detail page showed one file — the type's primary file — in a bare `<pre>`, and its manifest only as raw JSON in a modal. It now shows **every file in the artifact**, and renders the manifest as a readable view instead of dumping it.

## API — two generic read-only endpoints

Registered at the router root, so there is no per-type duplication:

```
GET /api/packages/{scope}/{name}/files?version=
  200 { entries: [{ path, size, media_kind, inline? }] }

GET /api/packages/{scope}/{name}/files/content?path=&version=
  200 raw bytes
```

The index is a flat, path-sorted list of real files; the client builds the tree. Small UTF-8 files carry their full text inline so the common case is one round-trip, and the content route serves both preview and download for everything else. A file is never truncated and presented as complete — either fully inline, or fetched whole.

Measured on the fattest real archive on hand (98 files, 2.7 MB zipped, 54 fonts making up 94 % of the raw bytes): the index costs ~114 KB instead of 2.7 MB, and the fonts move only if someone clicks download.

Both routes go through a single choke point, `readPackageSnapshot()`, so a later snapshot cache or byte-range reader lands in one place. **The cache is deliberately not here**: `unzipPackageArchive()` records compressed bytes, decompressed bytes, and the exact inflate duration; `readPackageSnapshot()` records final snapshot bytes and file count. This keeps decompression cost separate from storage and integrity-check latency.

### Behaviour worth knowing

- **Draft reads overlay the DB draft columns on top of the stored ZIP**, because the editor writes the row before re-uploading the artifact. Which entry `draft_content` is authoritative over is per type and mirrors `parsePackageZip`: `prompt.md` for agents, `SKILL.md` for skills, `INTEGRATION.md` for integrations — and only when the bundle actually ships one, since otherwise the column is a copy of the manifest rather than a file — and nothing for mcp-servers.
- **Version reads are the pinned bytes**, integrity-verified through the same SRI gate the download route uses, with no overlay.
- The validator is resolved before any bytes are fetched, so revalidating an exact version costs one query rather than a storage read plus a decompression.
- `Cache-Control` is always `private`: drafts, ranges, dist-tags, and yanked versions use `no-cache`; an exact non-yanked version gets only `max-age=300`, because version numbers can be deleted and republished. Content responses are always `application/octet-stream` + `nosniff` + `attachment`.
- Access combines `hasPackageAccess` with the resolved package type's `*:read` permission, before validator or ETag handling, and also hides ephemeral inline-run shadow packages.

The 1 MiB inline/preview ceiling is one constant, `PACKAGE_FILE_INLINE_MAX_BYTES` in `@appstrate/core/package-files`, imported by both ends. It was two declarations kept in manual lockstep by a comment; a drift would have meant the UI offering a preview the index never carries `inline` for. The module holds that constant and imports nothing, since the SPA bundles it into the browser.

## UI — one artifact tab, one manifest view

**Package AFPS tab** (the former content tab, id kept so deep links still land): the file tree. Generic across all four package types. It previously existed only for non-agents, which left an agent's `prompt.md` reachable solely through the actions dropdown's companion-file item — removed here, so agents gain the tab rather than losing the affordance.

**Overview tab**: a rendered view of the manifest. This started as "build a product view per type" and shrank once the schemas were read: all four AFPS types share the same ~21 metadata fields, so it is **one generic component**, not four. Only the genuinely type-specific tail differs — an integration's source and auths, an mcp-server's server and tools. Skills have no tail at all; their schema declares no field the shared block does not already cover.

The same component is mounted in the integration page's existing About tab, replacing its bespoke `MetadataBlock`. Integrations do not route through the unified detail page — they have their own — so one component, two mount points.

The manifest modal is gone. It served the raw object, now reachable two better ways: rendered in Overview, and as `manifest.json` in the Package AFPS tab, where the explorer applies the draft overlay the modal did not.

Package bytes are author-controlled and render as **source only** - HTML, SVG, Markdown and JSON included. Manifest fields render as JSX text; the only link path accepts an `http(s)` URL through `normalizeHttpUrl`. No package byte is injected, framed, or blob-rendered.

Tree state and manifest narrowing stay in `lib/package-file-tree.ts` and `lib/package-manifest.ts` as pure functions. Component render tests cover their composition without requiring a browser DOM.

## Review

Each part was implemented, then reviewed **blind** by an independent pass that was not told what the author intended. Follow-up passes found and fixed the defects below:

- `?version=latest` was served `immutable` for a year, so publishing a new version never reached the client.
- The ETag identified the snapshot rather than the representation, so a validator obtained from the index could be replayed against a path the client had never fetched and get a bodyless `304`.
- Neither route sent `Vary`, so switching application re-issued an identical URL and the browser answered from cache — showing an artifact the new application never installed.
- Source-only rendering was initially protected by a 474-line raw-source/import scanner. It is now covered at the actual boundaries: a narrow guard pins Monaco and inert download blobs, while behavior tests prove manifest URL rejection.
- Tree row identity is a file's kind and path together (an artifact may hold both a file `docs` and a file `docs/a.md`); only the React key accounted for that, so two rows claimed `tabIndex=0`.

## DRY / KISS / YAGNI cleanup

The final cleanup removes 1,325 net lines: the generic tab-overflow framework became local horizontal scrolling on its two call sites; dead fields and one-line forwarding wrappers were removed; cache naming now describes short freshness rather than immutability; and the 474-line import allowlist scanner became focused security assertions.

## Pre-existing issues found and deliberately NOT fixed here

1. `ROUTE_CONFIGS.integration.storageFileName` returns `"manifest.json"`, while an integration's `draft_content` holds `INTEGRATION.md`. The package UPDATE path therefore writes INTEGRATION.md markdown into the stored ZIP's `manifest.json`. Touching it would change the create/update write path for every package type.
2. `packages/core/src/validation.ts` accepts `repository` as any non-empty string with no URL shape check, and `setup_guide` is absent from the manifest schema entirely. Server-side validation there would be defence in depth but is a breaking AFPS schema change.

## Relationship to #1122 (merged)

Replacing `MetadataBlock` with the generic overview closed an XSS on the way: the integration page linked the manifest's `repository` with no protocol check, so a published integration carrying `"repository": "javascript:…"` executed on the platform origin when clicked. #1122 fixed that same defect on `main` independently, and also covered a **second** instance this PR does not touch — `setup_guide.steps[].url`, a field with no server-side schema at all.

This branch is rebased onto it. The `integration-detail.tsx` conflict was resolved in favour of `main`: the incoming side still held the pre-#1122 `SetupGuideSteps`, so taking it would have reintroduced the vulnerability #1122 had just removed. The manifest view's own link builder now delegates to `normalizeHttpUrl` from `@appstrate/core/url` rather than carrying a second copy of the check, and render/reader tests cover both accepted `http(s)` values and rejected protocols.

## Tests

- Focused file-explorer and RBAC suites: 209 passing.
- `bun run check` passes locally and in the pre-push hook.
- A broad local run completed 10,242 tests; its 19 failures are outside this diff (18 MITM fixtures tied to local CA/tooling plus one unrelated retry-worker timeout). GitHub CI remains the merge oracle.


pierrecabriere:
Heads-up from #1123 / #1124: the RBAC gap this PR deliberately did not widen is now fixed on `main` (#1124) — every `GET` on the packages router carries `requirePermission(<type>, "read")`, and `/{version}/download` additionally gained the `hasPackageAccess` call it was missing.

The two routes introduced here (`/{scope}/{name}/files` and `/files/content`) are the only ones left unguarded, since they don't exist on `main` yet. After rebasing, one line each right after `loadFileExplorerPackage` closes it:

```ts
const pkg = await loadFileExplorerPackage(c);
await requirePackageReadPermission(c, pkg.type);
```

`requirePackageReadPermission` is already in `routes/packages.ts` on `main` (resolves the resource from the row's `type` via `READ_GUARD_BY_TYPE`, reuses the same guard so the denial audit + 403 shape stay identical). The OpenAPI operations for both routes will need a `"403": { $ref: "#/components/responses/Forbidden" }`.

PR #1121, opened 2026-08-06
feat(system-packages): ship the 14 method skills as visible system packages

otarbes:
The "default agent fleet" strategy is a library of **connector-agnostic method skills** that the chat copilot assembles into agents on demand: an agent is `method skill × real connector × trigger`. Rather than pre-wiring one agent per integration (which multiplies with every CRM or helpdesk a customer happens to use), the method lives once and the copilot resolves the connector against what that user actually has connected.

The copilot ships with the product and names these 14 skills in its recipe table. The skills themselves did not ship: they had to be imported by hand into every instance. On a fresh install they simply do not exist, so an agent the copilot generates fails validation:

```
dependencies.skills.@appstrate/crm-update:
  Required skill '@appstrate/crm-update' is not installed
```

The copilot was pointing at an empty shelf.

## What this does

The 14 become ordinary system packages, synced at boot like the 65 integrations, and **listed** (no `unlisted` `_meta`) so a user can read the method their agent relies on. Scope moves from `@default/*` to `@appstrate/*`, consistent with every other first-party package.

| Skill | Does what |
|---|---|
| `minutes-actions` | transcript → summary, decisions, assigned actions |
| `data-analysis` | tabular data → indicators, trends, recommendations |
| `sourced-rag` | cited answers over an internal knowledge base |
| `email-reply` | inbox triage + reply drafts, never auto-sent |
| `incremental-digest` | periodic digest of what changed since the last checkpoint |
| `meeting-prep` | pre-meeting brief (attendees, history, context) |
| `customer-research` | 360° account / prospect view |
| `triage-sentiment` | ticket classification (urgency, category, sentiment) + escalation |
| `doc-extraction` | structured fields out of invoices, forms, contracts |
| `sprint-report` | progress report from tickets and PRs |
| `code-review` | review structured by severity |
| `sourced-research` | sourced web research and competitive watch |
| `crm-update` | lead enrichment, interaction logging, stalled-deal detection |
| `content-writing` | content in the organization's voice |

Visible does not mean editable: a system package is immutable (403). An organization that wants to bend a method **forks** it, which the platform already supports (`forkPackage`), and keeps receiving upstream fixes on everything it did not fork. That is deliberately the opposite of cloning the library into every org on generation, which would freeze each copy at clone time and strand every future correction.

Independent of #851: these are listed packages and need none of the `unlisted` machinery. Zero file overlap with that branch.

## Verification

End to end on a clean instance, current `main`:

- **Distribution** — 14 packages listed, `source: system`; the three chat skills stay unlisted and absent from the catalogue.
- **Generation and save** — the copilot generates *and saves* an agent declaring `@appstrate/data-analysis`, on both frontier engines: Claude Opus 5 (picks `createAgent` from the operation index) and GPT-5.6 Sol (reaches it via `search_operations`, since Codex trims that index). `createAgent` 201 in both cases.
- **Real run** (`RUN_ADAPTER=docker`) of the generated agent over a pasted sales CSV:

  ```json
  {"tool":"read","args":{"path":"/workspace/.pi/skills/@appstrate/data-analysis/SKILL.md"}}
  ```

  The agent opens the skill, computes with pandas, and returns the skill's exact output shape (`resume`, `indicateurs`, `tendances`, `recommandations`, `anomalies`). Every figure checks out against the input — total 8 238, 162 units, 686.50 average, March peak 2 527, Quebec 71.0 %.

  Most telling, the skill's anti-hallucination rules are visibly applied rather than merely present: the CSV declares no currency, so the agent reports "unités monétaires" instead of inventing one, and it flags that order counts are not derivable from the columns instead of approximating them.

`bun run check` 33/33. `bun run scripts/build-system-packages.ts --check`: 80 packages validated (65 integration, 1 mcp-server, 14 skill).

## Open question for review

This adds 14 packages to the shipped set — repo weight, build time, maintenance surface. That is a maintainer call and the reason this is its own PR rather than folded into #851. Happy to trim the initial set if you would rather start smaller; HR and legal method skills were already deferred for the same reason.


otarbes:
Closing: the design changed before anyone reviewed this.

Shipping 14 separate system packages splits the *recipes* (curated in the copilot's SKILL.md) from the *methods* (one package each). Those two documents drift — which is exactly what produced the defects fixed in the companion branch: two skills never referenced by the recipe table, four rows routing to the wrong method. Patching that by hand treats the symptom.

The replacement keeps one source of truth and moves ownership to the org:

- the recipes and the methods live together in the copilot package (`SKILL.md` as the index, `references/*.md` as the methods), so they cannot drift apart;
- when the copilot builds an agent, it materialises the one relevant method as a package in the user's own org via `createSkill`. The org owns it, edits it without hitting the system-package 403, and versions it on its own timeline;
- `forked_from` keeps the provenance, so an upstream correction can be *offered* as a diff rather than silently never arriving.

This also drops two problems this PR would have introduced: 14 system entries saturating the 15-slot caller-context hint budget (org-owned skills are truncated first, since the ordering puts `source = 'system'` ahead), and a catalogue full of generic methods the customer never asked for.

No review time was spent here. Reopening if the new approach does not hold up.

Issue #834, opened 2026-07-05
Docker orchestrator: cached egress network ID goes stale → all runs fail with "network not found"


pierrecabriere:
## Symptom

Every run fails at container start:

```
Docker start container failed: 404 {"message":"failed to set up container networking: network dceac411bc393b5d98a543cd0e92838db0de558aac620df23d089b7dcfcd0a6d not found"}
```

Sidecar/agent containers are left in `Created` state, still referencing the dead network ID. The only recovery is restarting the API.

## Root cause

`DockerOrchestrator` creates the shared `appstrate-egress` network once at boot and caches its **ID** in memory (`this.egressNetworkId`, `docker-orchestrator.ts:96-105`). Every sidecar is then created with `networkId: this.egressNetworkId!`.

If the network disappears while the API process is alive, the cached ID is never revalidated and every subsequent run fails with the 404 above. Three ways the network dies under a live API:

1. **Concurrent API process shutdown** — `shutdown()` (`docker-orchestrator.ts:108-110`) unconditionally removes the egress network. A second API instance (integration test run, stray `bun run dev`, preload-booted test app) shutting down deletes the network the first instance still uses.
2. **Concurrent API process boot** — `cleanupOrphanedNetworks()` (`docker.ts:681-684`) targets `appstrate-exec-*` **and** `appstrate-egress` by name. The comment says "only safe to call at boot", but boot of process B is mid-operation for process A.
3. **Manual `docker network prune`** or Docker daemon restart.

The dev setup hits case 1/2 routinely: running the integration test suite (which boots its own app) while `bun run dev` is up is enough to break the dev server silently.

## Suggested fix

- Make the egress network resolve-by-name at use time, or catch the 404 on container create, recreate `appstrate-egress`, refresh the cached ID, and retry once.
- Stop deleting `appstrate-egress` in `shutdown()` and in the boot sweep when it still has attached endpoints (Docker already refuses deletion of networks with live endpoints — but the current `.catch(() => {})` + name-based sweep combination still races on the empty-network window between runs). Treat it as durable infra, like the platform network.
- Also clean up stale `Created` run containers referencing missing networks in the boot sweep (currently they only match once the network 404 is fixed manually).

## Repro

1. `bun run dev` (RUN_ADAPTER=docker), trigger one run to force orchestrator init
2. In another shell: `docker network rm appstrate-egress` (simulates concurrent-instance shutdown)
3. Trigger a run → 404 network not found; API never self-heals

[...]

PR #1203, opened 2026-08-23
fix(orchestrator): never reclaim a workload whose ownership cannot be proven

### LINKED ISSUE #1130: fix(docker): boot cleanup must preserve active sibling workloads
## Problem

Starting one Appstrate API instance can kill active agent and sidecar containers owned by another instance on the same Docker daemon.

The boot sequence first performs the safe, database-backed pass in `apps/api/src/lib/boot.ts`: stale run IDs are selected from the current database and stopped through `orchestrator.stopByRunId(runId)`. Immediately afterwards, `DockerOrchestrator.cleanupOrphans()` calls `cleanupOrphanedContainers()`.

`cleanupOrphanedContainers()` currently lists every container carrying `appstrate.managed=true`, then force-removes all of them regardless of lifecycle state. The label is shared across Appstrate instances and worktrees, so it does not establish ownership.

When the active instance polls the removed agent through `waitForExit()`, Docker returns `404`. That path currently converts the missing container into exit code 137, so the run is reported as a `SIGKILL` even though the platform no longer has evidence of the container's real exit status.

This is related to #955, but distinct. Issue #955 covers intentional sidecar teardown after a successful run. This issue covers cross-instance boot cleanup removing another instance's active workloads.

## Observed incident

While a long agent run was active on one local instance, a second API instance started against the same Docker daemon. The agent and sidecar died together with code 137 at the second instance's boot time. Docker emitted no OOM event.

The same target later completed successfully while both already-started instances remained active:

- agent peak: 198.6 MiB of 1.5 GiB
- sidecar peak: 130.4 MiB of 256 MiB
- `OOMKilled=true` samples: 0

The dangerous event is the global cleanup during boot, not the steady-state presence of a second instance.

## Safety invariant

A global Docker sweep must never stop or remove a non-terminal workload whose ownership it cannot prove.

The database-backed orphan pass remains the authority for stopping active stale runs by ID. The daemon-wide cleanup pass should only reclaim inert Docker residue.

## Proposed correction

In `apps/api/src/services/docker.ts`:

1. Read the `State` returned by `GET /containers/json?all=true`.
2. Remove only containers whose state is `exited` or `dead`.
3. Preserve `running`, `created`, `restarting`, `paused`, and any unknown state.
4. Return the number of successful removals, not the number of listed managed containers.
5. When `waitForExit()` receives `404`, reject with an explicit infrastructure error such as `DockerContainerDisappearedError` instead of returning 137.

No change to the Core `RunOrchestrator` interface is required. Its existing `cleanupOrphans()` contract already says that it cleans resources left by a previous crash. The defect is local to the Docker adapter implementation.

## Verification already performed

A throwaway implementation was tested against the real Docker integration harness on commit `33cc5315`.

Added regression cases:

- `cleanupOrphanedContainers` removes a terminal managed container while preserving an active managed container
- `waitForExit` rejects instead of reporting `SIGKILL` when its container disappears during polling

Results:

```text
TEST_DOCKER=1 bun test apps/api/test/integration/services/docker-api.test.ts
38 pass
0 fail
73 expect() calls
```

`apps/api` typecheck also passes with the prototype.

## Tradeoff

A running container absent from the current instance's database will no longer be force-removed by the daemon-wide boot sweep. This is intentional: without an ownership signal, killing it risks terminating a valid sibling run. Runs known to the current database are already handled by the stale-heartbeat pass and watchdog. Any future stronger cleanup of foreign running residues requires an explicit ownership model, not the shared `appstrate.managed=true` label.

## Acceptance criteria

- Boot cleanup never removes managed containers in a non-terminal state.
- Terminal managed containers, orphan networks, and unused workspace volumes are still reclaimed.
- A Docker `404` during `waitForExit()` is reported as container disappearance, not exit code 137.
- Docker integration tests cover both regressions and pass with `TEST_DOCKER=1`.
- `apps/api` typecheck passes.
- A smoke test starts a run on instance A, boots instance B on the same Docker daemon, and confirms that the run on A remains alive.

pierrecabriere:
Closes #1130.

## The invariant

> A global sweep must never reclaim a non-terminal workload whose ownership it cannot prove.

`appstrate.managed=true` marks a per-run resource but identifies **no owner**: every Appstrate process on a host writes it. The boot sweep force-removed everything carrying it, so a second instance killed a sibling's live agent and sidecar mid-run — and the victim's `waitForExit` then reported the disappearance as exit code 137, sending the investigation after an OOM that never happened.

"Two instances" is not an exotic setup. It is **every redeploy** (Coolify health-gates the new container before draining the old one, same daemon, same socket), two worktrees on one machine, and integration tests running beside a dev server. The DB-backed pass two blocks above already knew this — `boot.ts` says *"`listOrphanRunIds` already excludes runs a live sibling instance heartbeats, so this never kills another instance's in-flight run"* — and the global sweep then undid that guarantee. Same class of bug as #834, which fixed it for the shared egress network.

The issue covers the Docker adapter. Auditing for the same invariant turned up **two more violations**, one of them in the default adapter.

## 1. Docker sweep — reclaim only what is provably finished

The reclaim allowlist is exactly what `docker container prune` itself reclaims, **minus** a TTL on `created`. Both halves were measured against a real daemon rather than assumed:

- `POST /stop` on a `created` container returns **304 and leaves it `created`** — so no stop-then-sweep sequence can ever make it terminal. Excluding `created` entirely would leak it forever, holding a network endpoint that in turn blocks the orphan-network reclaim.
- `docker container prune` removes `created` containers and preserves `running` ones.

So: `exited`/`dead` always; `created` past `RUN_BOOT_DEADLINE_SECONDS` (a sibling between `createWorkload` and `startWorkload` sits there for milliseconds — past the platform's own boot deadline the row is abandoned by definition); everything else — `running`, `paused`, `restarting`, `removing`, and any state a future Engine adds — preserved. Strictly more conservative than Docker's own reaper, never more permissive.

Runs this instance owns are unaffected: boot stops them by id first, which turns them into `exited` rows the sweep then reclaims.

## 2. `waitForExit` — a vanished container has no exit code

404 no longer becomes 137. It rejects with `DockerContainerDisappearedError`, which `execute-background` turns into a `failed` terminal carrying the real cause. `SidecarExitWatcher` consumes its teardown expectation on that path too, so a removal **we** asked for is not reported as a watcher error merely because it surfaced as a disappearance instead of an exit code.

## 3. Process adapter — the same fault, in the default adapter

`RUN_ADAPTER=process` is the default, and its sweep probed `kill(pid, 0)`, saw the workload was **alive**, and SIGKILLed it for that exact reason. Its workspace reap was worse: `os.tmpdir()` is shared regardless of working directory, so booting one worktree `rm -rf`'d a running agent's `/workspace` in another. The uid filter it relied on does not separate two instances run by the same user.

Both now consult an `.appstrate-owner` marker holding the creating platform's pid. Pids are global, so it answers from any cwd — which matters because `DATA_DIR` is not, and a sibling worktree's pidfiles are invisible from here. Pid reuse can only make a dead owner look alive: it defers a reclaim, it never causes a wrongful kill. A crashed platform's own residue still fails the check, so orphaned live subprocesses are still reaped — that path is deliberately preserved and covered by a test.

## Also fixed

The sweep reported the number of containers **listed** rather than removed, claiming success when every removal failed. Three duplicated list-then-act blocks are now one `listContainers`/`removeContainers` pair, mirroring the `removeNetworksMatching`/`removeVolumesMatching` helpers that already existed for the other two resource types — the duplication is where the bug lived. The partition runs in one pass against one clock reading; evaluating the predicate twice let a `created` container crossing the deadline between reads fall out of both halves.

## Verification

Every regression test was **positive-controlled** — reverted to the old behaviour to confirm it fails, then restored. A test that cannot fail proves nothing.

| Check | Result |
|---|---|
| `TEST_DOCKER=1` docker-api | 43 pass / 0 fail |
| Adjacent suites (watchdog, launchers, egress) | 36 pass / 0 fail |
| Positive control, Docker (3 regressions) | **3 fail** on old code |
| Positive control, process (2 regressions) | **2 fail** on old code |
| `tsc --noEmit`, eslint, knip (run directly, no turbo cache) | clean |

The existing test `cleans up labeled containers and networks` started two `sleep 60` containers and asserted they were removed — it **encoded the bug**, so it is rewritten to use terminal containers.

End-to-end smoke test of the acceptance criterion (instance A holds a live agent + sidecar, instance B runs its real boot sweep):

```
instance B boot sweep → {"containers":1,"networks":0,"volumes":0}
A's live workloads surviving: 2/2 | crashed-run residue reclaimed: true
```

## Operational note

A run whose container disappears now fails with the disappearance message instead of `exited with code 137`. That is the point — but any runbook or alert matching the string `exited with code 137` needs updating.


pierrecabriere:
### Note sur `codecov/patch`

Le chiffre de `apps/api/src/services/docker.ts` (10,5 %) est un artefact de mesure, pas une absence de tests.

`apps/api/test/unit/docker-image-heal.test.ts:73` importe volontairement une **seconde copie du module** pour isoler son faux démon :

```ts
const specifier = "../../src/services/docker.ts?fake-daemon";
docker = (await import(specifier)) as DockerModule;
```

Le lcov n'émet qu'un seul bloc `SF:apps/api/src/services/docker.ts`, et les compteurs des deux copies ne fusionnent pas — c'est la copie `?fake-daemon`, qui n'exécute jamais le balayage, qui est rapportée.

Mesuré localement sur les lignes ajoutées par cette PR :

| Exécution | Lignes du balayage couvertes |
|---|---|
| `docker-orphan-sweep.test.ts` seul | **45/46** |
| suite `apps/api/test/unit/` complète | 4/64 |

Les tests passent dans les deux cas — seule l'attribution change. Les trois autres fichiers touchés sont à 100 % de couverture des lignes ajoutées (`docker-errors.ts` 8/8, `process-orchestrator.ts` 25/25, `sidecar-exit-watcher.ts` 6/6).

Les cas contre un vrai démon restent dans `docker-api.test.ts`, derrière `describeRequiresDocker` — aucun workflow ne pose `TEST_DOCKER`, ils ne tournent donc qu'en local (43 pass, avec contrôle positif : chaque régression a été vérifiée comme échouant sur le code d'avant).

PR #1222, opened 2026-08-26
fix(files)!: accept the legacy `doc_` id — the rename never touched files.id

pierrecabriere:
**Production regression, live since `v1.0.0-beta.53`: every stored file 404s on preview and download.** Reported from the UI, reproduced against the live database.

## The path

`loadFileForPreview` and `resolveFileForActor` (`apps/api/src/services/files.ts:1220,1285`) test `FILE_ID_RE` **before any SELECT**:

```ts
if (!FILE_ID_RE.test(fileId)) return null;   // → 404, without touching the DB
```

`FILE_ID_RE` accepted only `file_`.

## Why that was safe, and why it wasn't

`file-uri.ts` asserted the row id prefix *"was `doc_` until the rename was finished at the physical layer"*.

It never was. `0043` renamed the **table** (`ALTER TABLE documents RENAME TO files`), `0044` rewrote **`storage_key`** — neither touches `files.id`, and no other migration does. Measured on production after `0044` shipped:

| | |
|---|---|
| `files` with id `doc_` | **521** |
| `files` with id `file_` | **0** |
| `file_links` → `doc_` | 25 |

The rename was finished everywhere except the one place this validator reads.

## Why widen instead of migrating the ids

The id is opaque; its prefix is read by nothing but this validator. Rewriting it means 521 ids, 25 foreign keys, 521 `storage_key` values that **embed the id in their path**, and moving 521 storage objects a second time. The one thing that could have forced a rewrite is a persisted `appfile://doc_…` reference — measured: **0 rows** in `runs.input`.

New rows still mint `file_` via `prefixedId("file")`. `doc_` is read-only history.

## The sibling a narrow fix would have left open

`EMBEDDED_FILE_URI_SCAN` matched only `appfile://file_…`. But `fileUri()` is a bare concatenation, so a legacy row yields `appfile://doc_…`, and `inline-run.ts` / `files.ts` / `state/runs.ts` all read "not found" as "not referenced" — a file silently dropped from a prompt, no error. No such row exists today; attaching any existing file to a run mints one. Widened too.

## Tests

Two existing assertions pinned the wrong behaviour and are corrected. The retired **`document://` scheme** is still refused; the **`doc_` id** it addressed is not — the split is now pinned in both directions.

- `bun run check` — 35/35, **`Cached: 0`** (nothing masked by the turbo cache)
- `packages/core` — 1099 pass / 0 fail
- `apps/api/test/unit` — 1628 pass / 0 fail
- **Negative control** — restoring the `file_`-only regex fails 2 of the new assertions, so the guard discriminates rather than merely passing.

### DIFF
--- a/packages/core/src/file-uri.ts
+++ b/packages/core/src/file-uri.ts
@@ -23,17 +23,19 @@
  * afterwards so historical `runs.input` rows, persisted chat attachments and
  * model-authored prompts stayed resolvable. That compatibility became
  * unreachable when the rename was finished at the physical layer: every URI
- * ever written under the old scheme addresses a `doc_` id, and
- * {@link FILE_ID_RE} accepts only `file_`. So the pair `document://` +
- * `file_xxx` — the only thing the accept-path could still have matched — is a
- * form nothing has ever emitted, and a genuine `document://doc_xxx` fails on
- * the id, not the scheme. One spelling is read and one is written, and they
- * are the same one.
+ * ever written under the old scheme addresses a `doc_` id. Only the SCHEME is
+ * unread: `document://` is never parsed, while the `doc_` ID it addressed is
+ * still perfectly live and {@link FILE_ID_RE} accepts it.
  *
- * The row id prefix is `file_`, minted by `prefixedId("file")`. It was `doc_`
- * until the rename was finished at the physical layer; nothing reads the old
- * spelling any more, which is exactly why the old scheme has nothing left to
- * address.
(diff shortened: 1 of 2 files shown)


pierrecabriere:
Approche rejetée : la plateforme ne doit accepter aucune forme legacy. Le correctif est une migration de données (`files.id` `doc_` → `file_`), pas un élargissement du validateur. Remplacée par la PR de migration.

PR #1223, opened 2026-08-26
fix(db)!: migrate every `doc_` file id to `file_` — finish the rename in the data

pierrecabriere:
**Production regression, live since `v1.0.0-beta.53`: every stored file 404s on preview and download.**

`FILE_ID_RE` accepts only `file_`, and `loadFileForPreview` / `resolveFileForActor` (`apps/api/src/services/files.ts:1220,1285`) test it **before any SELECT**:

```ts
if (!FILE_ID_RE.test(fileId)) return null;   // → 404, without touching the DB
```

`file-uri.ts` asserted the row id prefix *"was `doc_` until the rename was finished at the physical layer"*. It never was — `0043` renamed the **table**, `0044` rewrote **`storage_key`**, and nothing between `0000` and `0052` touches `files.id`. Production carried **521** rows with a `doc_` id and **0** with `file_`.

## The platform keeps accepting exactly one form

Widening the validator to admit `doc_` was written, then rejected: it makes a retired spelling permanently legal in a design whose whole point is that one spelling is written and the same one is read. **The data moves instead.**

## Four writes — measured on a restored copy of production

| target | rows | change |
|---|---|---|
| `files.id` | **521** | `doc_x` → `file_x` |
| `file_links.file_id` | **25** | both FKs dropped and recreated verbatim around it |
| `runs.input` | **64** | `document://doc_x` → `appfile://file_x` |
| `chat_messages.content` | **59** | bare `doc_x` → `file_x` |

Neither FK carries `ON UPDATE`, so the parent id cannot move while a child points at it — hence the drop/recreate. The unique index the composite FK binds to (`uq_files_id_org_id`) is untouched.

`runs.input` gets `appfile://`, not just a renamed id: those references were written when `document://` was still read (pre-#1177), so porting the spelling preserves what the row **meant** instead of freezing a form nothing can resolve.

## What the pattern deliberately misses

Writes 3 and 4 match `doc_` followed by a **strict UUID**. Afterwards **0** rows keep a full-UUID reference, and the 11 chat rows that still contain the string `doc_` are conversation text *discussing* the format:

```
"docId": "doc_123"          ← placeholder inside example JSON
`document://doc_xxx`        ← quoted in prose
{"documentId": "doc_..."}   ← truncated placeholder
```

Rewriting those would corrupt what a person actually wrote.

## What is deliberately NOT touched

**`files.storage_key` keeps its `doc_` path segment, and no storage object moves.** Nothing derives the key from the id and nothing parses the id back out of it — `parseStorageKey` (`services/files.ts:116`) splits `{bucket}/{path}` and returns the bucket only; the path is written once at creation and afterwards only read back from the row. Moving 521 objects a second time in one day is real risk for a string no reader interprets.

**`run_logs` keeps its `document://doc_…` prose** — the immutable text a run actually emitted, and nothing resolves a URI out of log prose.

## Rehearsal

`pg_dump` of live production → throwaway `postgres:16-alpine` → the migration in one transaction:

```
ALTER TABLE ×2   UPDATE 521   UPDATE 25   DO   UPDATE 64   UPDATE 59   COMMIT
```

Verified after: **0** `doc_` ids, **521** `file_`, **0** orphaned `file_links`, both foreign keys present, `0` rows with `document://doc_` remaining.

`bun run check` 35/35 with `Cached: 0`; `packages/core` + `packages/db` 1109 pass / 0 fail.

The schema is structurally unchanged (FKs recreated verbatim), so `0053_snapshot.json` is `0052`'s.

### DIFF
--- a/packages/db/drizzle/0053_file_ids_doc_to_file.sql
+++ b/packages/db/drizzle/0053_file_ids_doc_to_file.sql
@@ -1,0 +1,110 @@
+-- Rename every `doc_` row id to `file_`, and every reference to one.
+--
+-- ═══ WHY THIS EXISTS ═══
+--
+-- `FILE_ID_RE` (`@appstrate/core/file-uri`) accepts only `file_`, and
+-- `loadFileForPreview` / `resolveFileForActor`
+-- (`apps/api/src/services/files.ts:1220,1285`) test it BEFORE any SELECT and
+-- return null on a miss. A `doc_` id is therefore a 404 that never reaches the
+-- database. On v1.0.0-beta.53 that took out EVERY stored file on production at
+-- once: preview and download, all 521 of them.
+--
+-- The prose in `file-uri.ts` asserted the row id prefix "was `doc_` until the
+-- rename was finished at the physical layer". It never was. 0043 renamed the
+-- TABLE (`ALTER TABLE documents RENAME TO files`), 0044 rewrote `storage_key`
+-- — neither touches `files.id`, and no migration between 0000 and 0052 does.
+-- The rename was finished everywhere except the one place a validator reads.
+--
+-- Widening the validator to accept `doc_` was considered and REJECTED: it makes
+-- a retired spelling permanently legal in a design whose whole point is that one
+-- spelling is written and the same one is read. The data moves instead.
+--
+-- ═══ WHAT IS NOT TOUCHED, AND WHY ═══
+--
+-- `files.storage_key` keeps its `doc_` path segment. Nothing derives it from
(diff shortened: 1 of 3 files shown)


pierrecabriere:
Fermée : une migration de données n'a pas sa place dans les migrations drizzle. Le renommage a été appliqué à la prod par un script SQL ponctuel (521/25/64/59, vérifié). Les migrations drizzle décrivent le SCHÉMA ; la réparation de données va dans `scripts/migration/`, et aucun code de compatibilité antérieure n'est introduit. Voir la convention ajoutée au projet.

PR #1224, opened 2026-08-26
docs: forbid transitional code, add /audit-legacy, and its first run

pierrecabriere:
Closes the loop on today's two production incidents, both caused by the same rule being broken in opposite directions.

## The rule — `docs/NO_TRANSITIONAL_CODE.md`

1. **No backward-compatibility branches.** A validator accepts one form. If production data does not match, the data is wrong. A retired name must **fail loudly** — `RETIRED_ENV_RENAMES` is the correct shape and already exists.
2. **Data repair lives in `scripts/migration/`**, never in a drizzle migration.
3. **No dead transition scaffolding** — deleted in the PR that completes the transition.

Grounded in the two post-mortems: the data half of #1177 went into drizzle migrations and was then deleted (leaving every beta.51 database with neither half, silently under-granting API-key permissions — `0046` exists only to restore it), and the id half was never written at all (521 `doc_` ids against a `file_`-only validator → every stored file 404'd on beta.53).

Also adds `scripts/migration/` with its README and `0001-file-ids-doc-to-file.sql`, the script actually applied to production today.

## The audit — `/audit-legacy`

Six parallel opus sub-agents. First run included: **9 live violations**.

| finding | blast radius |
|---|---|
| `OAUTH_ALLOWED_INTERNAL_IDP_HOSTS` aliased at 3 sites | 🔴 **security** — SSRF allowlist |
| `github-action` reads `runId`, response has `id` | 🔴 breaks every run |
| `github-action` sends `config`, schema is `.strict()` | 🔴 400 on every run |
| pinned-SDK version branch + upload `s: 0` + `bundle` alias + a 0-reader `@deprecated` export | 🟠 clean deletions |

The security one was found **independently by two agents**, and its two readers disagree on the empty string compose produces sixteen times here.

Full report: `claudedocs/audit-legacy-2026-08-26.md`.

## The audit found a defect in itself

Its one false positive came from agents reading sibling repos' working trees rather than `origin/main` — `cloud` reported one major stale when main had been bumped hours earlier. The command now requires `git show origin/main:<path>`. A stale read presented as current state: the audit's own version of the failure it hunts.

### DIFF
--- a/.claude/commands/audit-legacy.md
+++ b/.claude/commands/audit-legacy.md
@@ -101,11 +101,20 @@ find config keys read under two names; find module-contract branches keyed on a
 version or capability that every module now has.
 
 **Agent F — Cross-repo & tests.** `cloud/`, `connect-helper/`, `github-action/`,
-plus `test/` fixtures across the workspace. Out-of-tree consumers are the ONE
-legitimate reason a shape survives its in-tree caller — so a finding here must
-state whether the consumer still needs it (check the consumer's source, do not
-assume). Also flag test fixtures that pin a retired shape, which is how a
-compatibility branch acquires a test that justifies keeping it.
+plus `test/` fixtures across the workspace.
+
+> **Read `git show origin/main:<path>` in sibling repos, never the working tree.**
+> A sibling checkout sits on whatever branch someone left it on, which can
+> predate main by weeks. The first run of this audit reported `cloud` pinned to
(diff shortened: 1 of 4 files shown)


pierrecabriere:
Superseded by the implementation PR, which contains these two commits — the convention had to be readable on that branch for the work to be checked against it.

PR #1229, opened 2026-08-27
Retire what the third audit found, and close the gaps that let it survive

pierrecabriere:
Third `/audit-legacy` pass. Six parallel sub-agents swept the workspace against `docs/NO_TRANSITIONAL_CODE.md`; this lands every in-tree finding, plus the durable half of each — the gate, register or record that would have caught it.

Report: `claudedocs/audit-legacy-2026-08-27.md` (gitignored, local).

One commit per phase, each self-contained.

## What was retired

| Commit | Finding |
| --- | --- |
| `58119fcf` | Three records describing a state the code had left: knip tags `knip.config.ts` already claimed were dropped, two module hooks that no longer exist, a test header contradicted by its own body |
| `22a5b756` | A dead `report` log channel, and an `isAgentText` fallback for runs predating the `assistant_message` stamp |
| `6914bafc` | The retired `source_code` key, stripped in silence on three package-JSON bodies |
| `7a6e4902` | `chat_messages.parent_id` and `.format` — two columns the schema's own comment called informationless |
| `d1d833da` | Three gaps in `verify:no-migration-dml` |
| `175af9c7` | Two fields declared optional that the server always sends |
| `2d27b225` + `49b7b61f` | A read-time repair of a bare `Authorization` auth-scheme prefix |

## Three things worth a reviewer's attention

**The audit was wrong about the `Bearer` finding, and checking saved it.** It called the repair dead on the strength of a code-side defaults table. The real producers are the manifests, and **51 in-repo integrations wrote the bare form** — every one in `Authorization` position. Deleting the repair alone would have sent `Authorization: BearerTOKEN` from 51 integrations at once. All 51 are fixed with a patch bump, directory rename and archive rebuild, since a published version is immutable and the fix would otherwise stay inert in production.

Chasing the last two fixtures then turned up a **second implementation of the same rule** in `scripts/conformance/auth-live.ts`, which was also a live bug: it turned the composite prefix `"Token token="` into `"Token token= tok"`, so the conformance probe had been sending a header shape no run emits.

**The data half is written, because the exposure is real.** System packages are unexposed — `resolvePublishedManifest` short-circuits on the in-memory registry the rebuilt archives replaced, and `integration_pins` pins a connection, not a version. But **org-imported and org-published integrations** store their manifest verbatim in `packages.draft_manifest` / `package_versions.manifest`, validated by the validator of their era and never revalidated on write. `scripts/migration/0005-afps-bare-auth-scheme-prefix.sql` moves them: its `WHERE` is exactly the condition it removes, every other auth key passes through untouched, and it was rehearsed twice against a PGlite fixture covering each shape. Counts are `UNMEASURED` per `0004`'s precedent — no production dump exists here, and the header carries the operator's own verify query.

**Two gates were blind in the direction that mattered.**
- `verify:no-migration-dml` could not see `EXECUTE format('UPDATE …')`, because `sanitize` blanks literals. Not hypothetical: `EXECUTE format(…)` is already how this directory writes catalog-guarded DDL, 23 times. It now reads command strings as the SQL they become, exempts `CREATE FUNCTION` bodies (while keeping `DO` blocks flagged), and covers `MERGE`.
- `verify:openapi` step 7 compared only *spec-optional + type-required* — the opposite direction from the `ResolvedRunConfig` violation. It now compares both, which found one more instance (`EnrichedRun.package_ephemeral`). Both are fixed rather than registered, so `KNOWN_REVERSE_DRIFT` ships **empty**, with its contract intact.

## Verification

- `bun run check` — **36/36 tasks**, type coverage 99.05%
- `TEST_TIER=0 bun test` — **11657 pass, 0 fail** across 861 files
- `verify:no-migration-dml` — 55 files, same 8 grandfathered names, no ninth. Regression control: `findDml` with grandfathering bypassed gives byte-identical results before and after, and all 23 live `EXECUTE format(…)` DDL sites stay clean
- `build:system-packages:check` — 66 packages validated under the new install gate
- Migration `0054` is DDL only; `DROP COLUMN` is metadata-only in Postgres

Both column drops were re-verified independently before anything was dropped, and the new ordering test was mutation-checked — flipping `loadMessages` to `created_at` fails it.

## Known limits, stated rather than glossed

- Three cases remain outside the dynamic-SQL reading (a concatenated command, an interpolated verb, and — in the safe direction — a `format(…)` building a message). All written into §2.
- 129 of 213 2xx JSON responses declare their schema inline; an inline response with no registry entry is compared in neither direction. `ResolvedRunConfig` sat in that blind spot for its whole life.
- `invalid_manifest` maps onto 404, so a stale pre-gate row presents as a missing integration rather than a bad prefix. Loud, but not self-explaining. A negative test pins it and names `0005` as the fix.

## Not in this PR

Three findings live in sibling repos and need their own PRs: `github-action` sends a `config` key that `runAgentBodySchema` has rejected since #1187 (broken today) and ships two AFPS 1.x example manifests that no longer validate; `cloud` adapts to a `BeforeUsageParams` shape four majors below its own declared floor.


pierrecabriere:
## Adversarial re-review — 7 defects found and fixed

Five reviewers re-read the full diff hunting for regressions rather than confirming it, across disjoint slices. Verdicts were SOUND WITH CONCERNS ×4 and **DEFECTIVE** ×1. Two follow-up commits fix everything found.

**The through-line:** the code was almost entirely correct. Six of seven defects were a *written claim nobody checked* — which is precisely what this PR exists to eliminate. It was reproducing its own target defect in its own artifacts.

### Behaviour — `16b8415e`

1. **A bare prefix could still reach the wire, silently.** The branch asserted a bare scheme "is refused at install time by the manifest validator". False for the hand-authored local creds file: `resolveLocalDeliveryPlan` feeds `entry.injection.headerPrefix` straight to the injector, never touching `integrationManifestSchema`. An `--integrations local --creds-file` user with `"headerPrefix": "Bearer"` would have started sending `Authorization: BearerTOKEN` — the exact silent failure this PR removes, on the one path the gate could not see. Now gated at creds-file load; both surfaces share one predicate in `@appstrate/afps-shared`.

2. **`DO '…'` wrote rows and passed the gate.** `DO` takes a string constant; `$$…$$` is one spelling, `'…'` the other. Measured against main:

   | case | main | fixed |
   |---|---|---|
   | `DO 'BEGIN UPDATE …; DELETE …; END';` | **0** | **2** |
   | `DO LANGUAGE plpgsql '…'` | **0** | **1** |
   | `DO $$…$$` | 1 | 1 |
   | `CREATE FUNCTION … BEGIN ATOMIC` | **1** (false +) | **0** |

   Also fixed: an `E'…\'…'` literal desynchronised the quote walker and silently blanked the rest of the file; a dynamic `MERGE` reported 3×. Regression control byte-identical — 17 findings, same 6 files, 23 live `EXECUTE format(…)` DDL sites still clean.

3. **The CLI crashed against a platform >6 days old.** `fetchRunConfigPayload` ends in an unvalidated cast over an arbitrary remote instance; `input` landed 2026-08-21; the published CLI has no version handshake. `inherited?.input.values` threw `undefined is not an object`. The `?.` is **not** restored — the cast boundary now refuses by name, pointing at the instance and at `--no-inherit`.

4. **Two registers excused things nobody checked.** `KNOWN_DRIFT`/`KNOWN_REVERSE_DRIFT` had no liveness check while this same PR argues for one in the migration gate. Now fail on a dead key or a field suppressing no finding. The reverse check was also blind to `| undefined` (vs `?`) — reads `ts.TypeFlags.Undefined` now, no new findings across the 29 entries.

### Records — `48966836`

5. **0054 justified an irreversible `DROP COLUMN` with the wrong argument.** A reviewer found a second client-authoritative writer in `29d0dd586`. Checking rather than accepting changed the answer *twice*: `git tag --contains` is empty, it lives only on two abandoned branches, and the endpoint was gone before the file reached mainline — so on any released build `upsertMessage` really is the only writer, and on a branch build it is not. The header now rests on the claim true either way — **nothing reads them** — with pre-flight queries so an operator can measure. `DROP COLUMN` gains `IF EXISTS`, matching every precedent since 0040.

6. **0005 inverted its own urgency.** It warned of a "silent 401". Wrong: gate (1d) runs on every read of a stored manifest, so the integration fails validation and does not load — a 404 naming the wrong cause, omission from the list, a 412 on runs. This is not a backlog item, it belongs in the deploy window. Better: the halves are **order-free** and running 0005 *first* gives a zero-length outage window, since `"Bearer "` was already left alone by the old code.

7. **CHANGELOG had no entries and carried one this branch made false** ("still accepted, now stripped by non-strict Zod"; "no runtime behaviour changes"). Corrected in place; 4 entries added, 3 BREAKING — `detect:breaking` models neither request-body `additionalProperties` tightening nor module-owned schemas, so it is blind to all of them by construction.

### Release obligation

**`@appstrate/afps-shared@0.6.0` must be published before the next `@appstrate/core` release.** In-tree resolves `workspace:*` so CI is unaffected, but npm has only 0.5.0 and a published core declaring `^0.6.0` would fail to install for `cloud` and `connect-helper`. Recorded at `packages/core/CHANGELOG.md`.

### Verification

`bun run check` 36/36 · `TEST_TIER=0 bun test` **11684 pass / 0 fail** across 861 files · gate regression control byte-identical · 51 manifests confirmed to differ **only** in version+prefix, the 15 untouched verified clean · TS↔SQL regex equivalence proven codepoint-by-codepoint · zero TODO/FIXME/`any`/`@ts-ignore` added.

One correction to an earlier claim of mine in this thread: I first "confirmed" the `DO` bypass with a probe that passed the filename as the SQL argument, so it measured nothing. Re-measured correctly above — the finding was genuine.

PR #1230, opened 2026-08-27
Remediate the post-beta.51 audit: close the gates, then what they missed

pierrecabriere:
Remediation of the post-beta.51 audit. Eight commits, one per phase, each landing green.

The premise: `bun run check` passed on all 42 commits since the last deploy. It passed because **nothing required it to** — and because four of the checks inside it could not fail.

## The structural finding

The `main` ruleset is `["pull_request", "non_fast_forward"]` with `required_approving_review_count: 0` and no `required_status_checks`. No CI check blocks a merge; `CODEOWNERS` is inert. That is not fixable in a PR — see **Ops actions** below — but it explains the shape of everything here, and why commits like "close the 28 P0 defects a green check could not see" keep being necessary.

## What each phase did

**Security.** Three holes, each reproduced before being fixed.

The `authorized_uris` allowlist matched the raw agent-supplied target, and authority wildcards compile to `[^/]*` — but `?` and `#` end an authority too. `https://attacker.example?.salesforce.com/x` satisfied a `*.salesforce.com` pattern while contacting `attacker.example`, and the SSRF net only refuses private ranges, so a public attacker host passed **with the injected credential attached**. 13 shipped integrations use wildcard-host patterns.

The sidecar's control surface had no inbound auth, documented as "the security boundary is the per-run Docker network" — which stopped being true when integration runners were attached to that network and handed the `sidecar` hostname. A `source.kind: "local"` integration could reach `/llm/*` and spend the org's provider key. Now behind a per-run token that is deliberately **not** `RUN_TOKEN`, so the zero-knowledge invariant holds.

Auditing alias masking found **five** signature-family fields crossing the boundary, not the two reported — and `contentSignature` is emitted by 4 of 5 backing shapes, so it is its *absence* that identifies the vendor.

**Gates.** The OpenAPI lint set its exit code only from `severity: error` while every violated rule sat at `warn`, so its allowlist — whose comment promises "any NEW violation still surfaces" — was decorative. The conformance monitor captured its exit code and never re-raised it: green while reporting 4 real failures, with #1206 open and unread since 2026-08-24. The refresh-strategy ceiling blocked growth but passed silently if you also bumped the constant. Nothing checked the compose version fallback, stale at `beta.41` across 82 sites while releases reached `beta.53`. Each is now able to fail, and was watched failing.

**Correctness.** A partial launch leaked container + network + volume, with `RUN_TOKEN` and model credentials readable via `docker inspect`, because handles were assigned after the await. The atomic concurrency cap was fail-open. Four outbound calls had no deadline, including the run-liveness heartbeat — whose stall caused the watchdog to reap containers that were alive, the exact failure it exists to prevent.

**Transitional code.** Four retired shapes still worked, including an OIDC magic-link fallback that the social leg twenty lines above explicitly refuses against the same threat.

**Database.** `audit_events.space_id` had `ON DELETE SET NULL` twelve lines below a docblock arguing the table must outlive what it describes. Dropping that FK would have **silently broken the unapplied `0003`**, which derives its column list from FKs into `spaces` — amended with `UNION` so both deploy orders rewrite exactly once. Two constraint names exceeded Postgres' 63-byte limit, so the catalog held a name the schema didn't know; the next `drizzle-kit generate` touching either FK would have aborted the batch.

**Duplication.** 56 route bodies are now `.strict()`. Three of six consolidations were **refused** on measurement — and diffing the two `guardedFetch` copies found the shared one downgrading a 302'd PUT/PATCH/DELETE to a bodyless GET, against WHATWG step 11.

**Tests.** Six tests that could not fail now can, each verified by mutating the real source and watching the intended test fail alone.

**Docs.** ~30 false claims, three of which instructed the wrong action — including a testing skill telling readers to create module migrations the contract removed.

## Six audit findings were refuted

Verification was required before deletion, and it paid:

- **`turnLimitReached`'s "redundant" booleans** — before #1173 two writers computed genuinely different things; production holds rows where they diverge.
- **knip "blind" on `packages/ui`** — it isn't; a documented rule suppresses vendored shadcn families and names `DialogClose` as a keep.
- **`ModelGenerationControls`' "single call site"** — there are two; deleting those props would have broken the chat model picker.
- **`/spaces` "unreachable"** — an e2e page object used it.
- **`scheduler.ts:1028`** — unreachable through its own function, which gates on a scoped read first.
- **`evaluateBudget` simplification** (mine) — would have changed what gets billed.

## Verification

`bun run check` → **exit 0, 38/38 tasks**. Full suite → **11,696 pass / 0 fail** across 874 files. More test than source: **+5,196 test lines vs +3,691 source**.

One pre-existing failure is **not** from this branch: `scheduler.test.ts` "fails fast … when a declared integration package is missing" fails identically on unmodified `main` (27 pass / 1 fail, tier 3).

## Ops actions this PR cannot perform

1. **Add `required_status_checks` to the `main` ruleset.** Without it the other seven commits are advisory.
2. **Publish `@afps-spec/schema@0.7.0`**, then bump the three `^0.6.1` pins. Until then manifest `config` remains a live accept-path — it is *declared and retained*, not merely tolerated like an unknown key.
3. **Rehearse and apply `scripts/migration/0003`**, then `VALIDATE` its three `NOT VALID` constraints. Note this PR amended `0003`.
4. **Drain in-flight containers before deploying** — pre-#1204 runtimes still stamp `dataschema`, which ingestion now rejects.

## Follow-ups deliberately not in scope

- `@appstrate/core` dead exports would force a 10.0.0 major and a consumer lockstep.
- The 62 remaining custom-error-class sites; no lint rule exists that isn't wrong 60 times.
- `listInstalledPackages` returns `draft_manifest` under `spaces:read`, bypassing per-type scopes.
- knip 6.32.2 now derives what `manifestEntries()` writes by hand — 127 hints, needs measurement.


pierrecabriere:
## Known issue: the integration job is ~2.9× slower on this branch

Recording this openly rather than leaving it in the timeout bump.

**Measured.** Integration job: main is stable at **383–529s** across six commits (including #1229); this branch runs **1044–1200s**. Real, reproducible, and ours. The distribution matters: median per-test is flat (0.29ms → 0.40ms) and the 10,468 sub-50ms tests are untouched, while the mean triples (25ms → 85ms). The cost is concentrated in the ~1,100 DB/IO-bound tests.

**What I ruled out**, each by local measurement rather than argument:

| Hypothesis | Result |
|---|---|
| New migration cost | tier 0: ours **faster** (3.04s vs 3.83s) |
| Coverage instrumentation | proportionate (17.71s vs 16.41s for 27 more tests) |
| Two in-process PGlite instances poisoning the run | no effect (+1.27s for their own 13 tests) |
| The two new indexes / real Postgres | identical (4.97s vs 4.91s) |
| Redis client churn from `closeResumableStore` | **1** connection for the whole run |
| Leaked transaction from the new `SELECT … FOR UPDATE` | early `return` inside `db.transaction` commits and releases |

**What reproduces.** With a real Redis at 373-file scale: ours **103.79s** vs main **87.57s** (+18%), while main is unaffected by Redis being present at all (86.32s → 87.57s). CI runs 878 files, so a per-file cost extrapolates in the right direction. I have not isolated which change causes it — a local bisect needs a per-commit `bun install`, which I did not run.

**Fixed separately, and proven:** the one actual failure was `mcp audit + rate limiting`, which truncates the DB but not Redis and so depended on unspent limiter budget. 24 other suites already flush; that one did not.

**The `timeout-minutes: 10 → 20` bump was premature** — I raised it before diagnosing and it then hit 20 too. It stays only so the suite can finish and be reviewed. It is not a fix, and the comment in `test.yml` says to split the suite rather than raise it again.

pierrecabriere:
## Correction: the "+18% with Redis" figure was my measurement error

I posted that number above. It was wrong, and the bisect that chased it found nothing because there was nothing to find.

**The error.** That run reused a database an earlier suite had already populated, so `truncateAll()` (a `DELETE FROM` per table, per test) had far more rows to remove on every `beforeEach`. Every other measurement used a fresh database. I compared a dirty DB against clean ones and reported the delta as a regression.

**Re-measured under identical conditions — fresh DB, real Postgres, real Redis, `apps/api/test`:**

| commit | | tests | time |
|---|---|---|---|
| `main` @ 7d3f7d146 | | 4442 | 87.57s |
| `229f691f5` | commit 3 of 10 | 4430 | 83.38s |
| `958ddf032` | commit 5 — chat/resumable | 4453 | 81.91s |
| `f8d21ea57` | commit 7 — 56 strict bodies + DRY | 4491 | 86.84s |
| `51997ffae` | commit 9 — CI fixes | 4491 | 88.01s |
| `0a4ecac6e` | **HEAD**, post-merge | 4492 | **91.87s** |

**+4.9% wall-clock for +1.1% more tests.** Flat across every commit. There is no local regression, at any point on this branch, in any configuration I could construct — including the one I claimed showed it.

**So the CI figure (1044s vs main 383s) is still unexplained, and I am not going to guess at it again.** What is now established:

- it is not the migration, the indexes, coverage, PGlite, Redis client churn, a leaked transaction, or the resumable-store change — each ruled out by measurement
- it does not reproduce locally at 373 files against real Postgres and real Redis
- the one genuine CI failure (`mcp audit + rate limiting`) had its own cause — a `beforeEach` that truncates the DB but not the Redis-backed limiter budget — fixed in `0a4ecac6e` and proven

The remaining difference between my harness and CI is scale (878 files vs 373) and the runner itself. Bisecting that means pushing commits and watching CI, roughly 20 minutes per data point.

**`timeout-minutes: 20` should be treated as unjustified** until that is done. I raised it before diagnosing, which was the wrong order, and the diagnosis has not yet supported it.

PR #1265, opened 2026-09-05
feat(cli): ship org MCP with the synced Claude plugin

### LINKED ISSUE #1261: skills sync: ship the organization's MCP server in the Claude Code plugin, and a one-command Codex install
Follow-up to #1251 (`appstrate skills sync` + the `appstrate/claude-plugins` marketplace). The plan (`docs/plans/claude-code-skills-sync.md`) rejected this in D12 as out of scope and listed it under follow-ups as "nearly free". This issue specifies it.

## Goal

A user who installs the `appstrate` Claude Code plugin gets the organization's MCP server (`<instance>/api/mcp/o/<orgId>`) alongside the organization's skills, without a manual `claude mcp add`. Codex users get a one-command equivalent.

## Today

- The generated plugin tree (`apps/cli/src/lib/skills-sync/targets.ts`) holds exactly `.claude-plugin/plugin.json`, `README.md`, `skills/`. No MCP.
- Every user who wants the org MCP adds it by hand (`claude mcp add --transport http appstrate <url>`), one entry per org, and authenticates via `/mcp`.
- Codex gets `~/.agents/skills/` from the same run and nothing else. The CLI README states it never writes `~/.codex/config.toml`.

## Part 1 — Claude Code: emit `.mcp.json` in the generated plugin

Verified against the current plugin docs (`code.claude.com/docs/en/plugins-reference`, `/mcp`, `/managed-mcp`):

- A plugin declares MCP servers via `.mcp.json` at its root (or `mcpServers` in `plugin.json`). Remote Streamable HTTP form: `{ "appstrate": { "type": "http", "url": "…" } }`.
- Plugin-provided remote servers get the same OAuth path as user-level ones: RFC 9728 protected-resource discovery + dynamic client registration. That is exactly what `apps/api/src/modules/mcp` serves (`/.well-known/oauth-protected-resource/api/mcp/o/:org`), and what the existing hand-added entries already rely on (`type: http`, no headers). First use = one browser login through `/mcp`, then available every session.
- Plugin servers go through the same per-server consent as a project `.mcp.json`, and tools are namespaced `mcp__plugin_appstrate_appstrate__<tool>`.
- Enterprise policy (`managed-mcp.json`, `allowedMcpServers` / `deniedMcpServers`) applies to plugin servers too. There is no `disableCommandPluginSources` equivalent for MCP.
- `/reload-plugins` picks up plugin MCP changes without a restart. The docs do not say whether a command-source plugin update that changes `.mcp.json` reconnects automatically.

### Design

- `.mcp.json` content is a pure function of the profile: `(instance, orgId)`. It stays byte-identical across syncs, so the plugin hash (its version, D6) only moves when skills move or the pinned org changes. No timestamps, sorted keys, same rule as the rest of the tree.
- Emit it only in the connected tree. The pre-login setup plugin (`setupPluginFiles`) has no org, so no `.mcp.json` there.
- `pluginFixedFiles()` takes the profile. `.mcp.json` joins `PLUGIN_ROOT_ENTRIES`; `pluginTreeMatches` already compares fixed files by content, which covers the "skills unchanged, MCP entry changed" case (org switch, CLI upgrade).
- Scope mismatch to state in the README: skills are per space, the MCP endpoint is per org. Same org-level scoping as the endpoint itself; RBAC inside the endpoint is unchanged.
- No opt-out flag: the marketplace command string is static (D9), and the policy case is already covered by the Claude Code MCP denylist. If an opt-out is ever needed it goes in the CLI profile, not in the marketplace entry.
- Server name `appstrate` (so the tools read `mcp__plugin_appstrate_appstrate__search_operations`). One server = the pinned org, same limit as the skills (D9). Multi-org stays manual.

### Rollout effect

The first sync after the CLI upgrade changes the plugin hash for every installed user: Claude Code reinstalls the plugin and prompts once for the new MCP server. Document it in the CLI README (`appstrate skills`) and the marketplace README.

### To verify before merging

- [ ] With the server denied by `deniedMcpServers` / a `managed-mcp.json` that does not list it, the plugin's skills still load. (Docs say only the servers are filtered; not exercised.)
- [ ] `appstrate org switch` + `claude plugin update appstrate@appstrate` swaps the endpoint and Claude Code reconnects, or `/reload-plugins` is needed; state which in the README.
- [ ] Fresh install on a machine with no existing `appstrate` MCP entry: consent prompt, `/mcp` login, tool list.

### Not in v1

Reusing the CLI's keyring token instead of a second OAuth login. Plugin `.mcp.json` supports `headersHelper` (a command that prints headers), and the MCP route accepts API keys (`apiKeyGrantable`, integration tests use `mcp:read` / `mcp:invoke` scoped keys). Open question: whether the device-flow token the CLI stores is accepted on `/api/mcp/o/:org`. More code and a second credential path; the OAuth flow is one browser round-trip and already proven. Revisit only if the double login is a real complaint.

## Part 2 — Codex: a one-command install, not a plugin

Verified against `learn.chatgpt.com/docs/plugins`, `/docs/extend/mcp`, and `openai/codex@main` (`codex-rs/core-plugins`, `codex-rs/codex-mcp/src/plugin_config.rs`, `codex-rs/rmcp-client`):

- Codex has plugins (`.codex-plugin/plugin.json` with `mcpServers` → `.mcp.json`, `oauth_resource` for RFC 8707) and does DCR / CIMD through rmcp, with RFC 9728 discovery.
- But marketplace plugin sources are `local | url | git-subdir | npm` only. There is no command source. Install copies the tree into `~/.codex/plugins/cache/…`, nothing re-runs per session, local marketplaces are never auto-pulled, and a changed local plugin needs `codex plugin add` again plus a new thread.

So a generated Codex plugin would reintroduce the manual step the whole feature exists to remove. Rejected.

### Design

- New opt-in command, e.g. `appstrate mcp install --target codex`, that shells out to `codex mcp add appstrate --url <instance>/api/mcp/o/<orgId>` and prints the follow-up `codex mcp login appstrate`.
- Codex owns the write to `[mcp_servers.appstrate]` in `~/.codex/config.toml` and is idempotent by server name. The sync never touches the file; the README rule stands.
- Same command with `--target claude` can wrap `claude mcp add --transport http …` for users whose org blocks command-sourced plugins (`disableCommandPluginSources`), mirroring the existing `--target claude-user` fallback for skills.
- Minimal alternative if the command is judged not worth it: document the two `codex mcp` commands in the CLI README under "Codex, and running without a Claude Code plugin".

## Deliverables

- [ ] `targets.ts`: profile-dependent `.mcp.json`, root entries, tree match; unit tests (byte-stable across runs, absent in the setup plugin, changes on org switch).
- [ ] `apps/cli/README.md` → `appstrate skills`: what the plugin now ships, the one-time consent + `/mcp` login, the org-vs-space note, the reinstall on first sync after upgrade.
- [ ] `appstrate/claude-plugins` README: same, plus `deniedMcpServers` as the opt-out.
- [ ] `docs/plans/claude-code-skills-sync.md`: flip D12, move the follow-up into decisions, record the Codex rejection (no command source) so it is not re-proposed.
- [ ] Part 2 command (or the README-only fallback) with a test around the argv it builds.
- [ ] Release: `cli@<version>` tag; the marketplace command string does not change, so no push to `appstrate/claude-plugins` is required for Part 1 beyond the README.

pierrecabriere:
## Summary

The synced Claude Code plugin supplies skills but currently leaves users to configure Appstrate MCP separately. Connected syncs now emit a deterministic `.mcp.json` for the profile's instance, organization and pinned space (`X-Space-Id` header, validated server-side against the org), using the existing OAuth flow without copying CLI credentials. MCP operations therefore land in the same space the skills were synced from.

Writing and matching share the expected files, so an organization/instance/space switch or missing MCP file repairs the plugin even when no skill needs downloading. Setup plugins omit MCP, and a previously connected plugin with zero skills survives a missing profile. Atomic replacement and shared-directory ownership remain in place.

Documentation covers independent OAuth, the space header and its stale-space failure mode, manual-server coexistence (manual connections without the header land in the org default space), opt-out, and Codex's accepted documentation-only setup. Endpoint changes require a new Claude session: the tested SDK/inline-plugin reload retained the old connection, while marketplace update itself requested a restart.

Closes #1261. Companion marketplace documentation: https://github.com/appstrate/claude-plugins/pull/1.

## Validation

- CLI suite: 1,457 passed; focused skills suites: 79 passed (`skills-command` + `skills-plan`).
- Appstrate MCP and resource-audience suites (Tier 0): 32 passed.
- `bun run check`: 38/38 tasks passed; CLI build and direct CLI typecheck passed.
- Independent Standards and Spec reviews: no actionable findings.
- Real Claude 2.1.261 with the exact generated plugin: connects; manual MCP coexists; denylist and per-project disable retain skills; a fresh session selects the changed endpoint.
- Real command-source marketplace install/update in copy mode: changing only `.mcp.json` changed the installed cache hash.
- Real Claude Code 2.1.263 against a Tier 0 instance of this branch: `claude mcp login` on the plugin server completed CIMD OAuth with the `X-Space-Id` header present; `get_me` resolved the pinned space (skill of space B listed), and the same plugin without the header resolved the org default space (0 skills). See the smoke-test comment.
- Complete OAuth flows against a local fixture for organizations A and B: registration, state/PKCE, callback, resource-specific token exchange, then authenticated tools/list. B required its own login; A's bearer was not reused for B. Tests used isolated config and a fake keychain executable, with no real credentials or model calls.

## Rollout and validation limits

Publish the next `cli@<version>` after merge and align the companion README publication. The marketplace command is unchanged; older installed CLIs remain skills-only until updated.

The local fixtures do not establish live Appstrate-account browser login, automatic once-per-session refresh, or system-level `managed-mcp.json` exclusive mode. The latter was not exercised because Docker was unavailable and host policy was left untouched. Existing server audience checks and real-client denylist/disable behavior were tested. No server authentication or RBAC implementation changes are included.


pierrecabriere:
Pushed 2003e5d58: the generated `.mcp.json` now carries `headers: { "X-Space-Id": <pinned space> }`, so MCP operations run in the same space the skills come from instead of the org default space. The server already validates that header against the org (`resolveMcpSpaceRow`) and forwards it on every in-process dispatch. A space switch now rewrites the plugin like an org/instance switch; the "space-only change keeps bytes" test became "login-only change keeps bytes" plus a third case in the endpoint-update loop. README documents the header, the stale-space 404, and the `--header` / `http_headers` opt-in for manual Claude Code and Codex connections. The plan records the release note about `mcp__plugin_appstrate_appstrate__*` tool names vs a manual `mcp__appstrate__*` allowlist. Follow-up for agents as slash commands: #1268.

pierrecabriere:
Real-client smoke test of 2003e5d58 (Claude Code 2.1.263, Tier 0 instance started from this branch on :3100, throwaway account):

- Fixture: org with default space A and a second space B; one skill published in B only; CLI profile pinned to B (API key seeded in the keyring).
- `appstrate skills sync --print-path` produced a plugin with `skills/space-b-skill` and a `.mcp.json` carrying `headers: { "X-Space-Id": "<B>" }`.
- `claude --plugin-dir <plugin> mcp login plugin:appstrate:appstrate --no-browser` completed the real CIMD OAuth flow (authorize → consent → callback) against the platform, with the header present. Custom headers and OAuth coexist as the Claude Code docs state.
- `claude -p --plugin-dir <plugin>` calling `mcp__plugin_appstrate_appstrate__get_me`: `skills_total: 1` with `@smoke/space-b-skill` → resolved space B.
- Control: same plugin with the `headers` block removed, re-authorized, same call: `skills_total: 0` and the default-space hello-world agent → resolved space A. The two halves discriminate.
- Observed on the way, unrelated to this PR (present on main): the very first `authorize` for a fresh CIMD client returns `invalid_scope` for `mcp:read mcp:invoke offline_access`; the second attempt succeeds. Looks like the scopes set by `markClientSelfService` in `onClientCreated` land after the in-memory client is validated. Filed separately.

PR #1274, opened 2026-09-07
feat(cli,mcp,api): local authoring loop for packages — pull, status, push, publish; sync per profile and across spaces

### LINKED ISSUE #1267: OIDC: a DCR client registered without `scope` can never be granted `mcp:read` / `mcp:invoke` (breaks `claude mcp add` on a fresh instance)
## Summary

A client that self-registers through Dynamic Client Registration (RFC 7591) on `POST /api/auth/oauth2/register` **without a `scope` field** is stored with the identity scopes only (`openid profile email offline_access`). From then on, every `GET /api/auth/oauth2/authorize` that asks for `mcp:read mcp:invoke` is bounced back to the client's `redirect_uri` with `error=invalid_scope`, before any login or consent screen. The client can never be granted the MCP scopes, so the zero-config "Path B" onboarding (`claude mcp add --transport http ...`, then `/mcp` -> Authenticate) cannot complete on a fresh instance.

Claude Code's DCR request carries no `scope` (its registration body is `client_name`, `redirect_uris`, `grant_types`, `response_types`, `token_endpoint_auth_method`; the scope it later requests at `/authorize` comes from the protected resource metadata plus `offline_access`), so this is exactly the request shape that hits the bug.

## Environment

- Reproduced on a clean checkout of `origin/main` at `df1e32073bda84df8e71b9ccddfbb659a53497ac` (`chore: bump compose version fallbacks to 1.0.0-beta.57 (#1259)`), unmodified.
- Tier 0: `bun apps/api/src/index.ts`, PGlite, `.env.example` values, `MODULES=oidc,webhooks,mcp,core-providers,@appstrate/module-chat` (the default), `NODE_ENV=development`, `APP_URL=http://localhost:4100`, `AUTH_BOOTSTRAP_OWNER_EMAIL` set so a bootstrap organization exists.
- `@better-auth/oauth-provider` 1.7.0-beta.4 (from the lockfile).
- Also reproduced, identically, on Tier 1 (`appstrate install --tier 1 --yes`, CLI 1.0.0-beta.57) running the published image `ghcr.io/appstrate/appstrate:1.0.0-beta.57` (`org.opencontainers.image.revision` = the same `df1e32073...`), PostgreSQL, port 4200. Same responses at every step, only the ids differ.
- Not specific to the environment or the DB backend, so Tiers 2 and 3 were not run (same image, same code path).

## Steps to reproduce

1. Discovery advertises the MCP scopes:

```sh
curl -s http://localhost:4100/.well-known/oauth-authorization-server | jq .scopes_supported
# [..., "openid", "profile", "email", "offline_access", ..., "mcp:read", "mcp:invoke"]
```

2. Register a client the way Claude Code does, with no `scope`:

```sh
curl -s -X POST http://localhost:4100/api/auth/oauth2/register \
  -H 'Content-Type: application/json' \
  -d '{"client_name":"Claude Code (repro)","redirect_uris":["http://localhost:9999/callback"],"grant_types":["authorization_code","refresh_token"],"response_types":["code"],"token_endpoint_auth_method":"none"}'
```

Response (HTTP 200):

```json
{"client_id":"VJpRaMpSllDKClcmRgFPHTIloklimKJs","scope":"openid profile email offline_access","client_id_issued_at":1788634140,"client_name":"Claude Code (repro)","redirect_uris":["http://localhost:9999/callback"],"token_endpoint_auth_method":"none","grant_types":["authorization_code","refresh_token"],"response_types":["code"],"public":true,"disabled":false,"skip_consent":false}
```

3. Start the authorization the MCP client starts (scope from the PRM plus `offline_access`, `resource` = the org's MCP endpoint):

```sh
curl -s -o /dev/null -w "%{http_code} %{redirect_url}\n" \
  "http://localhost:4100/api/auth/oauth2/authorize?response_type=code&client_id=VJpRaMpSllDKClcmRgFPHTIloklimKJs&redirect_uri=http%3A%2F%2Flocalhost%3A9999%2Fcallback&scope=mcp%3Aread%20mcp%3Ainvoke%20offline_access&state=x&code_challenge=E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM&code_challenge_method=S256&resource=http%3A%2F%2Flocalhost%3A4100%2Fapi%2Fmcp%2Fo%2Fc4a54b9f-0a89-4dd3-9913-ae658ccb90fc"
```

Response:

```
302 http://localhost:9999/callback?error=invalid_scope&error_description=The+following+scopes+are+invalid%3A+mcp%3Aread%2C+mcp%3Ainvoke&state=x&iss=http%3A%2F%2Flocalhost%3A4100%2Fapi%2Fauth
```

Same result with or without the `resource` parameter; the scope check fires first.

4. Control: the same registration with an explicit `scope` works, which shows only the default is at fault:

```sh
curl -s -X POST http://localhost:4100/api/auth/oauth2/register -H 'Content-Type: application/json' \
  -d '{"client_name":"Claude Code (repro, explicit scope)","redirect_uris":["http://localhost:9999/callback"],"grant_types":["authorization_code","refresh_token"],"response_types":["code"],"token_endpoint_auth_method":"none","scope":"mcp:read mcp:invoke offline_access openid"}'
# -> {"client_id":"KsHxseHgAgeqDBagGCODfiiYSXdJUaMY","scope":"mcp:read mcp:invoke offline_access openid", ...}
```

The same `/authorize` request with that `client_id` redirects to the login page as expected:

```
302 http://localhost:4100/api/oauth/login?response_type=code&client_id=KsHxseHgAgeqDBagGCODfiiYSXdJUaMY&redirect_uri=http%3A%2F%2Flocalhost%3A9999%2Fcallback&scope=mcp%3Aread+mcp%3Ainvoke+offline_access&state=x&...&resource=http%3A%2F%2Flocalhost%3A4100%2Fapi%2Fmcp%2Fo%2Fc4a54b9f-0a89-4dd3-9913-ae658ccb90fc&exp=...&sig=...
```

For reference, the MCP endpoint's challenge and PRM, which is where the client gets the scopes it asks for:

```
POST /api/mcp/o/<orgId> (no token)
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer resource_metadata="http://localhost:4100/.well-known/oauth-protected-resource/api/mcp/o/<orgId>", scope="mcp:read mcp:invoke"

GET /.well-known/oauth-protected-resource/api/mcp/o/<orgId>
{"resource":"http://localhost:4100/api/mcp/o/<orgId>","authorization_servers":["http://localhost:4100/api/auth"],"scopes_supported":["mcp:read","mcp:invoke"],...}
```

## Expected

A scope-less DCR registrant gets the same self-service scope set a registrant may request (identity + module end-user scopes, i.e. `mcp:read` / `mcp:invoke`), so that `/authorize` for the MCP resource proceeds to `/api/oauth/login` and the consent screen decides what is actually granted.

## Actual

The registrant is stored with `openid profile email offline_access` only, and `/authorize` with `scope=mcp:read mcp:invoke offline_access` is rejected with `invalid_scope` before login. `claude mcp add` never completes on a fresh instance.

## Root cause

`apps/api/src/modules/oidc/auth/plugins.ts`:

```ts
allowDynamicClientRegistration: true,
allowUnauthenticatedClientRegistration: true,
clientRegistrationDefaultScopes: [...OIDC_IDENTITY_SCOPES],
clientRegistrationAllowedScopes: selfServiceScopes,
```

`clientRegistrationDefaultScopes` is the identity set alone. The oauth-provider plugin's `/authorize` handler validates the requested scopes against `client.scopes ?? opts.scopes`; since the stored `scopes` is a non-empty array, the module scopes are not in it and the request fails.

`markClientSelfService` (`services/oauth-admin.ts`) does backfill the full self-service ceiling, but **only when `scopes` is empty** (`if (!row.scopes || row.scopes.length === 0)`), which is the CIMD case. The DCR default is non-empty, so the after-hook on `/oauth2/register` never widens it. The comment there ("The DCR register path dodges this via `clientRegistrationDefaultScopes`") assumes the default covers the module scopes; it does not.

The existing integration test (`test/integration/services/dcr-cimd.test.ts`, "registers a public client unauthenticated with identity scopes") always passes an explicit `scope`, so the scope-less default is not exercised.

## Proposed fix

One line in `plugins.ts`:

```ts
clientRegistrationDefaultScopes: selfServiceScopes,
```

plus a test in `dcr-cimd.test.ts` asserting that a registration without `scope` yields the same set as an explicit request for the full allowed set. @otarbes has this fix (with the test) on a local branch, not pushed yet, and can open a PR if that is the direction you want.

## Why the cloud org is not affected

Clients that already exist there were registered (or repaired) before, and once a client row carries the MCP scopes the check passes. The bug only shows up on a fresh instance or for a freshly registered scope-less DCR client.

otarbes:
## Summary

A local authoring loop for packages, at parity between the CLI and the MCP server, plus two fixes found on the way. Motivation: TRACTR is moving its 48 team skills (and its agents and integrations) onto Appstrate as the single registry. Editing them means: pull the draft into a working folder, edit with any tool, push the folder back as the draft (annex files included), sync the draft to the machine, publish. Today the platform only offers the classic import (publishes and freezes in one step) and the JSON `PUT` (manifest + SKILL.md, no annex files).

```mermaid
flowchart LR
  subgraph machine["Workstation / coding agent"]
    wd["~/Appstrate Packages/<org>/packages/<type>s/<name>"]
    view["~/.claude/skills (sync view)"]
  end
  subgraph cloud["Appstrate"]
    draft["draft (manifest + files, lock_version)"]
    ver["published versions (immutable)"]
  end
  wd -- "packages push  /  push_package_files" --> draft
  draft -- "packages pull  /  pull_package_files" --> wd
  draft -- "packages publish  /  create<Type>Version" --> ver
  ver -- "skills sync --source published" --> view
  draft -- "skills sync --source draft" --> view
  wd -. "packages status  /  package_status" .-> draft
```

### API

- `POST /api/packages/import?draft=true` writes the uploaded archive as the package draft, annex files included, and cuts no version. `draft_overwrite` protection still applies.
- `lock_version=N` on the same route lets a client re-push its own work without `force`: the import is accepted only if the draft's lock is the one the client last saw. A chat edit in between yields the existing `409 draft_overwrite`.
- Response carries `draft: true`, `draftVersion`, `lock_version`.

### CLI

- `appstrate skills sync` is **per profile** (ledger keyed `<profile>:<target>`, legacy ledger adopted) and accepts `--space` repeatedly (`syncSpaces` in the profile). Two profiles on one machine no longer delete each other's skills; a machine can receive the union of several spaces.
- `appstrate packages pull | status | push | publish` (with `skills` aliases) for skills, agents, integrations and MCP servers. Working copies live under `~/Appstrate Packages/<org slug>/packages/<type>s/<name>` (`workDir` in `config.toml`); `~/Appstrate` is refused because it is an instance install directory that `uninstall --purge` removes.
- `push` synthesises the manifest from `SKILL.md` frontmatter, remembers the draft lock per profile, bumps the patch when the declared version is already published, and waits out the import rate limit (`429`, `Retry-After`).
- `status` compares a working folder to the draft on demand, file by file, with `--diff`, and flags a draft edited elsewhere.

### MCP

- New tools `pull_package_files`, `package_status`, `push_package_files` (optional `space`, all package types), same routes as the CLI, so a coding agent connected over OAuth has the same loop. Internal dispatch carries `internalDispatchHeader()`: without it an OAuth caller (token confined to the MCP resource, RFC 8707) got a 401 where an API key passed.

### Fixes

- **OIDC DCR default scopes** (`09b6d07c3`, closes #1267): a dynamically registered client that names no `scope` now receives the self-service scopes (`mcp:read mcp:invoke offline_access`). Before, `claude mcp add` against a fresh instance failed with `invalid_scope`. Reproduced on `origin/main` and on image `1.0.0-beta.57`. Independent of the rest, safe to cherry-pick first.
- `chore(web)`: regenerated `apps/web/src/api/schema.d.ts` for the new import parameters; `fix(openapi)`: the `lock_version` response field uses the 3.1 `["integer", "null"]` form.

## Ordering with #1260

#1260 (granular roles) and this branch touch five of the same files (`routes/packages.ts`, `openapi/paths/packages.ts`, `modules/mcp/router.ts`, `modules/mcp/tools.ts`, `modules/oidc/auth/plugins.ts`). #1260 goes first; I will rebase this branch on it once merged. The OIDC fix is independent and can be cherry-picked before either.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update (`apps/cli/README.md`, `docs/guides/connecting-mcp-clients.md`)
- [ ] Refactoring

## Verification

- CLI: `bun test` in `apps/cli`, 1483 pass (38 in the four new/extended files).
- API: `TEST_TIER=0 bun test` from the repo root on `packages-import-draft.test.ts` (6) and `mcp-package-drafts.test.ts` (7, agent and cross-space cases included): 13 pass; `oidc/.../dcr-cimd.test.ts`: 8 pass.
- End to end on a local instance built from this branch (PGlite): push of a skill with scripts → draft without version → draft sync → re-push without `--force` → publish → published sync; the same loop through the MCP tools from Claude Code over OAuth; 48 skills pushed and published across four spaces, sync counts per pinned space and per union verified; an agent pulled from the cloud, pushed and published locally.

## Checklist

- [x] `bun run check` passes (38/38 tasks on `60d6d9969`)
- [x] `bun test` passes
- [x] I have read the [Contributing Guide](../CONTRIBUTING.md)


pierrecabriere:
This PR carries two mechanisms with different lifecycles, and they are easier to review and to land separately:

1. **Distribution** — `Appstrate → local install`, one way, refreshed by the marketplace `command` source once per session. Multi-space sources, per-profile ownership, cleanup on logout.
2. **Authoring** — `working folder ↔ Appstrate draft`, an explicit cycle: `packages pull` / `status` / `push` / `publish`, `POST /api/packages/import?draft=true`, `lock_version`, and the three new MCP tools.

The first half is now split out as #1294, rebased on current `main` (this PR is `CONFLICTING`). It is CLI-only — no server change, no migration, no new environment variable, no API surface change — and it commits to one rule: **one active context per destination**, profiles do not accumulate installations, and `X-Space-Id` stays on the pinned space independently of which spaces supply the skills.

Suggested follow-up once #1294 lands: rebase this PR on it and keep only the authoring loop plus the OAuth DCR scope fix. That also resolves the conflicts, since the overlap is entirely in `apps/cli/`.

The review findings that belong to the authoring half stay open here and are worth addressing in the rebased PR:

- `packages status` excludes `manifest.json`, so a manifest-only change can report "Nothing to push" — most visible for agents, integrations and MCP servers, whose configuration lives there. `package_status` has the same gap and can answer `clean: true`.
- `pull --version latest` records the **draft**'s `lock_version` while fetching published files, so the next `push` can overwrite unpublished draft edits without a `409`.
- The draft lock check is not atomic: `lock_version` is verified before the write but is not part of the `UPDATE`'s `WHERE`, so two concurrent pushes can both be accepted.
- The text/binary heuristic can decode as UTF-8 bytes that should travel as base64, corrupting some binary annex files.

None of these touch the distribution half, which is why the split is clean.

otarbes:
Superseded by #1499, which carries the authoring half rebuilt on current main as you suggested: CLI only, on the draft file-operation PUT, following the home / share / activation model from #1437. The four review findings are addressed there, plus a fifth (the lock is now kept per working folder, not per package). The distribution half landed in #1294. Branch kept for reference.

Issue #1280, opened 2026-09-08
core: ModuleInitContext.getSendMail types the mailer as void-returning, so no module can await delivery


pierrecabriere:
## What

`packages/core/src/module.ts:1217` declares:

```ts
getSendMail: () => Promise<(to: string, subject: string, html: string) => void>
```

The returned mailer is typed as **void-returning**. The platform supplies `sendMail`, which is `async (…) => Promise<void>`, so at runtime it does return a promise — but TypeScript tells every module consumer the result is `void`.

Consequences:

- **No module can sequence on mail delivery.** `await ctx.getSendMail()(…)` typechecks as `await void`.
- Any module that does call it gets a floating promise the type system cannot see, so the new backend `@typescript-eslint/no-misused-promises` rule flags the platform's own call site rather than the contract that causes it.

No rejection is dropped today: the platform's `sendMail` try/catches internally and logs, so the impact is expressiveness, not a crash.

## Where it surfaces

`apps/api/src/lib/modules/registry.ts:161` now carries the branch's only `eslint-disable` for `@typescript-eslint/no-misused-promises`, pointing here. Found by the type-aware lint added in `chore/gate-hardening`.

## Why it was not fixed in that branch

Changing the signature to `(…) => Promise<void>` is a **published-core contract change**: a module supplying its own `() => void` mailer stops typechecking. That needs a `@appstrate/core` major, plus the consumer lockstep (`cloud`, `connect-helper`) the release process requires — not something to fold into an unrelated PR.

## Suggested resolution

1. Change the declared return type to `Promise<void>`.
2. Bump `@appstrate/core` accordingly and follow the documented publish order (tag core, then bump consumers).
3. Delete the `eslint-disable` in `registry.ts` — the gate then proves the fix landed.

PR #1281, opened 2026-09-08
chore(ci): harden the quality gate — CVEs, migrations, dropped promises, images, bundle, cycles, a11y

pierrecabriere:
## What

Hardening pass on the quality gate. The gate was already unusually strong on **conformance** — that the code matches the specs it declares (OpenAPI, module contracts, env docs, compose defaults, casing, dead code). It was blind to several classes of "is this safe to ship", which is what this closes.

`bun run check` goes from 18 task names / 38 executed tasks to 22 / 43. Cold gate 49 s → 1 m 32 s, measured in a CI-like environment with no `.env` and an empty turbo cache; the `check` job's budget is 10 minutes. Warm is unchanged (~4 s).

Every gate here follows the doctrine the existing ones established: it must **discriminate** (each ships with a negative control proving it goes red on a real defect), it never degrades silently, its accepted-violation list is checked in **both directions** so it cannot rot into a museum of old fixes, and its population is **discovered** rather than listed.

## The gates

| Gate | Question nothing answered before | Where |
|---|---|---|
| `audit:deps` | Is a known CVE sitting in `bun.lock`? | CI job (needs network) |
| pre-commit gitleaks | Is a secret about to enter git history? | `.husky/pre-commit` |
| `verify:migration-drift` | Was the Drizzle schema edited without a migration? | `check` |
| `lint:migrations` | Does this migration take a hazardous lock? | `check` |
| type-aware ESLint | Is a promise dropped on the floor in the backend? | `check` (lint) |
| jsx-a11y | Does this control have an accessible name? | `check` (lint) |
| image scan + SBOM + provenance | Do we ship images with known CVEs, unattested? | `release.yml` |
| `verify:bundle-size` | Did the SPA's first-paint weight regress? | `check` |
| `verify:import-cycles` | Did a circular import appear? | `check` |
| Codecov `project` | (was `informational: true`) | Codecov |

### 🔴 better-auth — account takeover, and this PR is deliberately red on it

`better-auth@1.7.0-beta.4` is inside **GHSA-qq9h-g4jm-xgf3**, *Account takeover via pre-account hijacking on magic-link and email-OTP sign-in*, CVSS 8.3. The `magicLink` plugin is enabled and instrumented in `packages/db/src/auth.ts`, so this is a live sign-in route, not a dormant transitive advisory.

It is the **one advisory deliberately left out** of the `ACCEPTED_ADVISORIES` baseline, so `Dependency Audit` is red until it is fixed. That is the intended behaviour, not an oversight. The fix is `1.7.0-beta.10` and it is not a one-line bump — two local patches are pinned to beta.4 and need re-authoring, with CIMD/OAuth revalidation. Tracked in a private security advisory (draft, unpublished).

### 🔴 Four images shipped a fixable CRITICAL — fixed here

`runtime-pi/runners/{node,python,binary}` and `runtime-pi/sidecar` still pinned `alpine:3.21.3`, which carries `libcrypto3`/`libssl3` `3.3.3-r0` — **CVE-2026-31789**, CRITICAL, fix available. Every other image had already moved to 3.22.4.

Bumped after measuring that nothing else moves: both tags ship **node v22.23.2** and **Python 3.12.14**. Verified after by rebuilding all four — `libcrypto3 3.5.6-r0` in the runners, `3.5.8-r0` in the sidecar, runtimes unchanged.

### 🔴 Four dropped-promise bugs — fixed here

Found the moment the type-aware rules were switched on:

- **`process-orchestrator.ts`** — `findAvailablePort` did not await `Server.stop()`, which settles when the socket is released. It returned a port still bound: the EADDRINUSE-at-sidecar-boot the function's own comment says it exists to prevent.
- **`storage-fs.ts`** (published `@appstrate/core`) — `FileSink.write()` un-awaited in the streaming upload loop. The `highWaterMark` backpressure the surrounding comment is entirely about was never applied, and a mid-write failure surfaced as an unhandled rejection instead of entering the catch that closes the sink and rolls the destination back.
- **`bun-probe.ts`** (published `@appstrate/core`) — async pipe writes guarded by a sync `try`, so a broken pipe became an unhandled rejection.
- **`local-queue.ts`** — `run().finally(...)` with no terminal catch, so a throw from the caller-supplied backoff strategy escaped as an unhandled rejection.

A fifth was **not** fixed here: `ModuleInitContext.getSendMail` types the mailer as void-returning, so no module can await mail delivery. Reconciling it is a published-core contract change needing a major and consumer lockstep — #1280, and the branch's single `eslint-disable` points at it.

### 🟡 Three a11y fixes that change rendered output

49 jsx-a11y violations, all resolved, no new translated strings — every accessible name came from text already on screen. Three changed the DOM: the import drop zone was mouse-only (its file input is `display:none`, hence unfocusable) and gains a tab stop; the connections accordion header was a `div onClick` and becomes a real `button` with `aria-expanded`; the package card title becomes a real `Link`, which also fixed a double navigation on cmd-click.

## Baselines, and why each is shaped differently

Four gates land on a codebase that already violates them, and the shape of each baseline follows from what can actually clear it:

- **advisories** (37) carry an `expires` date, because a CVE *can* be fixed by upgrading and the date forces the re-decision. 2026-10-31 where the package is reachable from product code, 2026-12-31 for dev-tree and never-imported ones — the dependency edge is recorded per entry, established with `bun pm why`.
- **migrations** (109) carry **no** expiry: a migration is keyed in drizzle's journal by file hash, so it is never edited, and no upgrade makes it not take the lock it takes.
- **import cycles** (7) carry prose describing each cycle's actual shape.
- **a11y** has no baseline at all — 49 was small enough to fix, and eslint's `reportUnusedDisableDirectives` under `--max-warnings 0` makes the eight at-site directives self-cleaning.

All of them fail on a stale entry. Neither the migration nor the cycle gate has an `--update-baseline`: a regeneration flag on that kind of list is a way to clear a new finding without reading it.

## The one baseline entry that is not grandfathered

Merging `main` brought `0056_space_roles.sql` (#1260) and the migration gate went red on three findings. They are baselined with a reason true of that file rather than fixed: 0056 already bounds its locks (`SET LOCAL lock_timeout = '3s'`, `statement_timeout = '60s'`, with a header explaining that on expiry the batch aborts and boot fails its health gate), and the remedies squawk names buy nothing there — `DROP CONSTRAINT` and `ADD COLUMN` already hold ACCESS EXCLUSIVE on the same tables earlier in the same transaction. Editing it would invalidate the journal hash for every environment that applied it, in exchange for no lock change. The gate's own prose was corrected where this made it false.

## ⚠️ Action required outside this PR

**`main` has no required status checks.** The ruleset (`Protect main`, 14614228) carries only `pull_request` and `non_fast_forward`; `required_status_checks` is absent and `branches/main/protection` 404s. Every gate in this repo — including all of these — is mergeable red today.

`CONTRIBUTING.md` now records the expected set and the `gh api` command that applies it. Applying it is a repository-settings change and is deliberately **not** done here.

## What is not covered

- The image scan runs **after** the push, so `:latest` moves before it can object; what it blocks is the GitHub Release. Scanning pre-push means building each image twice inside the one job with a history of overrunning its timeout.
- Only `linux/amd64` is scanned — Trivy resolves a multi-arch index to the host platform. The failing arm is OS packages, identical across both halves of a base-image tag.
- jsx-a11y never crosses a component boundary, so `<Button size="icon"><Trash2/></Button>` with no `aria-label` — the commonest icon control here — is invisible to every rule in the set. Closing that needs a component-resolving rule or a runtime axe pass; axe was evaluated and deliberately not added, since the e2e suite is label-gated and so could not fail the PR that caused the regression.
- `apps/cli`, `runtime-pi` and `e2e` are outside the type-aware lint scope.
- The 24-file import cycle in `apps/api/src/services` is recorded, not untangled — that is a refactor, not a gate fix.

## Verification

- `bun run check` — 43/43 green, including in a simulated CI environment with no `.env` and a cold cache.
- `bun test scripts/test/` — 369 tests. Every new gate has a colocated test with positive and negative controls.
- `bun test apps/api/test/unit/` — 1796 tests, 0 fail.
- `bun test packages/{core,afps-runtime,module-chat,ui}` — 2296 tests, 0 fail.
- `bun test` from `apps/web` — 829 tests, 0 fail.
- Every new gate has a pasted RED-on-defect / GREEN-on-clean transcript in its commit message.


pierrecabriere:
Closing: most of this branch has already reached main another way — better-auth 1.7.3 (#1287), the alpine 3.22.4 bump, the four dropped-promise fixes, type-aware ESLint, and #1280 (#1292). The dependency-audit gate is re-extracted against today's main, with a re-measured baseline, in #1489. The other gates (migration lint, bundle size, import cycles, jsx-a11y, image scan) are not being pursued for now.

PR #1283, opened 2026-09-08
chore(deps): better-auth 1.7.3 — persisted RFC 8707 resources, pinned CIMD transport, schema contract

pierrecabriere:
Dependabot PR #1249 bumped the `better-auth` family from `1.7.0-beta.4` to `1.7.2` and every job went red on a single error:

```
SyntaxError: Export named 'oauthToSchema' not found in module
  node_modules/.bun/@better-auth+oauth-provider@1.7.2/.../dist/index.mjs
```

`@better-auth/cimd` stayed pinned at `1.7.0-beta.4` because it is patched and `patchedDependencies` keys on the exact version, so it kept importing an export the 1.7 line had removed. That is only the symptom. The real work is that 1.7 reshapes the OAuth provider: the audience model, the CIMD fetch boundary, the database schema, and three registration rules all changed.

This PR does that work, targeting **1.7.3** rather than 1.7.2, and then uses the bump to retire the parts of the OIDC module that 1.7.3 now provides.

## Why 1.7.3 and not 1.7.2

1.7.2 makes `account.issuer` a required column with a `(issuer, accountId)` unique index and an identity backfill. 1.7.3 reverts that to the 1.6 core account schema and tells anyone who applied it to undo it. Landing on 1.7.2 would mean writing a migration in order to delete it one release later.

1.7.3 also makes both local patches obsolete, so the tree ends with **no patched dependencies at all**.

## Commits

| | |
|---|---|
| `4c56468` | bump the family to 1.7.3, drop both patches |
| `02d4a18` | align the Drizzle schema, migration `0057` |
| `86865df` | bump `zod` to 4.5.4 |
| `79b017a` | audiences: in-process array → `oauth_resources` rows |
| `4b8c7a5` | CIMD metadata through the pinned upstream transport |
| `93e0265` | absorb the registration and client-auth rules |
| `cd7b267` | plugin factory, and unlink by row id |
| `447bc7c` | drop the `patches/` COPY from the three Dockerfiles |
| `1cfbfb1` | present-tense docs, dead `patches/**` CI filters, stale README section |
| `a8f5497` | client scoping read from platform-owned columns, not the client-influenced `metadata` JSON |
| `9eb53ca` | token verification and resource challenges through upstream |
| `aa7593e` | Better Auth's rate limiter on the platform's shared storage |
| `4cd6552` | arm that limiter in production and the test harness only |

## The bump itself

**`zod` was a hard blocker.** `@better-auth/core@1.7.3` needs `zod ^4.5.4` and calls `z.validate`, added in 4.5.0. The root `overrides.zod` pinned `4.4.3` across the tree, so core silently resolved a zod without that function:

```
isLoopbackHost("localhost") -> TypeError: z.validate is not a function
```

That reaches production through `oidc/services/redirect-uri.ts`, which validates every registered redirect URI, and it took the whole `apps/api` suite down at preload. zod 4.5 is an additive minor; nothing else moved.

**Audiences are persisted.** `validAudiences` is gone. The AS resolves a requested `resource` against `oauth_resources` per request and answers `invalid_target` for an identifier with no row. The two platform identifiers are seeded through `resources`; the mcp module writes one row per org and its reconcile is symmetric (a lost `onOrgDelete` converges too). The 60 s tick stays for the verifier half only, which is per-process.

**The CIMD fetch boundary moved.** 1.7.3 makes the transport the application's responsibility and requires resolve-once, address pinning and no redirects. We use upstream's conforming implementation (`@better-auth/cimd/node`: HTTPS only, one `dns.lookup(all: true)` whose every answer must be publicly routable, connection pinned, 5 s / 5 KB) and keep the platform's literal denylist on the hook upstream documents for it, because upstream's public-routability check accepts the run network's Docker aliases `sidecar`, `agent` and `host.docker.internal`. Documents are validated against the `mcp-2026-07-28` profile.

**Migration `0057`** adds what 1.7.3 writes, three tables, and drops `oauth_clients.public` and `oauth_clients.type` after folding both forward. Three columns (`jwks`, `jwks_uri`, `reference_id`) had been missing since before this bump; the drift was invisible because beta.4 registered no schema check.

## Behaviour changes worth reviewing

1. **DCR loopback redirects.** 1.7.3 validates redirect URIs against the application type and assumes `web` for a DCR body that declares none, refusing `http://127.0.0.1:<port>/callback` — the `claude mcp add` path. Upstream exposes no option, so a `before` hook fills the field only when absent or `null`. Admin-created clients derive `application_type` from their redirect URIs the same way.
2. **Registered client-auth method is enforced**, and 1.7.3 reads a NULL method as `client_secret_basic`. A client registered under beta.4 without declaring a method (NULL + generated secret) authenticated with the secret in the POST body and will now get `invalid_client`. Every first-party client matches its registered method; the affected population is third-party DCR rows. **Count them before deploying** — see below.
3. **Registration scope is the ceiling.** 1.7.3 persists the union of default and allowed scopes. Default and ceiling are now one array in `plugins.ts`, so the union cannot fail open.
4. **`client_credentials` is no longer advertised.** No client in the tree can use it (`grant_types` never include it, `client_credentials_scopes` is empty), so discovery stops offering a grant that can only fail.
5. **`enforcePerClientResources: false`**, against upstream's `true` default. `clientRegistrationDefaultResources` must name identifiers present in the static boot-time `resources` array, and the per-org MCP resources are created at runtime, so `true` would 400 every self-service client at first mint. Confinement is unchanged: the org is extracted from `aud` and the membership join runs downstream.

## Retired in favour of upstream

- **Client scoping from columns.** `self_service` and `level` were read from `oauth_clients.metadata`, a JSON the provider owns and an RFC 7591 registration body may set — a client could name the flag that decides whether the token endpoint confines it to one protected resource. `self_service` is a column (0057 §E folds the key), the claim builder moves to the provider's `extensions` seam (handed the resolved client row), `cimd.onClientRefreshed` re-asserts the columns, and nothing reads the JSON.
- **JWKS cache** (TTL, single-flight, unknown-kid retry) → `verifyJwsAccessToken` from `@better-auth/core/oauth2` with an in-process `jwksFetch`. −83 lines.
- **`WWW-Authenticate` serialization** → `createResourceServerChallenge` (RFC 6750 quoting, RFC 9728 path-insertion, RFC 9449 DPoP challenge).
- **Rate limiting.** Better Auth's limiter counted in per-process memory, so a fleet of replicas granted each caller one budget per replica. It now runs on the platform's limiter factory (Redis when configured) through `rateLimit.customStorage`, armed in production as upstream defaults to (its generic rules, three sign-ups per ten seconds per IP among them, would otherwise throttle development and e2e) and armed by the test harness so the integration suite exercises the rules. It keys on the IP the platform resolves under `TRUST_PROXY` (stamped on the request; `advanced.ipAddress.ipAddressHeaders` names that header alone). The OIDC guards' per-IP limiters for `/oauth2/*` and `GET /device` duplicated the provider's per-path rules and are gone; budgets are the tighter of upstream's defaults and the previous platform ceilings (token 20, authorize 30, introspect 60, revoke 30, register 5 per minute per IP; `/device` 5 per 10 min). The per-(client, IP) token limiter was a sub-key of the per-IP rule with the same budget and could never fire first; removed. Device, CLI and login-email limiters stay local because the CLI parses their RFC-shaped 429 bodies.

**Evaluated and not adopted, on evidence:** passing the signed `oauth_query` on the email login/register legs. Upstream's `after` hook then resumes authorize *inside* the Better Auth call, which inverts the ordering our handlers depend on (org membership and end-user provisioning must fail before authorize runs) and does not remove the pending-client cookie, which the social legs still need. `oauthDeviceAuthorization()` (a real OAuth device grant) could replace `/cli/token` and ~700 lines of `cli-tokens.ts`, but it retires an endpoint every published CLI calls; that is a product decision and is left out of this PR.

## Two failures that only appear in a shared process

Both were green file-by-file and red in-suite, which is how CI runs integration tests.

A Better Auth **plugin object is single-use**: `@better-auth/cimd`'s `init()` registers its extension with a fresh object literal while upstream dedupes by object identity, so a second init on the same objects trips `extensions register client discovery id "cimd" more than once`. `createAuth()` held the built list, so the first rebuild poisoned the process. It now takes one options object with a plugin factory, the rate-limit storage and the client-IP header.

That fix exposed a second coupling the stale objects had been masking: the self-service scope ceiling is read at plugin construction, and the test app installs a permissions snapshot narrowed per file, so a rebuild permanently lost `mcp:*` from the ceiling. The preload thunk restores the production invariant.

## Deploying

One migration (`0057`), no new environment variable.

**What orders it.** `applyCoreMigrations()` runs before `createAuth()` inside `bootCritical()`, so a process that serves an auth request has already applied the file. The 1.7.3 drizzle-adapter schema check does **not** observe the database: it diffs the TypeScript drizzle objects against the fields the plugins declare and never queries Postgres. The deploy risk is an **old replica still serving after the `DROP COLUMN`s**: it writes `public` and `type` on every client insert and fails with 42703. Roll forward; do not leave both builds live.

**Pre-flight, before the roll:**

```sql
SELECT client_id, name FROM oauth_clients
WHERE token_endpoint_auth_method IS NULL AND "public" = false;
```

A non-zero count is the set of third-party clients that will get `invalid_client` after the roll (behaviour change 2). Fold them to `client_secret_post` by hand, or notify their owners.

Closes #1249.


pierrecabriere:
Merged to main as part of #1287 (squash 649277242), which carried every commit of this PR plus the post-audit hardening. Nothing left to merge here.

Issue #1275, opened 2026-09-07
RBAC follow-up: recreate org_role enum without 'viewer' (migration 0058)


pierrecabriere:
Migration `0056_space_roles.sql` adds `guest` to the `org_role` enum but cannot drop `viewer` (`ALTER TYPE … DROP VALUE` does not exist). `scripts/migration/0008-org-viewer-to-guest.sql` moves the rows; the enum value stays until a migration recreates the type.

`meta/0056_snapshot.json` and later snapshots deliberately list the enum without `viewer`, so every real database (fresh installs included) diverges from the snapshot until this lands.

Scope of `0058_drop_org_viewer.sql`:
- guard: fail loudly if any `organization_members.role`, `org_invitations.role` or `oauth_clients.signup_role` row still reads `viewer` (name `scripts/migration/0008` in the message);
- recreate `org_role` as `('owner','admin','member','guest')`, swap the columns, drop the old type;
- no data writes.

Reference: `docs/architecture/RBAC_PERMISSIONS_SPEC.md` §12. Introduced by #1260.

PR #1293, opened 2026-09-09
feat(rbac): move the invitations `0008` leaves reading `viewer` (#1275)

pierrecabriere:
First half of #1275. The second half — the migration that recreates `org_role` without `viewer` — is #1291, which stays in draft until the RBAC rollout is deployed (reason below).

## The gap

`scripts/migration/0008-org-viewer-to-guest.sql` restricts its invitation UPDATE to `status = 'pending'`, and `org-viewer-to-guest-migration.test.ts` asserts that an expired one keeps `role = 'viewer'`. That scope is correct on its own terms: a pending row also gets the `space_assignments` snapshot `0008`'s step 5 verifies, while an accepted, expired or cancelled one grants nothing and needs none — widening the WHERE would make `0008` flip rows its own coverage check cannot account for.

What it leaves behind is history reading `viewer`. Nothing reads those rows today, so the gap was invisible — until the type narrowing, which **cannot cast them**: `ALTER COLUMN … TYPE org_role USING role::text::org_role` fails on any installation that has ever had an accepted viewer invitation. That is most of them.

## What this does

`0012-org-invitation-history-viewer-to-guest.sql` maps them to `guest` — the successor `0008` chose for the same offers, so the history reads as the same decision taken against today's vocabulary.

`status <> 'pending'` is **load-bearing, not tidiness**: a pending row swallowed there would become `guest` without the snapshot that makes its acceptance equivalent, and `0008` would no longer match it on a later run. There is a test for exactly that, run in the wrong order on purpose.

One transaction, fenced, idempotent, no INSERT and no DELETE. Rows UNMEASURED — the script prints its own before/after counts, per README requirement 4.

## Where it goes in the rollout

Step 5 of the RBAC runbook, right after `0008`, in the same maintenance window. Doing it there is what lets the narrowing ship in a **later** release against a database this window left clean — drizzle applies the pending batch in one transaction and the row scripts run after it, so the narrowing cannot ride the same release as `0056`.

## Tests

Three cases added to `org-viewer-to-guest-migration.test.ts`, next to the one that asserts what `0008` leaves:

- both non-pending statuses move (`expired` and `cancelled` — the scope is `<> 'pending'`, not a list);
- run **before** `0008`, a pending viewer invitation is untouched, keeps its empty snapshot, and `0008` still finds it and snapshots it afterwards;
- a second run changes nothing.

`TEST_TIER=0 bun test apps/api/test/integration/db/org-viewer-to-guest-migration.test.ts` — 10/10.


pierrecabriere:
Folded into #1291, which already carries this commit as the first of its three — one PR, one release. The `0012` rationale is section 1 of that PR's description.

PR #1294, opened 2026-09-09
feat(cli): sync skills across spaces, one active context per destination, and clean up on logout

pierrecabriere:
## Summary

Skill **distribution** only: the Claude Code plugin (and the shared Codex / `claude-user` directories) can now be fed by several spaces of the pinned organization, and an explicit `appstrate logout` removes what that profile installed.

This carves the distribution half out of #1274. That PR mixes two mechanisms with different lifecycles — one-way distribution `Appstrate → local install`, refreshed by the marketplace `command` source once per session, and an explicit authoring loop `working folder ↔ Appstrate draft` (`pull` / `status` / `push` / `publish`, `POST /api/packages/import?draft=true`, `lock_version`, three new MCP tools). Only the first is here. #1274 keeps the authoring loop and the OAuth DCR scope fix; it should be rebased on this branch once this lands.

No server change, no migration, no new environment variable, no API surface change. `apps/cli/` and `docs/plans/` only.

## The rule this PR commits to

**One active context per destination.** Profiles do not accumulate installations.

| Action | Behaviour |
| --- | --- |
| Sync | installs the **union** of the skills of the selected spaces, deduplicated by package id |
| Switch organization or profile, then sync | replaces the previous context in the requested targets |
| Drop a selected space | removes only the skills no longer in the resulting union |
| Explicit `appstrate logout` | resets the generated plugin to the setup plugin (no business skills, no `.mcp.json`) and removes that profile's managed skills in shared targets |
| Log out **another** profile | current installation untouched |
| Network / auth / preparation failure | previous installation preserved — a technical failure is not a removal request |

The `.mcp.json` header `X-Space-Id` stays on the **pinned** space, independently of `--space` / `syncSpaces` (decision D12). Selecting extra skill sources must not silently move the MCP execution context.

## Space selection: your access IS the selection

Default: **every space this profile is a member of, with `skills:read` there**. `GET /api/spaces` answers per principal (`listSpacesForPrincipal`), but it answers what the caller may *see*, not what it may *use*: a `closed` space an org member has not joined is listed too, with `access: "none"`, so they can ask to be added. Sync reads `access` and `permissions` on each row and keeps only the usable spaces. Being granted a space is what puts its skills on the machine, and losing one is what takes them off — keeping any that another usable space still supplies.

The review of this PR caught the first version of that rule: it used every listed id, so one closed space in the organization made `GET /api/packages/skills` answer 403 `not_a_space_member`, which failed the whole run for every member who had not joined it — and under `--print-path` Claude Code silently kept the stale plugin. The test stub now carries `access` / `role` / `permissions` and answers 403 for a space-scoped read of an unusable space, so the selection is covered rather than passing vacuously.

`--space` and `syncSpaces` **narrow** that set. They never widen it.

```sh
appstrate skills sync --space spc_prod --space spc_team   # narrow this run
```

```toml
# config.toml — a persistent restriction
syncSpaces = ["spc_prod", "spc_team"]
```

`syncSpaces` holds IDs and is cleared when the organization changes, because a space id belongs to one org. Setting `syncSpaces = []` installs no business skills and keeps the MCP connection.

**A lost space is not a failed run.** A configured id the profile can no longer use — gone, or listed without membership / `skills:read` — is dropped with a note saying which; the spaces that ARE usable still sync. An explicit `--space` naming nothing, or naming a space it cannot use, still fails — it was typed just now. A listing that *errors* (503, network) still fails the whole run: that is a fault, not a revocation, and telling the two apart is the point.

The first draft of this PR did none of that: the default was the pinned space alone, and `configuredSpaceId` threw on an unreachable id — so one lost space stopped every other space from syncing, and the lost space's skills stayed installed. The plan described the precedence as a mechanism and never argued for it; it is now **D14**, **D15** and **D16**.

## One copy per package, at the organization's `latest`

A package is an **organization-level** row that `space_packages` attributes to any number of spaces, so the same skill is normally listed by several of the spaces you reach. It is downloaded once, from the first that listed it — the listings run concurrently through `mapWithConcurrency` but merge in selection order, so the choice stays deterministic and the request count stops being serial in the number of spaces.

That choice never decides *which version*: `versions/latest` resolves by dist-tag, and `buildVersionDetailDto` → `getVersionDetail(itemId, versionSpec)` takes no space. `space_packages.version_id` — the version a space pins for its own runs — is therefore **not** what lands on the machine. The code already did this silently; with several spaces it stops being a detail, so it is now D16 and a paragraph in the CLI reference.

## How the replacement is made safe

- `SyncContext` (`profileName`, `instance`, `userId`, `orgId`) is recorded per target, alongside the existing recorded root. The pinned space is **not** part of it: since D14 it selects no skill, it only fills the `X-Space-Id` header of `.mcp.json`, which `pluginTreeMatches` already compares by content. So `appstrate space switch` rewrites the header at the next sync without being an installation switch — a package that persistently fails to resolve no longer blocks it. A pin that moves *during* a run still aborts the swap (checked in `validate()`), because the `.mcp.json` bytes were computed before the downloads.
- `contextChanged` is computed only when a ledger for the current root already exists, so a fresh install is not treated as a switch.
- On a context switch the run refuses to proceed on partial data: an unresolved catalogue entry or a missing downloaded tree aborts before any swap, and `writePluginTree` gets `strict` so an incomplete plugin is never committed.
- A `validate()` callback re-reads the active profile inside the staging swap; if the context or `syncSpaces` moved under the run, it throws instead of installing a mixed tree.
- Failed removals now **retain** their ledger ownership instead of dropping it, so cleanup can be retried rather than stranding the directory as unmanaged.
- `removeManagedDir` no longer swallows every `lstat` error — only `ENOENT` means "already gone".

## Logout

`cleanupProfileSkills` runs under the same `flock(2)` sync lock, so a concurrent sync cannot reinstall the files behind the cleanup. Credentials are wiped even when the remote revoke or the skill cleanup fails — a cleanup failure must never leave the user signed in. Failures are reported with the exact retry command.

A ledger names its context. One that does not is refused whole — see below — so no code path has to decide whose directories those are.

Cleanup changes the generated source immediately; Claude's installed cache is separate. `claude plugin update appstrate@appstrate` plus a fresh session applies it. This is documented, not promised away.

Also fixed here: `login` no longer preserves the pinned org/space when the same `userId` logs in against a **different instance**.

## One form written, the same one read

`TargetState.context` is **required**, and `isTargetState` refuses a ledger without it. An older file therefore claims nothing and goes through the existing "Sync state could not be used and has been ignored" path: the run re-materializes and the next successful sync owns the tree again. No dual-read path, no `legacy` flag, no three-valued per-entry context — `docs/NO_TRANSITIONAL_CODE.md` §1 and §4.

The first draft did carry that shape, and it bought a consequence nobody chose: a context-less ledger compared unequal to the current context, so the **first sync after upgrading** ran all-or-nothing, and a skill erroring persistently kept it there permanently — the context is only stamped once a run completes, so an aborted run never escapes. Removing the legacy shape removes that with it.

**The cost, stated plainly.** A ledger from `v1.0.0-beta.56` or `.57` — the two releases that shipped `skills sync` — is refused. `claude-plugin`, the default and the marketplace path, is regenerated whole and costs nothing. Skills already synced into `codex` or `claude-user` are reported unmanaged once, with the existing remedy on stderr, and are removed by hand before sync owns them again.

## Documentation drift corrected in the same pass

`apps/cli/README.md` still documented `appstrate logout --all` (the flag no longer exists), described the sync lock as pid-based (it has been `flock(2)` since the bun:ffi lock landed), and showed `space switch <id-or-slug>`.

## Tests

`apps/cli/test/skills-multi-context.test.ts` is new (multi-space isolation, failure recovery); `skills-command.test.ts`, `login-org.test.ts`, `logout.test.ts` and `config.test.ts` extended. Covered: duplicates across spaces, dropping a space, a visible-but-unjoined closed space (default, `syncSpaces`, `--space`, pinned), a member role without `skills:read`, context switch, space switch over an unresolvable skill, pin moved mid-run, offline logout, unmanaged files, legacy-ledger migration, sync/logout concurrency.

```
TEST_TIER=0 bun test  (from apps/cli)  →  1494 pass, 0 fail
bun run check                          →  40/40
```

The legacy-ledger test became a test of the new contract: a ledger naming no context is refused, the run says so, and one sync re-establishes ownership.

## Out of scope

Working folders, `pull` / `status` / `push` / `publish`, draft import, `lock_version`, the new MCP tools, authoring for agents / integrations / MCP servers, and the OAuth DCR scope fix — all stay in #1274.

Still open, unchanged from the plan: keychain access from a non-interactive background process on macOS, and a ledger read under a genuinely different `HOME`, are unit-tested only.


pierrecabriere:
Post-review pass, one commit (`5b157f767`).

The ledger's optional `context` and its three-valued per-entry twin (`SyncContext | null | undefined`) were a dual-read path plus a retirement guard with no recorded expiry — `docs/NO_TRANSITIONAL_CODE.md` §1 and §4. `TargetState.context` is now required and `isTargetState` refuses a ledger without one, which routes an older file through the existing *ignored, re-materialize everything* path.

That deletes `ManagedSkill.context`, the provenance ternary in `carry()`, and cleanup's `legacy` flag and warning — and with them a consequence nobody chose: a context-less ledger compared unequal to the current context, so the first sync after upgrading ran all-or-nothing, and a persistently erroring skill kept it there for good, since the context is only stamped once a run completes.

Cost is in the CLI reference: a `beta.56`/`.57` ledger is refused; `claude-plugin` regenerates whole at no cost, `codex` / `claude-user` directories are reported unmanaged once and removed by hand.

Also from the same review — `CONTEXT_KEYS` derived from `SyncContext` with `satisfies` so `isContext` and `sameContext` cannot fall behind a field; `selectedSpaces` split in two (the `if (explicit)` branch was invariant across the map, and `resolveSpaceRef` is ID-only, so where it was reached it could only throw); `syncSpaces` trimmed and deduplicated at the config boundary.

`bun test` 1480 pass / 0 fail, `bun run check` 40/40.

pierrecabriere:
Second pass (`877ee2ed3`): **access is the selection**.

The first draft made a grant mean nothing on its own — a space had to be added to `syncSpaces` or `--space` before its skills reached the machine — and made a revocation worse than a no-op: `configuredSpaceId` threw on an id the profile no longer reached, so one lost space stopped every *other* space from syncing, and the lost space's skills stayed installed.

`GET /api/spaces` already answers per principal (`listSpacesForPrincipal`), so the grant is the whole mechanism. The default is now every space it returns; `--space` and `syncSpaces` narrow that set and never widen it. Joining a space installs its skills at the next sync; leaving one removes them, keeping any another reachable space still supplies.

A configured id the profile no longer reaches is dropped with a note instead of failing the run — losing access is a decision made elsewhere. An explicit `--space` naming nothing still fails, and a listing that *errors* (503, network) still fails the whole run: a fault is not a revocation.

Two things the review surfaced that were true but unwritten, now decisions in the plan:

- **D16** — a package is an org row that `space_packages` attributes to spaces, and `versions/latest` resolves by dist-tag with no space argument (`getVersionDetail(itemId, versionSpec)`). So the machine gets the org's `latest`, never the version `space_packages.version_id` pins for that space's runs. Harmless with one space, a real semantic with several — it is now in the CLI reference.
- The per-space listings were serial. They now run through `mapWithConcurrency` and merge in selection order, so dedup stays deterministic while the request count stops scaling serially with the number of spaces.

Three new tests: union of reachable spaces with a shared package downloaded once, grant then revocation, and `syncSpaces` narrowing with an unreachable id skipped. The stub gained `/api/spaces`, and its helper now separates the two halves the real API keeps apart — `space_packages` attribution versus what the principal reaches.

`bun test` 1483 pass / 0 fail, `bun run check` 40/40.

pierrecabriere:
Review follow-up — three commits pushed on top of c585acd31:

- **160568a24** `fix(cli): select only spaces the profile is a member of with skills:read` — `GET /api/spaces` also lists `closed` spaces an org member has not joined (`access: "none"`, "so they can ask"). The first version of D14 took every listed id as a source, so `GET /api/packages/skills` with `X-Space-Id` on such a space answered 403 `not_a_space_member` and failed the whole run — under `--print-path` Claude Code then kept the stale plugin. Selection now reads `access` and `permissions` on each row; a `syncSpaces` id that is listed but unusable is skipped with a note naming why, an explicit `--space` on one fails, the pinned-space MCP warning fires on `access: "none"` too. Test stub carries the three fields and refuses space-scoped reads of an unusable space, so the six new tests fail without the fix.
- **31989bb51** `refactor(cli): keep the pinned space out of the sync context` — the pin selects no skill since D14, it is `.mcp.json` content that `pluginTreeMatches` compares by content; keeping it in `SyncContext` made `space switch` an all-or-nothing installation switch that any persistently unresolvable package could block. `validate()` still aborts a swap when the pin moved mid-run. `isContext` requires the exact key set (no dual-read for the earlier shape).
- **28d2972a3** stub: real preset name.

`bun run check` 40/40, `TEST_TIER=0 bun test` from `apps/cli` 1494 pass / 0 fail. PR body updated accordingly.

Issue #1343, opened 2026-09-10
rbac: a custom bundle reports permissions it does not grant, and cannot be edited as stored


pierrecabriere:
## Summary

A custom bundle reports permission strings it does not actually grant, and any `PATCH` that resends the stored array is refused — so such a bundle is uneditable as-is, and an admin and a holder see different answers with no signal.

## Mechanism

- Write/read asymmetry: `toWire` returns `row.permissions` raw (`apps/api/src/services/space-roles.ts:82`), while resolution filters through `knownSpaceLevelPermissions()` (`apps/api/src/lib/space-role.ts:63`).
- So a bundle `["agents:read", "ghostmodule:write"]` stored while module M was loaded, read back without it: `GET /api/roles` still lists the ghost.

## Executed

```
GET   /api/roles                                → lists "ghostmodule:write"
PATCH /api/roles/:id  { name: "x" }             → 200 (ghost survives)
PATCH /api/roles/:id  { permissions: [ …exact stored array… ] }
                                                → 400 Unknown space-level permission
```

Reachable across a heterogeneous fleet where replicas disagree on `MODULES`, and after any module removal.

## Second-order effect

A bundle whose strings all become unknown still resolves to a non-null `SpaceRoleRef`, so `applySpacePermissions` does not raise `not_a_space_member`. The holder becomes "a member with zero permissions" — which in an **open** space is strictly *below* the implicit default they would have had with no row at all.

## Proposed fix

Make the wire projection agree with resolution — filter on read, or (better) reject unknown strings on write and reconcile stored rows when a module is unloaded. And decide what a fully-unknown bundle means: today it silently demotes below the open-space default.

Issue #1373, opened 2026-09-10
licence boundary: the generated SPA schema, the removal recipe and the isolation gate each leak past it


pierrecabriere:
## Summary

Three gaps let commercial-module content or dependencies cross the Apache-2.0 boundary without any gate noticing. None is a licence violation today; all three make the boundary weaker than the gates claim.

## 1. The generated SPA schema commits the commercial wire contract

`apps/web/src/api/schema.d.ts` is the one file exempted from the SPDX header gate, and it byte-commits `@appstrate/module-ee`'s entire wire contract. `scripts/lib/module-openapi.ts:56` discovers modules from the **filesystem**, not from `MODULES`, so the EE paths are always merged — in contrast to `baseline.json`, which is stripped.

## 2. The README's removal recipe is incomplete

`packages/module-ee/README.md:36-43` describes removing the module from a redistribution. Following it exactly leaves `apps/web` typecheck red: five Apache-2.0 web files dereference `Ee*` schema names — `use-billing.ts:15`, `:38`, `billing-contact.ts:17`, `billing.tsx:31`.

## 3. `verify-module-isolation` does not scan test trees

`scripts/verify-module-isolation.ts:350-356` scans no test tree outside `scripts/`, so a static commercial import from `apps/api/test/**` passes every gate. Already latent: `model-selection.test.ts:39` does exactly this with `module-claude-code`.

## Verified correct alongside

`bun run verify:license-boundary` passes (2510 files checked, 2420 Apache-2.0, 90 commercial, 1 exempt), `verify-module-isolation` passes, all 90 EE `.ts` files carry the SPDX header including tests and `drizzle/`, EE imports nothing deeper than `@appstrate/core/*`, and the zero-footprint suite is green (14/14).

## Proposed fix

- Strip EE paths from the generated `schema.d.ts` the way `baseline.json` is stripped, or gate the generator on `MODULES`.
- Make the README recipe complete, or make those five web files degrade rather than fail to typecheck.
- Extend `verify-module-isolation` to the test trees.

Issue #1313, opened 2026-09-10
security(rbac): an end-user token reads locked input values from any space in the org, private ones included


pierrecabriere:
## Summary

An OIDC end-user token pinned to space A can read `run-config` — including **locked input values** — for any space in the same organization, `private` spaces included. Two defects compose to produce it.

**This is live in production today.** The route predates `v1.0.0-beta.57` (verified with `git ls-tree v1.0.0-beta.57`); it was found while auditing the unreleased window, not introduced by it.

## Mechanism

`requireSpaceFromParam` (`apps/api/src/routes/spaces.ts:188`) proves only that the path space belongs to the org, then calls `applySpacePermissions`:

```ts
const space = await validateSpaceInOrg(spaceId, c.get("orgId"));
if (!space) throw notFound(...);
await applySpacePermissions(c, space);
```

`applySpacePermissions` (`apps/api/src/middleware/space-context.ts:70`) returns immediately for a credential with no org role:

```ts
if (!c.get("orgRole")) return;
```

It never compares the path space to the credential's **pinned** `spaceId` — the check `requireSpaceContext` performs (`space-context.ts:126-128`), and that `apiKeySpaceScopeGuard` performs for API keys only (`middleware/guards.ts:104` passes through when `authMethod !== "api_key"`).

An OIDC end-user token carries `orgId`, a pinned `spaceId`, **no `orgRole`**, and `permissions` written straight from its scope claim (`lib/auth-pipeline.ts:154-156`). That claim legitimately contains space-level strings, `agents:read` among them (`apps/api/src/modules/oidc/auth/scopes.ts:43`).

The second half: `apps/api/src/routes/agent-detail-handler.ts:189-193` emits `input.values` and `input.locked_fields` **unconditionally** — its own comment says so. `locked_fields` exists precisely so the launcher cannot *write* those fields; the projection lets a caller *read* them.

## Failure scenario

End-user minted for space A, org O, scope `openid agents:read runs:read`, header `X-Space-Id: <A>`, where **B is `private`**:

```
GET /api/spaces/<B>/packages/@secret/bot/run-config
→ 200
{"input":{"values":{"api_token":"SECRET-FROM-SPACE-B"},
          "locked_fields":["api_token"]},
 "modelId":…, "proxyId":…, "version_pin":…}
```

`run-config` is the one route under that `router.use` block whose permission string is end-user-grantable; its siblings need `spaces:read` / `agents:configure` and are safe **by accident**.

## Proposed fix

[...]

PR #1376, opened 2026-09-10
fix(roles): bound and deduplicate a custom role's permission array

### LINKED ISSUE #1342: rbac: space_roles.permissions is unbounded and un-deduplicated, and is walked on every request
## Summary

`space_roles.permissions` is validated as `z.array(z.string().min(1))` with no `.max()` and no dedupe, and the array is walked on **every request** of every holder.

## Mechanism

`apps/api/src/routes/roles.ts:43` and `:52`. `spacePermissions()` (`apps/api/src/lib/space-role.ts:68`) walks the array per request; `loadSpaceMemberships()` walks it per space in `GET /api/spaces`.

Only the JSON body limit bounds it.

## Executed

```
POST /api/roles  { permissions: [ …20 000 entries… ] }  → 201
```

Duplicates are stored verbatim.

## Proposed fix

Cap the array (the known vocabulary is small — `knownSpaceLevelPermissions()` gives the real upper bound), dedupe on write, and reject unknown strings on write rather than filtering them at read time. See the companion issue on the read/write asymmetry, which the same change resolves.

pierrecabriere:
Closes #1342

## Root cause

`space_roles.permissions` was validated as `z.array(z.string().min(1))` with no length bound (`apps/api/src/routes/roles.ts:43`, `:52`) and stored verbatim by `createSpaceRole`/`updateSpaceRole` (`apps/api/src/services/space-roles.ts`). `assertKnownPermissions` already refused an *unknown* string, but nothing refused the same *known* string repeated: `{ permissions: [ …20 000 × "agents:read"… ] }` was a 201, and the stored array is then re-walked on every request of every holder — `spacePermissions()` (`apps/api/src/lib/space-role.ts:68`) and once per space in `loadSpaceMemberships()` for `GET /api/spaces`.

## Fix

`assertKnownPermissions` becomes `normalizePermissions(permissions): string[]`, the single write-time normaliser both `createSpaceRole` and `updateSpaceRole` now store the result of:

- **Dedupe** — the walk accumulates into a `Set`, so a bundle stores each permission once. This is what makes the stored array bounded *by construction*: every entry must be in the vocabulary, and no entry appears twice, so `permissions.length ≤ knownSpaceLevelPermissions().size` forever.
- **Ceiling** — a body longer than the whole vocabulary can hold nothing but duplicates and is refused (400, `param: "permissions"`, naming both numbers) *before* it is walked.
- Unknown-string refusal and the empty-list refusal are unchanged in behaviour and in message.

The bound is **derived from `knownSpaceLevelPermissions()`, not a chosen constant**: modules contribute space-level permissions at boot, so a hand-picked `.max(N)` in the Zod schema would be both a second (wronger) bound and a cap that could refuse a legitimate "grant everything" role on a module-heavy deployment. Keeping it in the service — where the vocabulary is already loaded — leaves exactly one bound, exact under any `MODULES`. No route contract change, so no OpenAPI/api-types churn.

**No data repair script.** Custom roles shipped with #1260 and are not deployed, and unknown strings were already refused at write, so there is no stored array to normalise retroactively.

## Tests

`apps/api/test/integration/routes/roles.test.ts` — one new test, `stores each permission once, and refuses a list longer than the vocabulary`, paired as the file's discipline requires: an array of `size` duplicates → 201 storing `["agents:read"]`, and `size + 1` → 400 naming the size. It reads `size` from `GET /api/roles/vocabulary` rather than restating a number.

```
  apps/api/test/integration/routes/roles.test.ts \
  apps/api/test/integration/routes/space-member-management.test.ts \
  apps/api/test/unit/permissions.test.ts
→ 55 pass, 0 fail (335 expect() calls)

bunx tsc --noEmit -p apps/api → clean
bunx eslint + prettier on both touched files → clean
```

Negative control: with the ceiling short-circuited and the `Set` bypassed, the new test fails on both halves (48 stored duplicates, no 400) — it discriminates.

## Notes for the orchestrator

- **Overlap with #1343** (custom bundle reports permissions it does not grant / cannot be edited as stored): none in the diff. #1343 owns the read/write asymmetry — `toWire` reporting the stored array while `spacePermissions()` filters it against the live vocabulary — which survives this change, since a module unloading can still make a stored string unknown. This PR only makes the stored array short and duplicate-free; #1343 can build on `normalizePermissions` if it wants a shared normaliser.
- No migration (platform 0063 stayed unused — a Zod bound plus dedupe at write time was enough, and a DB constraint could not express "one entry per vocabulary member" without pinning the vocabulary into SQL), no env var, no deploy ordering.

### DIFF
--- a/apps/api/src/services/space-roles.ts
+++ b/apps/api/src/services/space-roles.ts
@@ -110,23 +110,46 @@ export async function listSpaceRoles(orgId: string): Promise<SpaceRoleWire[]> {
 }
 
 /**
+ * What a write actually stores: every entry known, each one once.
+ *
  * An unknown permission is a REFUSAL, never a silent drop: a role created with
  * a typo would 403 on the thing its author asked for and say nothing about why.
  * Naming the first offender is enough — the array is authored in a picker.
+ *
+ * The vocabulary is also the array's ceiling, in both directions. Duplicates
+ * are what let a body name three permissions in twenty thousand entries, and
+ * the stored array is re-walked on every request of every holder
+ * (`spacePermissions`, and once per space in `GET /api/spaces`), so they are
+ * collapsed here rather than carried forever; a body longer than the whole
+ * vocabulary can hold nothing but duplicates and is refused before it is
+ * walked. Deriving the bound from `known` rather than picking a number keeps it
+ * exact under any set of loaded modules.
  */
-function assertKnownPermissions(permissions: string[]): void {
-  if (permissions.length === 0) {
-    throw invalidRequest("A role must grant at least one permission", "permissions");
-  }
+function normalizePermissions(permissions: string[]): string[] {
   const known = knownSpaceLevelPermissions();
-  const unknown = permissions.find((p) => !known.has(p));
-  if (unknown !== undefined) {
+  if (permissions.length > known.size) {
     throw invalidRequest(
-      `Unknown space-level permission '${unknown}'. ` +
+      `A role can hold at most ${known.size} permissions — the whole vocabulary — ` +
+        `and this list has ${permissions.length}. ` +
         `See GET /api/roles/vocabulary for the permissions a role can hold.`,
       "permissions",
     );
   }
+  const granted = new Set<string>();
+  for (const permission of permissions) {
+    if (!known.has(permission)) {
+      throw invalidRequest(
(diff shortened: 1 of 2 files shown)


pierrecabriere:

Issue #1331, opened 2026-09-10
ee(stripe): 'as Stripe.Card' renders "undefined/ed" in the card-expiring email, and two casts blind the 22.x relocations


pierrecabriere:
## Summary

`event.data.object` for `customer.source.expiring` is cast to `Stripe.Card`, but in stripe 22.6.1 that object is a `CustomerSource` (`Account | BankAccount | Card | Source`). For a legacy `Source`, the card-expiring email renders `"undefined/ed"` as the expiry.

Two neighbouring casts are unnecessary in 22.6.1 and blind the compiler to exactly the relocations the SDK bump is about.

## Mechanism

`packages/module-ee/src/stripe/webhooks.ts:660`:

```ts
const source = event.data.object as Stripe.Card;
```

`Source` has no `exp_month` / `exp_year`, so the template computes `String(undefined).padStart(2, "0")` + `String(undefined).slice(-2)` → `"undefined"` + `"ed"`.

Also unnecessary in 22.6.1, and worth removing in the same pass:

- `:470` — `(event.data as { previous_attributes?: Record<string, unknown> })`. `CustomerSubscriptionUpdatedEvent.Data.previous_attributes?: Partial<Subscription>` is typed upstream.
- `:531` — `previousAttributes.items as { data?: … }`. Same. This cast erases the compiler's ability to flag a relocation of `items[].price.id`.

## Proposed fix

Narrow on `source.object === "card"` and skip (or handle) the other member types, then drop the two `previous_attributes` casts so the typechecker can see the next relocation.

Issue #1325, opened 2026-09-10
ee(billing): concurrent checkouts create two live subscriptions, and the superseded one is never cancelled


pierrecabriere:
## Summary

`createCheckoutSession` is a read-then-act with no reservation and no lock. Two calls before the first payment both pass the guard and open two Checkout sessions; if both are paid, one subscription is silently ignored by Appstrate and **never cancelled at Stripe**.

The function's own docstring claims the opposite: *"a double charge is the server's to prevent."*

## Mechanism

`packages/module-ee/src/stripe/checkout.ts:26-38`:

```ts
const [account] = await db.select({...}).from(billingAccounts).where(eq(billingAccounts.orgId, orgId));
if (!account) throw noBillingAccount();
if (planAction(account) !== "checkout") throw subscriptionExists();
```

…and then, ~40 lines later, `checkout.ts:79`:

```ts
const session = await getStripe().checkout.sessions.create({ ... });
```

Nothing between the two reserves anything. `planAction` only inspects the subscription already attached to the row.

Downstream, `attachableSubscription` (`webhooks.ts:112`) attaches the first paid subscription and `logSupersededSubscription` (`:120`) logs-and-skips the second — a `logger.info` line, no alert, no cancellation.

## Failure scenario

Reproduced (two sessions created; no real payment made). A user double-clicks, or opens the billing page in two tabs, and pays both. Result: two live Stripe subscriptions, one of which Appstrate does not know about and will never cancel — the customer is charged twice, indefinitely, and the second charge is invisible in the product.

## Proposed fix

Either reserve at session-creation time (write a pending-checkout marker on the row under a conditional update, and refuse or reuse when one is open and unexpired), or reconcile: when `logSupersededSubscription` fires, cancel the superseded subscription at Stripe rather than logging it. The second is worth doing regardless — the current behaviour leaves a paid, live subscription orphaned by design.

Issue #1324, opened 2026-09-10
ee(billing): upgrading Starter → Pro keeps the Starter credit quota, leaving the account blocked


pierrecabriere:
## Summary

`customer.subscription.updated` syncs the plan but never the quota. A Starter account that has spent its 20 000 credits and upgrades to Pro becomes a Pro account still capped — and still blocked — at 20 000.

## Mechanism

`packages/module-ee/src/stripe/webhooks.ts:475-482`:

```ts
const updates: BillingUpdate = {
  subscriptionStatus: subscription.status,
  cancelAtPeriodEnd: subscription.cancel_at_period_end,
  updatedAt: new Date(),
};
if (itemPeriodEnd) updates.periodEnd = itemPeriodEnd;
if (newPlan) updates.planId = newPlan.id;
```

No `creditQuota`. The field is written at six other places in the same file (`:250`, `:267`, `:323`, `:419`, `:439`, `:574`) — this branch is not one of them, and no other path compensates: `changeSubscriptionPlan` relies on this webhook.

## Failure scenario

Reproduced: Starter account at 20 000/20 000 → upgrade to Pro → `planId` becomes `pro`, `creditQuota` stays 20 000, the account stays blocked. The customer is now paying the Pro price for a Starter ceiling, and nothing in the UI explains it.

## Proposed fix

Set `creditQuota: newPlan.creditQuota` alongside `planId` in the same branch. Decide explicitly what a **downgrade** does when consumption already exceeds the lower plan's quota — the same line has to answer that, and the answer belongs in a test.

PR #1377, opened 2026-09-10
fix(ee): refuse to attach a subscription Stripe has already ended

### LINKED ISSUE #1323: ee(billing): a late checkout.session.completed revives a cancelled subscription with full quota
## Summary

A late `checkout.session.completed` delivered after `customer.subscription.deleted` re-attaches the **dead** subscription with `subscriptionStatus: "active"` and the plan's full credit quota. The attach never consults the subscription's current state at Stripe.

## Mechanism

`packages/module-ee/src/stripe/webhooks.ts:112`:

```ts
function attachableSubscription(orgId: string, subscriptionId: string): SQL {
  return and(
    eq(billingAccounts.orgId, orgId),
    or(noHeldSubscription(), eq(billingAccounts.stripeSubscriptionId, subscriptionId)),
  )!;
}
```

Only `customer.subscription.deleted` NULLs `stripeSubscriptionId` — the comment at `:86-90` says so explicitly. So after a cancellation `noHeldSubscription()` is **true**, and the first arm matches for any subscription id, dead ones included.

The handler then writes (`webhooks.ts:246-256`):

```ts
.set({
  planId,
  creditQuota: plan.creditQuota,
  stripeCustomerId: customerId,
  stripeSubscriptionId: subscriptionId,
  subscriptionStatus: "active",
  updatedAt: new Date(),
})
.where(attachableSubscription(orgId, subscriptionId))
```

## Failure scenario

Reproduced: `customer.subscription.deleted`, then late delivery of that subscription's own earlier `checkout.session.completed`. The account gets the same subscription back, an `active` status and 20 000 credits — for a subscription Stripe will never bill again.

## Proposed fix

The attach path must verify the subscription is live before granting entitlements — either by retrieving it from Stripe, or by recording the deletion (a tombstone / `canceled_subscription_ids`) and refusing to re-attach an id that was deleted. `noHeldSubscription()` correctly exists so a cancelled account's *next* paid checkout is not dropped; the fix must keep that while refusing the dead id.

pierrecabriere:
Closes #1323

## Root cause

`packages/module-ee/src/stripe/webhooks.ts` — `attachableSubscription()` (`:112` on main) is the shared attach predicate for `checkout.session.completed` and `invoice.paid`. Its `noHeldSubscription()` arm deliberately matches a cancelled account (only `customer.subscription.deleted` NULLs `stripeSubscriptionId`), so the org's *next* paid checkout is not dropped — but neither handler ever consulted the subscription's live state. A `checkout.session.completed` delivered after the cancellation therefore re-attached the **dead** id with `subscriptionStatus: "active"` and the plan's full `creditQuota` (`:246-256` on main). `invoice.paid` has the identical hole through the same predicate: a paid invoice for a cancelled subscription re-attached it and reset the credit counter.

## Fix

Both attach paths now gate on the subscription retrieved from Stripe — the event payload freezes at emission and Stripe guarantees no delivery order, so only the retrieved object can decide this. One shared predicate `isEndedSubscription()` over the existing `ENDED_SUBSCRIPTION_STATUSES` (`canceled`, `incomplete_expired`) from `src/config.ts`, plus one `logEndedSubscription()` line next to the existing `logSupersededSubscription()`.

- `checkout.session.completed`: the `subscriptions.retrieve` it already made *after* the write (for the confirmation email's period end) moves **before** the write and becomes the gate. Its failure now throws instead of being swallowed — that deletes the event claim so Stripe retries, exactly the reasoning already documented on the `invoice.paid` retrieve; the grant is billing-critical and must not proceed on unverified state. No second Stripe call is added: the email reuses the same object.
- `checkout.session.completed` also records `checkoutSubscription.status` instead of a hardcoded `"active"` — with the authoritative object in hand there is no reason for the row to claim a state Stripe does not hold (a checkout that opens a trial is `trialing`). This mirrors `customer.subscription.created`, which already writes the live status.
- `invoice.paid`: same guard right after org resolution.
- `attachableSubscription()`'s doc comment now states the obligation, so the next caller does not repeat the omission.

No tombstone table: Stripe's live state already carries the cancellation and also covers `incomplete_expired` and cases where the `deleted` event was never received, so extra state would be a second, weaker source of truth.

## Tests

`packages/module-ee/test/integration/services/webhooks.test.ts` — 3 regression tests: checkout on a cancelled subscription leaves the free account untouched; checkout records `trialing` rather than `active`; `invoice.paid` on a cancelled subscription grants nothing.

Negative control: with the guard reverted, exactly those 3 fail (27 pass / 3 fail); with the fix, green.

```
pg-test-lock.sh sh -c '. ./.env && bun test packages/module-ee/test'
  → 434 pass, 26 skip, 0 fail (460 tests, 31 files)
bunx tsc --noEmit -p packages/module-ee   → clean
bunx eslint / prettier on both touched files → clean
```

## Notes for the orchestrator

- No migration, no env var, no OpenAPI change. Deploy ordering unaffected.
- Behaviour change worth a release note: a paid invoice for an already-cancelled subscription no longer re-grants quota. That is the documented anti-abuse stance of the `customer.subscription.deleted` handler ("credits are NOT re-granted… users who need credits again must contact support"), now enforced on the attach side too.
- Overlap: #1331 (`Stripe.Card` cast) lives in the same file but in the `customer.source.expiring` case — no textual overlap. #1324 (`updates` missing `creditQuota` in `customer.subscription.updated`) and #1325 (concurrent checkouts) are untouched.

### DIFF
--- a/packages/module-ee/src/stripe/webhooks.ts
+++ b/packages/module-ee/src/stripe/webhooks.ts
@@ -10,6 +10,7 @@ import { logger } from "../logger.ts";
 import {
   getPlans,
   isPlanId,
+  ENDED_SUBSCRIPTION_STATUSES,
   HELD_SUBSCRIPTION_STATUSES,
   type Plans,
   type PlanDefinition,
@@ -108,12 +109,26 @@ function unattachedAccount(orgId: string): SQL {
 /**
  * The predicate for `checkout.session.completed` and `invoice.paid`: they carry
  * AUTHORITATIVE data and also write the account already carrying that subscription.
+ *
+ * `noHeldSubscription()` deliberately matches a cancelled account, so this predicate alone
+ * does NOT prove the subscription is alive — callers MUST first gate on the live object
+ * from Stripe ({@link isEndedSubscription}), or a late event re-attaches a dead id.
  */
 function attachableSubscription(orgId: string, subscriptionId: string): SQL {
   return and(
     eq(billingAccounts.orgId, orgId),
     or(noHeldSubscription(), eq(billingAccounts.stripeSubscriptionId, subscriptionId)),
   )!;
+}
+
+/**
+ * A subscription Stripe will never bill again. The event payload cannot decide this — it
+ * freezes at emission and Stripe guarantees no delivery order, so a `checkout.session.completed`
+ * delivered after the cancellation still describes a live subscription. Only the retrieved
+ * object does, which is why both attach paths retrieve before granting entitlements.
+ */
+function isEndedSubscription(subscription: Stripe.Subscription): boolean {
+  return ENDED_SUBSCRIPTION_STATUSES.has(subscription.status);
 }
 
 /** One line for an event on a subscription this org is not on — a replacement's tail. */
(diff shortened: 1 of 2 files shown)


pierrecabriere:

Issue #1363, opened 2026-09-10
cli: a malformed syncSpaces value in one profile bricks every command for every profile, logout included


pierrecabriere:
## Summary

A malformed `syncSpaces` value in **one** profile makes `readConfig` throw, which bricks every CLI command for **every** profile — including `logout`, so the user cannot get out of it.

## Mechanism

`apps/cli/src/lib/config.ts:150-156` — `readSyncSpaces` throws from inside `readConfig`'s per-profile loop (`:134`).

`readConfig` is the first line of `login`, `logout`, `whoami`, `token`, `api`, `openapi` and every `org` / `space` subcommand, none of which wraps it in a `try`.

## Failure scenario

```toml
[profiles.work]
syncSpaces = "spc_abc"      # string, not array
```

That is the natural TOML mistake — the flag's own help text says "repeatable". Profile `default` now fails too, and there is no command left that can fix or remove the bad profile.

The comment at `:143-149` argues against *skipping* the row, which is fair. Per-profile failure was the third option and is the one that matters here.

## Proposed fix

Fail the offending profile, not the file: mark it unusable with a clear message naming the profile and the key, and let every other profile load. At minimum, `logout` must remain reachable.

Issue #1362, opened 2026-09-10
cli: losing org membership exits 1 and leaves every synced skill on disk forever


pierrecabriere:
## Summary

Losing **organization** membership is treated as a fault rather than as a revocation: the sync exits 1 and every synced skill stays on disk indefinitely. Space-level revocation is handled correctly; org-level is not.

## Mechanism

`apps/cli/src/commands/skills.ts:601` calls `await listSpaces(profileName)` with no status discrimination, and the throw is caught at `:202-205` → exit 1. A 403 `not a member of this organization` is indistinguishable from a 503 at that call site.

By contrast `suppliesSkills` correctly drops a space the user can no longer read.

## Failure scenario

An offboarded employee's machine keeps every one of that organization's skills indefinitely. Under `--print-path` — the marketplace `command` source — Claude Code discards the failing run each session and keeps serving the stale plugin. `appstrate logout` is the only cleanup, and nothing says so.

## Proposed fix

Discriminate on status: a 401/403 is a revocation — clear the managed skills for that profile and say so — while a 5xx or a network error stays a fault that leaves the tree untouched.

PR #1378, opened 2026-09-10
fix(cli): refuse a space listing that omits the caller's standing

### LINKED ISSUE #1320: cli: syncing against a pre-RBAC server silently deletes every installed skill
## Summary

The skills-sync source selection introduced by #1294 reads two fields that do not exist on a server at `v1.0.0-beta.57`. Against such a server the CLI resolves **zero** source spaces and deletes every synced skill — exit 0, empty stderr, no note.

**This is a release-ordering gate**: the `cli@` tag must not ship before the platform, or the CLI needs a shape/version guard.

## Mechanism

`apps/cli/src/commands/skills.ts:574`:

```ts
const suppliesSkills = (space) =>
  space.access === "member" && space.permissions.includes("skills:read");
```

Both fields reach the wire through `spaceWireForCaller`, which arrived with the granular-RBAC work (#1260, merged 2026-09-08, **not deployed**). Verified: `git show v1.0.0-beta.57:apps/api/src/routes/spaces.ts` serialises through `toSpaceWire`, a bare `createdBy → created_by` rename carrying neither field.

`apiList` (`apps/cli/src/lib/api.ts:389`) ends in `return envelope.data as T[]` — a blind cast, no runtime shape check. So `undefined === "member"` is `false` for every row, `spaceIds` is `[]`, and `diffTarget` (`apps/cli/src/lib/skills-sync/plan.ts:401-408`) moves every ledger slug into `plan.removed`.

## Why nothing catches it

- The pinned-space warning at `skills.ts:608` tests `pinned.access === "none"`, and `undefined !== "none"` — so even that stays quiet.
- Under `--print-path` (the marketplace `command` source) the emptied plugin is installed.
- `apps/cli/test/helpers/skills-server.ts:53` defaults `access` to `"member"`, so no test can see it.

## Failure scenario

A self-hoster's CLI auto-updates before their instance does. Next Claude Code session: every synced skill is gone, silently.

## Proposed fix

Either a served-shape guard (refuse to compute a removal plan when `access`/`permissions` are absent — an old server is not "you lost access to everything"), or a platform version/capability check before the sync, or both. A blind `as T[]` on a cross-version wire boundary is the root cause worth fixing independently.

pierrecabriere:
Closes #1320

## Root cause

`apps/cli/src/lib/spaces.ts:50` returned `apiList<Space>(…)` — a blind cast on a
cross-version wire boundary. `access` / `permissions` arrived with granular space
roles (#1260, merged but not deployed); an instance at `v1.0.0-beta.57` serves
neither. Cast anyway, every row reads `access: undefined`, so
`suppliesSkills` (`apps/cli/src/commands/skills.ts:574`) answers no for all of
them, `spaceIds` is `[]`, and `diffTarget` moves every ledger slug into
`plan.removed`. A CLI that auto-updated ahead of its server deleted every synced
skill with exit 0 and empty stderr. The pinned-space warning stayed quiet for the
same reason (`undefined !== "none"`).

## Fix

One shape check at the boundary that produces `Space[]`: `asSpace` accepts a row
only when it carries the caller's own standing (`access` is `member` / `none`,
`permissions` is a string array) and otherwise throws
`ApiError(500, "…This Appstrate instance is older than the CLI…")`. `SpaceObject`
requires both fields, so a 200 without them is a server breaking its own
contract — the same grade of fault the neighbouring `apiList` envelope check
reports, rendered by `formatError` as one actionable line.

The guard lands before any plan is computed, so `skills sync` fails the run
(exit 1) and touches nothing on disk; under `--print-path` Claude Code discards
the run and keeps the existing plugin. Nothing is dual-read and no version
negotiation is introduced: one current wire shape, a loud refusal for anything
else. `createSpace` is untouched — nothing reads standing off a creation.

`org-command` / `space-command` fixtures modelled that pre-RBAC row (no
`access`, no `permissions`), which is part of why no test could see the bug;
they now carry the real wire.

## Tests

- `apps/cli/test/spaces.test.ts` — `listSpaces` rejects a row without standing,
  and one whose `permissions` are not strings.
- `apps/cli/test/skills-command.test.ts` — end to end: a plugin installed against
  a current server survives a sync against one that omits the fields; the run
  exits 1 instead of emptying it.

```
cd apps/cli && TEST_TIER=0 bun test test/spaces.test.ts test/skills-command.test.ts   → 99 pass / 0 fail
cd apps/cli && TEST_TIER=0 bun test test/space-command.test.ts test/org-command.test.ts \
  test/skills-multi-context.test.ts test/login-org.test.ts                            → 138 pass / 0 fail
bunx tsc --noEmit -p apps/cli                                                          → clean
bunx eslint / prettier on the five touched files                                       → clean
```

Negative control: with the guard reverted, all three new tests fail — the
end-to-end one because `skillsSyncCommand` resolves instead of exiting 1.

## Notes for the orchestrator

- No migration, no env var, no OpenAPI change (the server already serves both
  fields; only the CLI's reading of them changed).
- Release ordering the issue asked about is now enforced by the CLI itself: a
  `cli@` tag shipping ahead of the platform fails loudly instead of deleting
  skills. It does mean every `listSpaces` caller (`login`, `org switch`,
  `space list/switch`, `skills sync`) refuses a pre-RBAC instance — intended,
  and the message says what to do.
- Overlap: none in `skills.ts`. #1362 and #1363 own `selectedSpaces` /
  `syncSpaces`; this PR stays in `lib/spaces.ts`. The only shared file is
  `apps/cli/test/skills-command.test.ts`, where one `it` block was added inside
  the existing `skills sync — multiple spaces` describe.


pierrecabriere:

Issue #1370, opened 2026-09-10
ee(billing): re-enabling the module after a disabled window back-bills the whole gap and leaves orgs inconsistent


pierrecabriere:
## Summary

Nothing guards cursor staleness when the EE module is re-enabled after a period with it disabled. The sweeper resumes from the old watermark and back-bills the whole gap against **current** quotas, while orgs created and deleted during the gap are left inconsistent.

## Mechanism

`packages/module-ee/src/index.ts` `init()` seeds or reads the cursor with no age check.

Three compounding effects across a disabled window:

1. **Back-billing.** The sweeper claims every `llm_usage` row accumulated in the gap and debits it against today's quotas — firing soft-cap overshoot and quota-warning emails fleet-wide.
2. **Missing accounts.** `onOrgCreate` never fired, so every org created in the window has no `ee_billing_accounts` row and becomes a per-pass orphaned-org `error` needing a manual `repair:account`.
3. **Permanently unbilled usage.** `onOrgDelete` never drained, so orgs deleted in the window had their `llm_usage` cascade away unbilled.

`packages/module-ee/README.md`'s account of enable → disable → re-enable does not mention any of this.

## Proposed fix

At `init()`, compare the cursor's watermark age against a threshold and refuse to resume silently past it — require an explicit operator decision (resume, or fast-forward and forgive). Reconcile missing `ee_billing_accounts` rows on the same pass rather than emitting an error per org. And document the three effects in the README's enable/disable section.

adityawaghamare04:
### 🔀 Pull Request Created
I have opened a Pull Request with the verified solution patch: [https://github.com/appstrate/appstrate/pull/1375](https://github.com/appstrate/appstrate/pull/1375)

### Proposed Solution & Patch
## 🛠️ Proposed Solution (by Aditya Waghamare)

### Analysis
When the EE billing module is re-enabled after being disabled, the sweeper resumes from the stale watermark without checking its age. This causes three critical issues:
1. Massive back-billing of accumulated usage against current quotas, triggering false alarms and overage alerts.
2. Missing billing account rows for orgs created during the window, resulting in orphaned org errors.
3. Permanently unbilled usage for orgs deleted during the window.

The fix involves adding a staleness check during `init()` in `packages/module-ee/src/index.ts`, handling missing accounts reconciliation, and updating the README documentation.

### Fix & Implementation

```typescript
// packages/module-ee/src/index.ts
import { config } from '@appstrate/config';
[...]

Issue #1330, opened 2026-09-10
ee(billing): cost_credits is an int4 whose overflow throws inside the sweep transaction, wedging billing fleet-wide


pierrecabriere:
## Summary

`cost_credits` is an `integer` while `cost_usd` is `numeric(24,12)`, and the `unattributed` bucket is documented as growing forever. An int4 overflow throws **inside** the sweep transaction, so one organization's bucket wedges billing for every tenant, on every tick.

## Mechanism

`packages/module-ee/drizzle/schema.ts:100` declares `cost_credits` as `integer`. The credit computation is `round((cost_usd + delta) * 1000)::int`.

int4 tops out at 2 147 483 647, i.e. ~$2.15M cumulative for a single context. The `unattributed` bucket has no reset — it accumulates for the lifetime of the deployment.

## Failure scenario

Long-horizon, but the failure mode is what makes it worth fixing early: the throw lands inside the sweep transaction, so it is not one org's problem. Every tick fails, no ledger row is claimed, and billing stops fleet-wide until someone finds the overflowing row.

## Proposed fix

Widen `cost_credits` to `bigint` (or `numeric`), matching `cost_usd`'s precision posture. A migration on a column that has not yet overflowed is cheap; the same migration after the fact is an incident.

Issue #1329, opened 2026-09-10
ee(billing): repair:account is not transactional, and its guard makes an interrupted repair permanent


pierrecabriere:
## Summary

`repair:account` provisions the billing account and applies the recorded debt as two separate statements with no transaction, and its own idempotency guard makes any gap permanent.

## Mechanism

`packages/module-ee/src/billing/repair-account.ts:70-88` — `provisionBillingAccount(...)` commits, then an UPDATE applies the reconstructed debt. The guard at `:65-68` returns `already_provisioned` when an account row exists.

## Failure scenario

`bun run repair:account -- <orgId> <email>` is interrupted after the provision commits and before the UPDATE — Ctrl-C, a pool blip, a pod eviction. A re-run returns `already_provisioned` and writes nothing. The org's entire orphaned debt is silently written off, **and the only tool for it now refuses to touch the org**.

## Proposed fix

Wrap both statements in one transaction. Separately, the guard should distinguish "provisioned and settled" from "provisioned but debt not applied" so a re-run can complete an interrupted repair instead of declaring success.

Issue #1328, opened 2026-09-10
ee(billing): EE_RECONCILIATION_BATCH_SIZE above ~800 is dead code — the drain loop stops and its warning never fires


pierrecabriere:
## Summary

`EE_RECONCILIATION_BATCH_SIZE` above ~800 is dead: the replay window now shares the same 1000-row platform ceiling, so the within-tick drain loop stops iterating entirely *and* the warning that would have told the operator never fires.

## Mechanism

`packages/module-ee/src/billing/usage-recorder.ts:619` computes `limit = min(replaySpan + batchSize, 1000)`, and the ceiling is declared at `packages/module-ee/src/env.ts:31`.

At the default replay window of 200, forward capacity is ~800. So `result.processed >= batchSize` (`billing-sweeper.ts:305`) is unreachable for any batch size above ~800.

## Failure scenario

An operator raises the batch size to 1000 to clear a backlog. Throughput *falls* — from up to 50 batches per tick to one ~800-row pass — and `backlogRemains` becomes permanently false, so the "billing sweep hit the drain cap" warning that would have surfaced the problem never fires.

This is precisely the dead-code condition the comment at `env.ts:28-31` claims the 1000 bound prevents. The bound was correct **before** the replay window and was not re-derived when it landed.

Every sweeper test uses batch size 1, 4 or 100.

## Proposed fix

Derive the batch ceiling from `1000 - replayWindow` rather than from 1000, and add a test at a batch size in the dead range.

PR #1379, opened 2026-09-10
fix(module-ee): keep the maintenance tick running when metering is paused

### LINKED ISSUE #1326: ee(billing): EE_RECONCILIATION_INTERVAL_SECONDS=0 also disables cancellation retries, so deleted orgs keep being charged
## Summary

`EE_RECONCILIATION_INTERVAL_SECONDS=0` is documented everywhere as "disables the sweeper", but the early return also kills `retryPendingCancellations()` and `maybeResyncEntitlements()`, neither of which is metering. A deleted organization whose Stripe cancellation failed transiently then keeps being charged, indefinitely.

## Mechanism

`packages/module-ee/src/billing/billing-sweeper.ts:110-115`:

```ts
const intervalSec = env.EE_RECONCILIATION_INTERVAL_SECONDS;
if (intervalSec === 0) {
  logger.info("billing sweeper disabled (EE_RECONCILIATION_INTERVAL_SECONDS=0)");
  return;
}
```

The tick is the **only** caller of `retryPendingCancellations()` (`billing-sweeper.ts:269` → `org-cancellation.ts:94`) and of `maybeResyncEntitlements()` (`:263`).

Documentation that says otherwise: `.env.example:536`, `docs/ENV.md:77`.

## Failure scenario

During an incident an operator sets the interval to 0, reading it as "pause the billing sweep". An org is deleted; the Stripe cancel call fails transiently, so `org-cancellation.ts` stamps `cancel_requested_at` and leaves the row *"for the sweeper to retry"*. No tick ever runs. The customer's card keeps being charged for an organization that no longer exists.

## Proposed fix

Move `retryPendingCancellations` and `maybeResyncEntitlements` out of the interval guard — they need their own cadence, or their own flag. At minimum, the variable's name and documentation must stop promising that it only touches the sweep.

pierrecabriere:
Closes #1326

## Root cause

`packages/module-ee/src/billing/billing-sweeper.ts:113-116` returned before arming any
timer when `EE_RECONCILIATION_INTERVAL_SECONDS=0`. The timer's tick is the **only** caller
of `retryPendingCancellations()` (`org-cancellation.ts:94`) and of `maybeResyncEntitlements()`,
neither of which is metering. `onOrgDelete` deliberately keeps the billing row on a failed
Stripe cancel *"for the sweeper to retry"* — with metering paused that retry never happened,
so a customer kept being charged for an organization that no longer exists. The docs
(`.env.example`, `docs/ENV.md`) promised the variable only touched the sweep.

## Fix

Split the tick rather than overload the knob:

- `runMaintenance()` now holds the half that is not metering — the throttled entitlement
  reconcile (started, not awaited) and the pending-cancellation retry.
- `startBillingSweeper()` arms `runBillingSweepTick` at the configured cadence, or
  `runMaintenance` at `MAINTENANCE_INTERVAL_SECONDS` (300 s, the default sweep cadence) when
  the interval is `0`. `scheduleNext()` takes the tick it schedules, so the `0` is read in
  exactly one place and the tick functions stay free of the knob — `runBillingSweepTick()`
  still sweeps unconditionally when called directly by a test or an ops job.
- `inFlightSweep` → `inFlightTick`: it can now hold a maintenance-only pass, and
  `drainBillingSweeper()` awaits it at shutdown exactly as before.
- Docs corrected in `.env.example`, `docs/ENV.md`, `src/env.ts` and `src/index.ts`: `0`
  pauses **metering**, not the module's periodic worker.

No behaviour change when the sweeper is enabled: the tick does the same work in the same
order.

## Tests

- `packages/module-ee/test/integration/services/billing-sweeper.test.ts` — new case
  *"keeps a maintenance tick armed when metering is paused (INTERVAL=0)"*: with the interval
  at `0`, a second `startBillingSweeper()` hits the re-entry guard, which fires only when a
  timer is actually armed. Negative control: reverting the `scheduleNext` call in the
  `intervalSec === 0` branch fails it (`0 pass / 1 fail`).
- `packages/module-ee/test/helpers/setup.ts` — `useEeTestSeams()` disarms the tick the
  harness arms at `init()`. Under `EE_RECONCILIATION_INTERVAL_SECONDS=0` a timer now exists
  in tests, and a maintenance pass firing mid-suite would clear pending cancellations a test
  seeded.

```
pg-test-lock.sh … bun test packages/module-ee/test/integration/services/{billing-sweeper,post-signup,storage-entitlement}.test.ts
→ 104 pass, 0 fail, 318 expect() calls
bunx tsc --noEmit -p packages/module-ee → clean
bun run verify:env-docs / verify:compose-defaults / verify:license-boundary → all green
bunx eslint + prettier on the touched files → clean
```

## Notes for the orchestrator

- No migration, no new env var, no OpenAPI surface. Deploy ordering unchanged.
- **Behaviour change for operators**: with `EE_RECONCILIATION_INTERVAL_SECONDS=0` the module
  now runs a 300 s maintenance tick instead of nothing. Both jobs it runs are idempotent and
  cheap (one indexed `SELECT` returning zero rows in steady state; the entitlement reconcile
  is throttled to 1 h internally). If Stripe is the incident that made the operator pause
  metering, the retries fail and log — the same fail-and-retry path as normal operation.
- **Overlap with #1370** (re-enable after a disabled window back-bills the gap): same
  variable, different code. That issue is about cursor semantics in `usage-recorder.ts`; this
  PR only touches the scheduler in `billing-sweeper.ts`. The wording "disabled window" in
  #1370 now means "metering paused" — worth a rebase read, not a conflict.
- No overlap with #1328 (drain loop in `runBillingSweep`), #1330, #1329.


pierrecabriere:

Issue #1352, opened 2026-09-10
connect: two post-burn calls sit outside the try that #1302 added


pierrecabriere:
## Summary

#1302 wrapped the hosted connect start handler in a `try` to stop raw `application/problem+json` reaching the popup, but two calls that run **after** the single-use JTI is burned sit outside it.

## Mechanism

`apps/api/src/routes/integrations.ts:904` consumes the JTI. `:911-921` then runs `getCurrentScopesGranted` (a raw `db.select`) and `resolveStrategy` — both **before** the `try` opened at `:945`.

A DB blip in that window still produces the raw problem JSON in the popup, plus a burned link the user cannot retry.

## Proposed fix

Open the `try` at `:910`.

Issue #1346, opened 2026-09-10
connect: the portal's error completion is context-less and hijacks unrelated connect cards


pierrecabriere:
## Summary

The hosted connect portal's error completion passes `{}` as its detail although `claims.package_id` is in hand two lines above. A context-less completion matches **every** waiting connect surface, so a failing Gmail connect flips an open ClickUp card into an error naming the Gmail package.

Introduced by #1302, not yet deployed.

## Mechanism

`packages/core/src/connect-handshake.ts:233`:

```ts
return !detail.state && !detail.packageId;   // context-less → matches everything
```

…and both carriers fan out (`BroadcastChannel` + `postMessage`). This is exactly the bug `packageId` was introduced to stop (`connect-handshake.ts:190-196`).

`apps/api/src/routes/integrations.ts:968` passes `{}`. Every other caller passes context — `:426`, `:450`, `:454`, `:487`, `:490` all pass `{ state }` or `{ state, packageId }`.

## Proposed fix

One line:

```ts
return c.html(popupHtmlError(err.message, { packageId: claims.package_id }), …);
```

PR #1380, opened 2026-09-10
fix(connect): stop rendering error messages verbatim on session-less connect popups

### LINKED ISSUE #1345: security(connect): the hosted portal renders err.message verbatim to an unauthenticated caller
## Summary

The hosted connect portal renders `err.message` verbatim to an unauthenticated caller. One reachable message discloses an internal row id, an environment variable name, and the fact that a key-rotation / decrypt failure is in progress.

Introduced by #1302 (pre-commit this rendered as a generic 502), not yet deployed.

## Mechanism

`apps/api/src/routes/integrations.ts:968`:

```ts
return c.html(popupHtmlError(err.message, {}), err.status as ContentfulStatusCode);
```

Reachable message, from `apps/api/src/services/integration-connections.ts:977-982`:

> `OAuth client '<uuid>' for '<pkg>' auth '<key>' holds a client_secret that cannot be decrypted … Restore the CONNECTION_ENCRYPTION_KEY this client was registered with…`

This is served to any holder of a 10-minute connect link, on a route with no session. Third-party AS `error_description` text is also echoed verbatim (escaped, so not XSS).

## Proposed fix

Render a generic message to the popup and keep the detail in the log — or split `ApiError` into a public detail and an operator detail on this path. The commit's goal (stop showing raw `application/problem+json` in the popup) is achievable without forwarding the internal message.

pierrecabriere:
Closes #1345

## Root cause

`GET /api/integrations/connect/start` is session-less and its link is handed to
end users outside the org, but `apps/api/src/routes/integrations.ts:1006`
rendered the raw `ApiError.message` from `strategy.begin` into the popup. That
`detail` is written for an operator, so three reachable messages disclosed
internals:

- `integration-connections.ts:977` — `OAuth client '<uuid>' … cannot be
  decrypted … Restore the CONNECTION_ENCRYPTION_KEY …` (row id + env var name +
  "a key-rotation/decrypt failure is in progress").
- `integration-connections.ts:1051` — auto-provisioning failure, which embeds
  the upstream authorization server's **own prose** verbatim.
- `integration-connections.ts:1057` — names the package/auth key.

The same class of defect sat one handler above at `:487` on the equally
session-less `/callback`: the non-`OAuthCallbackError` fallback rendered
`Error: ${err.message}` — any DB fault, TypeError or SSRF refusal.

## Fix

Both sites now render a generic sentence and keep the detail on the log line
that was already there — the rule `oauth-error-diagnostic.ts` states for the
provider's `error_description` on this same surface.

- `/connect/start`, 4xx branch: `"This integration is not ready to be connected.
  Ask an administrator to finish setting it up, then open this link again."`
  Everything #1302 added is kept: the ApiError status (403 ≠ 502, so the popup
  still reads "permanent" instead of inviting a pointless retry) and
  `releaseJti` (the link stays reusable).
- `/callback` generic branch: `"Could not complete the connection. Please try
  again."`, matching the `OAuthCallbackError` branch beside it.

No new helper: the "public sentence in the page, operator detail in the log"
split is the existing doctrine here, and inventing a public/operator field on
`ApiError` for two call sites would be debt, not DRY.

Deliberately **not** changed: `:526`, where the `identity_mismatch` detail is
surfaced on purpose. It does name the other account, so it is a disclosure
judgement of its own — a product decision, not this issue's mechanism.

The issue's secondary claim — that a third-party AS `error_description` is
echoed on `/callback` — is stale: `oauth-error-diagnostic.ts` already refuses
free text and appends only the RFC 6749 code. The upstream prose did leak, but
through `provisioningFailure.message` on `/connect/start`, which this PR closes.

## Tests

`apps/api/test/integration/routes/integrations-hosted-connect.test.ts` — the two
#1302 tests that asserted the verbatim render are inverted, and one is added for
the issue's headline case (a client row whose ciphertext no longer opens):

- `renders a permanent 403, not a generic 502 — and not the operator detail`
- `keeps the auto-provisioning failure's own prose off a remote MCP popup`
- `never names the row or the env var when a client_secret cannot be decrypted`

```
  apps/api/test/integration/routes/integrations-hosted-connect.test.ts \
  apps/api/test/integration/routes/integrations.test.ts \
  apps/api/test/integration/routes/integration-oauth-flow.test.ts \
  apps/api/test/integration/routes/integrations-connect-run-errors.test.ts \
  packages/core/test/connect-handshake.test.ts
→ 130 pass, 0 fail (635 expect() calls)
```

Negative control: with the one-line render reverted, exactly those 3 tests fail
(12 pass / 3 fail) — they are not vacuous `not.toContain` assertions.

`bunx tsc --noEmit -p apps/api` and `-p apps/web` clean; `bun run
verify:openapi` all checks passed; `bunx eslint` / `prettier --check` clean on
every touched file.

## Notes for the orchestrator

- No migration, no env var.
- **OpenAPI**: the `403` description on `/api/integrations/connect/start` said
  "the page names the action to take", which the fix makes false, so it was
  rewritten. That drifted `verify:api-types`; `bun run generate:api` was run and
  `apps/web/src/api/schema.d.ts` is committed (a one-line description change, no
  type change). Not a breaking change — no baseline regenerated.
- **Overlap with #1346**: unavoidable, same call site. #1346 changes
  `popupHtmlError`'s **second** argument (`{}` → `{ packageId }`); this PR
  changes the **first**. They compose, but a rebase will conflict on that
  statement — take both edits. My tests deliberately assert nothing about
  whether the package id appears in the page, so #1346 will not break them.
- **Overlap with #1352**: none textually. #1352 moves the `try` opening earlier
  (`:910`); the branch bodies I touched are inside it either way.

### DIFF
--- a/apps/api/src/openapi/paths/integrations.ts
+++ b/apps/api/src/openapi/paths/integrations.ts
@@ -1015,7 +1015,7 @@ export const integrationsPaths = {
         },
         "403": {
           description:
-            "The space has no OAuth client registered for this auth and none could be auto-provisioned; the page names the action to take (HTML error page). The link stays reusable so a retry after the administrator registers a client needs no re-mint.",
+            "The space has no OAuth client registered for this auth and none could be auto-provisioned; the page says the failure is permanent and to ask an administrator, while the operator-facing detail naming the exact remedy stays on the server log — this route carries no session (HTML error page). The link stays reusable so a retry after the administrator registers a client needs no re-mint.",
         },
         "410": { description: "Invalid, expired, or already-used token (HTML error page)." },
         "500": {
--- a/apps/api/src/routes/integrations.ts
+++ b/apps/api/src/routes/integrations.ts
@@ -482,9 +482,15 @@ export function createIntegrationsRouter() {
         });
         return c.html(popupHtmlError(userMessage, { state }));
       }
-      const msg = err instanceof Error ? err.message : "OAuth callback failed";
-      logger.error("Integration OAuth callback failed", { msg });
-      return c.html(popupHtmlError(`Error: ${msg}`, { state }));
+      // Not an `OAuthCallbackError`, so nothing authored this message for a
+      // reader: it is a DB fault, a TypeError, an SSRF refusal. Same rule as
+      // the branch above and as `/connect/start` (issue #1345) — this page is
+      // session-less, so the text goes to the log and the user gets the
+      // generic sentence.
+      logger.error("Integration OAuth callback failed", { err: String(err) });
+      return c.html(
+        popupHtmlError("Could not complete the connection. Please try again.", { state }),
+      );
     }
 
     // Persist via the OAuth2 strategy. The exchange above reconstructed the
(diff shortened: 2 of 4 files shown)


pierrecabriere:

Issue #1315, opened 2026-09-10
security(files): the creator branch of delete/keep bypasses the API-key credential ceiling


pierrecabriere:
## Summary

The creator branch of file delete / keep bypasses the credential ceiling entirely. An API key minted with no `files:*` scope can delete — and pin against GC — every file its creator ever produced.

**Pre-existing**: `git diff v1.0.0-beta.57..origin/main -- apps/api/src/services/files.ts` shows this logic unchanged. It is live in production.

## Mechanism

`apps/api/src/services/files.ts:338`:

```ts
const manage = isCreator || (opts.canManage ?? false);
```

The routes carry no `requirePermission` middleware at all (`apps/api/src/routes/files.ts:174`, `:200`).

Under API-key auth the actor is the key's **creator** (RBAC spec §7.1), so `isCreator` is true for every file that user has ever produced — dashboard runs included.

## Failure scenario

An API key minted with `scopes: ["agents:run"]` and no `files:*`:

- `DELETE /api/files/<id>` succeeds for any file its creator produced.
- `POST /api/files/<id>/keep` succeeds, defeating the storage-expiry GC.

This is the one place the new `callerPermissions(c)` plumbing introduced by the RBAC work is short-circuited.

## Proposed fix

Keep the creator capability for session auth, but intersect it with the credential ceiling: a scope-limited key must not inherit its creator's ownership rights. Route the check through `callerPermissions(c)` like every other RBAC call site.

PR #1381, opened 2026-09-10
fix(files): keep a detached run deliverable as narrow as its run was

### LINKED ISSUE #1314: security(files): a detached run deliverable becomes readable to every files:read holder in the space
## Summary

When a run container is torn down, its published deliverables are **detached** rather than cascaded, and the gallery's third visibility arm makes every detached `agent_output` readable by anyone holding `files:read` in the space — including files produced by another member or by an end-user.

**This is live in production today** (present at `v1.0.0-beta.57`, `services/files.ts:1460`).

## Mechanism

`detachOrDeleteContainedFiles` (`apps/api/src/services/files.ts:1639-1645`) NULLs `files.run_id` instead of cascading the row away. The row still carries the `user_id` / `end_user_id` of the producing actor.

The gallery filter's third arm (`apps/api/src/services/files.ts:1471`) is unconditional:

```ts
or(
  ...(canReadEveryRun(permissions) ? [isNotNull(files.runId)] : []),
  eq(files.userId, actor.id),
  and(isNull(files.runId), isNull(files.chatSessionId), eq(files.purpose, "agent_output")),
)!
```

The same hole exists in the detached branch of `getFileForActor` (`files.ts:1240-1259`).

## Failure scenario

1. Bob (space `builder`) runs an agent that publishes `secret-report.md`.
2. A later run links the file; an admin bulk-deletes that agent's runs → `detachOrDeleteContainedFiles` NULLs `run_id`.
3. Alice holds preset `operator` — `files:read`, `runs:read`, **no** `runs:read-all`.

Alice now gets the file in `GET /api/files` with its real filename and `capabilities.download: true`, obtains a preview token, downloads the bytes at `/content`, and can pipe it into her own run as `appfile://`. The same path hands any member an end-user's detached deliverable.

Reproduced with a probe before and after `detachOrDeleteContainedFiles`.

## Proposed fix

Detaching must not widen visibility. Either copy the run's attribution onto the file at detach time and keep testing it (the row already carries both actor columns), or narrow arm 3 to `eq(files.userId, actor.id)` for callers without `runs:read-all`. The published-file attribution contract is already described at `apps/api/src/lib/run-visibility.ts:15-19` — a published file copies its run's attribution, so the ownership question is answerable without joining back to the run.

## Verification

Executed. Confirmed.

pierrecabriere:
Closes #1314

## Root cause

A run teardown **detaches** a consumed deliverable instead of cascading it away: `detachOrDeleteContainedFiles` sets `files.run_id = NULL` (`apps/api/src/services/files.ts`, "Protected → detach"). Two read surfaces then treated "no container" as "no ACL":

- `listFilesForActor` — the member filter's third arm, `and(isNull(runId), isNull(chatSessionId), eq(purpose, "agent_output"))`, was **unconditional**, so every detached deliverable was listed for anyone holding `files:read`;
- `getFileForActor` — the detached branch refused only a `user_upload` to a non-creator; an `agent_output` resolved for any member in the org+space, with `capabilities.download: true`.

Deleting a run therefore **widened** its output: a colleague with the `operator` preset (`files:read`, `runs:read`, no `runs:read-all`) who could never read the producing run got the file's real name, a preview token, the bytes, and could feed it into their own run as `appfile://`. The same path handed any member an end-user's deliverable.

## Fix

The teardown NULLs the *container*, never the actor columns, and a published file copies its run's attribution (`getRunAttribution` → `createFileFromStream` in `routes/runs-events.ts`; the invariant is stated in `lib/run-visibility.ts:15-19`). So the ownership question outlives the run, and both surfaces now answer it from the row itself:

- a detached `agent_output` is readable by **its producer, or a `runs:read-all` holder** — exactly the disjunction that applied while it was run-contained, so detaching neither widens nor narrows;
- a detached `user_upload` stays creator-only in full, unchanged (not even opened by `read-all`).

In the gallery this moves the detached arm *inside* the `runs:read-all` branch; the own-rows arm (`user_id = me`) already covers a member's own detached outputs. No permission, preset or route contract changes — `files:delete` lives with the presets that already hold `runs:read-all` (`admin`/`builder`), so no lifecycle capability is lost.

Small refactor to keep it DRY: the creator predicate the fix needs is extracted as `isFileCreator` and read by `getFileCapabilities` too, instead of re-deriving the user/end-user column split at each gate.

Docs updated where they stated the old rule: `docs/architecture/FILES.md` "Detached ACL", and the `GET /api/files` description (which said members see "their own files (and system-owned ones)").

## Tests

Extended the nearest existing case, `apps/api/test/integration/services/files.test.ts` → "detaches a consumed file when its producer's runs are deleted (chaining survives)". It now publishes with the run's attribution the way `runs-events.ts` does, asserts the teardown preserves `files.user_id`, and covers all three readers: producer resolves + lists it, a stranger with no grants gets `null` and an absent gallery row (the regression), a stranger with `runs:read-all` resolves + lists it (no narrowing of supervision).

```
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/integration/services/files.test.ts
  → 52 pass, 0 fail (245 expect)
… files-hardening-e2e.test.ts + modules/mcp-files.test.ts + unit/inline-run-context-files.test.ts
  → 36 pass, 0 fail
… routes/runs-schedules-read-permissions.test.ts + routes/admin-storage-deletion.test.ts
  + unit/inline-run.test.ts + unit/run-admission-lock-order.test.ts
  → 32 pass, 0 fail

bun run generate:api && bun run verify:api-types → up to date (schema.d.ts committed)
bun run detect:breaking         → 0 breaking, 0 non-breaking (no baseline change)
bunx eslint / prettier on touched files → clean
```

## Notes for the orchestrator

- No migration, no env var, no deploy ordering constraint. Behaviour change only, and it is a narrowing.
- **Overlap**: `apps/api/src/services/files.ts` is also #1315's file (file delete/keep vs the API-key ceiling). My diff touches `getFileCapabilities`'s `isCreator` line, the detached branch of `getFileForActor`, and the member arm of `listFilesForActor` — not the lifecycle/route path #1315 works on. The new `isFileCreator` helper is available to it.
- `docs/plans/run-isolation-runner-preset.md` (the merged #1307 plan) records "arm 3 … is left as is". Left untouched on purpose: it is a dated plan record, not a live contract; the live contract is `docs/architecture/FILES.md`, which this PR updates.


pierrecabriere:

Issue #1341, opened 2026-09-10
security(api): eleven routes run requireAgent() before the permission guard, leaking a catalog existence oracle


pierrecabriere:
## Summary

Eleven routes call `requireAgent()` **before** the permission guard, so a caller lacking the guarded permission gets 404 for a missing agent and 403 for an existing one — an enumeration oracle over the space's private agent catalog.

Two routes in the same file already apply the correct order and document why.

## Mechanism

`requireAgent()` (`apps/api/src/middleware/guards.ts:12`) 404s on an unreachable agent. Affected routes:

- `apps/api/src/routes/agents.ts:243, 341, 362, 400, 455, 530, 556, 584`
- `apps/api/src/routes/runs.ts:213, 823`
- `apps/api/src/routes/schedules.ts:303`

The correct order is stated and applied at `agents.ts:322-325` and `:389-391`.

## Failure scenario

An API key scoped `["files:read"]` enumerates the space's private agent catalog through `PUT /api/agents/@x/y/proxy`: 403 means the agent exists, 404 means it does not.

## Not this issue

`packages.ts:2273`, `:2306`, `:2367` and `spaces.ts:612`, `:678` also 404-before-403, but there it is genuinely forced by per-type gating and is documented as such.

## Proposed fix

Move `requireAgent()` after the permission guard on the eleven routes above.

Issue #1339, opened 2026-09-10
mcp: run_and_wait hard-codes runs:read and refuses a runs:read-all principal


pierrecabriere:
## Summary

`run_and_wait` hard-codes `runs:read` and refuses a principal holding only `runs:read-all`, breaking the stated `read-all ⊇ read` invariant. It is the only re-derivation of the run-read predicate outside `lib/run-visibility.ts`.

## Mechanism

`apps/api/src/modules/mcp/tools.ts:963`:

```ts
if (!ctx.permissions.has("runs:read")) { … }
```

`rg 'has\("runs:'` across `apps` and `packages` returns this one hit. `apps/api/src/lib/run-visibility.ts:47-55` documents why:

> `read-all` is a superset of `read`, not a companion to it: a principal granted only the wide permission reads every run in the space, so it opens every surface `read` opens.

## Failure scenario

An API key scoped `["mcp:invoke", "agents:run", "runs:read-all"]` — all three grantable — is refused before launch. The same key reaches the same data through `invoke_operation → runAgent → getRun`.

## Proposed fix

Use `canReadRuns(ctx.permissions)` from `lib/run-visibility.ts`.

Issue #1338, opened 2026-09-10
security(rbac): the runner preset reads the space's locked input values


pierrecabriere:
## Summary

The `runner` preset added by #1307 grants `agents:run` without `agents:read`, and the summary projection it receives is correct on everything except one field: `input.values` and `input.locked_fields` are emitted **unconditionally**. `locked_fields` exists so the launcher cannot *set* those fields; the summary lets a runner *read* them.

## Mechanism

`apps/api/src/routes/agent-detail-handler.ts:189-193`:

```ts
input: {
  ...(m.input ?? { schema: { type: "object", properties: {} } }),
  values: storedValues,
  locked_fields: lockedFields,
},
```

The comment above it states the choice explicitly ("Emitted unconditionally — a manifest with no `input` section can still carry stored values from an earlier manifest").

## Executed

Agent detail as a `runner` (no `agents:read`):

```
200 {"input":{…,"values":{"api_token":"ADMIN-ONLY-SECRET"},
              "locked_fields":["api_token"]}, …}
```

## Verified correct alongside

The #1307 claim ("nothing an author would call the agent's content") holds for everything else — `manifest`, `prompt`, `dependencies.skills`, `dependencies.mcp_servers`, `version_count`, `has_unarchived_changes` and `forked_from` are all correctly withheld from both the list and the detail. The bundle route, file explorer, versions and `versions/:version` stay on `agents:read`. `POST /api/runs/remote` provisions no workspace bundle, so the sink secret cannot be turned into `GET /runs/:runId/workspace`.

## Proposed fix

Emit the launch form's *schema* to a runner but not the stored values of locked fields — a locked field is by definition one the launcher never supplies, so its value has no place in the launch payload. The same field is the payload in the end-user cross-space issue.

PR #1382, opened 2026-09-10
fix(rbac): confine every space-pinned credential to its space on /api/spaces

### LINKED ISSUE #1313: security(rbac): an end-user token reads locked input values from any space in the org, private ones included

pierrecabriere:
Closes #1313

## Root cause

`apiKeySpaceScopeGuard` (`apps/api/src/middleware/guards.ts:104` on main) keyed the path-param check on `authMethod === "api_key"`, so it passed straight through for every other credential kind. An OIDC end-user token pins a `spaceId` too (`modules/oidc/auth/strategy.ts:321`) and carries **no** `orgRole`, so `applySpacePermissions` (`middleware/space-context.ts:70`) returns early for it and never compares the path space to the pinned one either. `/api/spaces` is not in `SPACE_SCOPED_PREFIXES`, so `requireSpaceContext` — the one place that *does* compare them — never runs on that router.

Net effect: an end-user minted for space A, holding the end-user-grantable `agents:read`, could call `GET /api/spaces/<B>/packages/{scope}/{name}/run-config` for any space B in the org, `private` included, and get back that space's stored `input.values` and `locked_fields`. `run-config` is the only route under the `/:spaceId/packages/*` block whose permission string is end-user-grantable (`modules/oidc/auth/scopes.ts:43`); its siblings need `spaces:read` / `agents:configure`, which are not.

## Fix

One guard, keyed on the **pinned space** instead of on the auth method, renamed `pinnedSpaceScopeGuard`. It is already mounted on `/:id` and `/:spaceId/*` of the spaces router, so no new call sites and no per-route work.

This is not a new rule — it is the rule `routes/runs-remote.ts:205-208` already writes down for its own body-supplied space: *"Enforced for EVERY auth method — gating on `authMethod === "api_key"` would leave the same escape open to any module strategy that pins a space (e.g. oauth2-end-user bearers)."* The spaces router had not caught up.

The same wrong signal appeared three lines below, in the `GET /api/spaces` listing narrowing (`authMethod === "api_key" ? filter : entries`); it is keyed on the pinned space now too, so the router does not carry one fixed line next to one unfixed one.

**Both changes are narrowing-only, with zero behaviour change today:**
- API keys always pin a space (`lib/auth-pipeline.ts:209`) → identical behaviour.
- Sessions and OIDC instance tokens pin none, and `requireSpaceContext` does not run on `/api/spaces` → identical passthrough, still gated per space by `applySpacePermissions`.
- Only credentials that pin a space *and* were not API keys change — which is exactly the leak.

The only other pinning strategy in the tree is the OIDC end-user one (`module-chat`'s loopback strategy pins no space), so the audited surface is complete.

Docs updated to match: `docs/architecture/SPACES.md`, `AGENTS.md`, `apps/api/CLAUDE.md` (the SPACES.md file:line references were already stale and are now correct).

## Tests

New: `apps/api/test/integration/routes/spaces-end-user-space-scope.test.ts` — a stub auth strategy minting the end-user shape (org, pinned space, no `orgRole`, `agents:read`), a `private` second space holding a locked secret, asserting 403 + no secret in the body, plus a positive control proving the token still reads its own space's `run-config`.

```
TEST_TIER=0 bun test apps/api/test/integration/routes/spaces-end-user-space-scope.test.ts   → 2 pass / 0 fail
```

**Negative control**: with the guard's condition reverted to `authMethod !== "api_key"`, that test returns **200** with the foreign private space's `run-config` — the leak reproduced, then closed.

Regression sweep (tier 0, from the worktree root):

```
spaces.test.ts, spaces-run-config.test.ts, api-keys-space-authority.test.ts,
space-membership.test.ts, space-member-management.test.ts,
space-package-install-permissions.test.ts, space-role-delegation.test.ts,
agents-space-packages-read-permissions.test.ts,
rbac-disjunction-and-space-package-reads.test.ts                       → 91 pass / 0 fail

agents.test.ts, view-as.test.ts, rbac-catalog-boundaries.test.ts, roles.test.ts,
error-paths.test.ts, openapi-response-validation.test.ts, strict-request-bodies (unit+int),
oidc space-social/space-smtp, test/integration/middleware/*                → 305 pass / 0 fail

spaces + api-keys-space-authority + view-as + space-membership + new file
(re-run after the listing change)                                          → 93 pass / 0 fail
```

`bunx tsc --noEmit -p apps/api` → clean. eslint + prettier clean on every touched file.

## Notes for the orchestrator

- **No migration, no env var, no OpenAPI change.** The route already declared 403 (`openapi-response-validation.test.ts` passes untouched). Deploy ordering: none.
- **Deliberately NOT fixed here — issue #1313's proposed fix #2** (gate `input.values` / `locked_fields` on `agents:read` in `routes/agent-detail-handler.ts:189-193`). That is **sibling #1338**'s territory — same field, different principal (the `runner` preset). The `run-config` route this issue is about *already* gates on `agents:read`, so closing the cross-space escape is the whole fix for #1313; #1338's principal legitimately holds the space but not `agents:read`, which is a projection problem, not a scoping one. No shared guard is warranted: the two defects meet at a field name, not at a check.
- **Deliberately NOT fixed here — the `setSpaceContextApplier` seam** (`space-context.ts:177-201`), which issue #1313 also names. Verified not exploitable: the only module that enters a *named* other space is `webhooks`, and `webhooks:*` is excluded from `OIDC_ALLOWED_SCOPES`, while for API keys `webhookScope()` already narrows to the bound space and `POST /api/webhooks` does its own explicit comparison. Enforcing the pinned space inside `applySpacePermissions` would preempt those checks and change the webhooks error path for cases that are already refused — cost without benefit. Worth a separate issue if the end-user scope allowlist ever grows.
- **Adjacent, left alone**: `POST /api/spaces` still refuses on `authMethod === "api_key"` ("API keys cannot create spaces"). That is a capability rule, not a space-scope check, so it is not the same defect; an end-user token cannot reach it anyway (`spaces:write` is not end-user-grantable).
- **Overlap with siblings**: none in files touched. #1341 (`requireAgent` before permission guard) and #1339 (`run_and_wait` hard-codes `runs:read`) are in different files; #1338 is in `agent-detail-handler.ts`, which this PR does not touch.


pierrecabriere:

Issue #1364, opened 2026-09-10
cli: the skills-sync flock never inspects errno, and cannot load at all on Windows


pierrecabriere:
## Summary

The skills-sync lock never inspects errno, so every `flock` failure reads as "another sync is running" — including "this filesystem does not support `flock`". And on Windows the lock cannot load at all, so `skills sync` can never succeed and `logout` silently skips skill cleanup.

## Mechanism

`apps/cli/src/lib/skills-sync/lock.ts:65-68` — the FFI binding (`:84-86`) exposes only the return value, so `EWOULDBLOCK`, `EBADF`, `EINVAL`, `ENOLCK` and `EOPNOTSUPP` all enter the same 500 ms poll loop and, after 60 s, throw `SyncLockBusyError("Another appstrate skills sync is running")`.

On a filesystem without `flock` — NFS without lockd, some 9p / virtiofs devcontainer mounts — every `skills sync` blocks the marketplace command for a minute and then names a concurrent process that does not exist. There is no degrade path.

`lock.ts:35-38` — the non-darwin branch is `["libc.so.6", "libc.musl-x86_64.so.1", "libc.musl-aarch64.so.1"]`. On `win32` all three `dlopen`s fail and `:92-94` throws:

- `skillsSyncCommand` → exit 1, every time.
- `logoutCommand:80-83` warns and skips skill cleanup (credentials are still wiped, correctly).

Windows is a supported target — the CLI carries explicit `win32` handling in `lib/keyring.ts`, `lib/doctor.ts`, `lib/install/os.ts` and `lib/dual-install-check.ts`. There is simply no `LockFileEx` path and no advisory-lockfile fallback here.

## Proposed fix

Read errno and distinguish "busy" from "unsupported" — on unsupported, fall back to an advisory lockfile (or proceed with a warning) rather than blocking for a minute. Add a `LockFileEx` path or the same advisory fallback for Windows.

## Verified correct alongside

The fd lifetime is right (`openSync` / `closeSync` in `finally`; Node fds are not GC-managed), the ABI and `LOCK_EX|LOCK_NB` constants are correct on both supported platforms, and not unlinking the lock file is the right call.

PR #1383, opened 2026-09-10
fix(cli): apply an organization revocation instead of failing the skills sync

### LINKED ISSUE #1362: cli: losing org membership exits 1 and leaves every synced skill on disk forever

pierrecabriere:
Closes #1362

## Root cause

`selectedSpaces` called `listSpaces` with no status discrimination (`apps/cli/src/commands/skills.ts:601` before this change), and the throw was caught by the whole-run handler at `:203-206` → `report.run` → exit 1. Every failure of `GET /api/spaces` was therefore a fault to retry, including the one that is a *statement about the caller's grants*.

Losing organization membership is exactly that statement. `DELETE /api/orgs/:orgId/members/:userId` → `removeMember` (`apps/api/src/services/organizations.ts:355`) deletes the membership row, the member's notifications, their space memberships and their schedules — and **nothing else**: no Better Auth session delete, no `revokeFamilyForUser`. So an offboarded user's CLI still authenticates fine, and `orgContext` (`apps/api/src/middleware/org-context.ts:64`) answers `403 "You are not a member of this organization"` — forever. Under `--print-path` (the marketplace `command` source) Claude Code discards the failing run each session and keeps serving the stale plugin, so the ex-employee's machine kept every one of that organization's skills indefinitely with nothing on any path to clear it.

## Fix

One new helper, `reachableSpaces`, between `selectedSpaces` and `listSpaces`: a 403 resolves to "no space supplies skills", which the ordinary removal plan then applies, with a `report.note` naming the remedy (`appstrate org switch`). No new deletion path — an empty source set is already a supported state, so the revocation reuses the machinery `suppliesSkills` uses for a *space*-level revocation. Everything else rethrows unchanged.

**Only 403, deliberately — not the 401/403 the issue proposed.** 403 is the server speaking about this profile's grants (no membership row, or an org role that cannot read the space catalog). A 401 is about the *session*: `apiFetch` already turns it into a re-login `AuthError` after a refresh attempt, and this file's existing doctrine is that a lapsed login must never take working skills away (`bootstrapPlugin`'s "an existing one is kept"). The evidence above closes the gap that would otherwise leave: offboarding provably produces 403, never 401, because member removal revokes no CLI token family. 5xx and network errors stay faults that leave the tree untouched.

## Tests

`apps/cli/test/skills-command.test.ts`, extending the existing `skills sync — multiple spaces` describe (+47 lines, one shared `failSpaceListing` helper):

- *removes every synced skill when the organization has revoked this profile* — sync, then 403 on `/api/spaces`: the run succeeds, the note is printed, `skills/` is empty.
- *keeps every skill when the space listing fails for anything but a revocation* — same setup with 503: `ExitError`, no note, the tree intact. Guards the branch against widening into "the listing did not come back".

```
cd apps/cli && TEST_TIER=0 bun test test/skills-command.test.ts   → 87 pass, 0 fail (261 expect)
bunx tsc --noEmit -p apps/cli                                     → clean
bunx eslint / prettier --check on both touched files              → clean
```

Negative control: with `err.status !== 403` flipped to `999` the revocation test fails with `process.exit(1) called` — the exact reported behaviour — so it exercises the fix rather than the setup.

## Notes for the orchestrator

- Based on `fix/issue-1320` (PR targets it, not `main`). That branch added `asSpace` in `apps/cli/src/lib/spaces.ts`; this change sits one layer up in `skills.ts` and does not touch it. The two are complementary and the distinction matters: #1320 refuses a *malformed* listing (a 200 that says nothing about standing, which must never be read as "member of nothing"), #1362 applies an *authoritative empty* one (a 403 that says exactly that).
- Overlap with siblings: none. #1363 (`syncSpaces`) and #1364 (flock) touch other regions of the CLI; nothing here reads `syncSpaces` or the lock.
- Behaviour change worth a release note: a 403 space listing now *removes* skills and exits 0 where it used to exit 1 and remove nothing. That is the point of the issue, but it is the one case where a sync deletes on the strength of a server refusal.

### DIFF
--- a/apps/cli/src/commands/skills.ts
+++ b/apps/cli/src/commands/skills.ts
@@ -11,6 +11,7 @@
 
 import { mapWithConcurrency } from "@appstrate/core/map-with-concurrency";
 import { resolveActiveProfile, type Profile } from "../lib/config.ts";
+import { ApiError } from "../lib/api.ts";
 import { listSpaces, resolveSpaceRef, type Space } from "../lib/spaces.ts";
 import { DEFAULT_IO, type CommandIO } from "../lib/io.ts";
 import { formatError } from "../lib/ui.ts";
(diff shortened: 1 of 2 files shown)


pierrecabriere:

PR #1384, opened 2026-09-10
fix(module-ee): bound the batch size by the read budget the replay window shares

### LINKED ISSUE #1328: ee(billing): EE_RECONCILIATION_BATCH_SIZE above ~800 is dead code — the drain loop stops and its warning never fires

pierrecabriere:
Closes #1328

## Root cause

A sweep pass asks the platform for `replaySpan + batchSize` rows, clamped at the
`usage.list` ceiling of 1000 (`packages/module-ee/src/billing/usage-recorder.ts:619`).
The replay window is read ON TOP of the batch, so the FORWARD slice is
`1000 - replaySpan` — ~800 at the default window of 200. The within-tick drain loop is
gated on `result.processed >= batchSize` (`packages/module-ee/src/billing/billing-sweeper.ts:341`),
so for any batch size above ~800 that predicate is unreachable on a backlogged ledger.

Raising the batch to 1000 to clear a backlog therefore made throughput **fall** — from up
to 50 drained batches per tick to a single ~800-row pass — and `backlogRemains` became
permanently false, so the "billing sweep hit the drain cap" warning that would have
surfaced it never fired. The `.max(1000)` on the batch (`packages/module-ee/src/env.ts:37`)
was derived before the replay window existed and was not re-derived when it landed; its own
comment claims to prevent exactly this dead-code condition.

## Fix

The two knobs share one read budget, so make that explicit instead of leaving each with an
independent maximum:

- `LEDGER_LIST_MAX_LIMIT` moves from `usage-recorder.ts` to `env.ts` (one definition, next to
  the two variables it bounds; the recorder imports it).
- A cross-field `.refine` on `eeEnvSchema` rejects `EE_RECONCILIATION_REPLAY_WINDOW +
  EE_RECONCILIATION_BATCH_SIZE > 1000` at boot, naming both variables. Same shape as the
  existing `WEBHOOK_TIMESTAMP_TOLERANCE_SECONDS` / `REMOTE_RUN_REPLAY_WINDOW_SECONDS` rule in
  `packages/env/src/index.ts:945`.

Rejected rather than clamped: a silently clamped batch is the undiagnosable version of the
same bug. With the rule in place the forward slice is exactly `batchSize`, so the drain loop's
accounting means what it says for every legal configuration.

Verified `eeEnvSchema.shape` survives the refine (Zod 4 keeps refinements in-schema) — both
`verify:env-docs` and `verify:compose-defaults` read that property through
`scripts/lib/module-env-schemas.ts`, and both stay green.

Comments corrected in the same pass: the batch-size and replay-window blocks in `env.ts` and the
scan-window block in `usage-recorder.ts` all asserted the now-false "the two knobs are independent
by construction".

## Tests

`packages/module-ee/test/unit/env.test.ts` — two cases: the boundary (800/200 legal, 801/200
refused, 1000/0 legal again, 501/500 refused) and that the boot message names both variables.
Every pre-existing sweeper test ran at batch size 1, 4 or 100, so the dead range was uncovered.

```
pg-test-lock.sh sh -c '... bun test packages/module-ee/test/unit/env.test.ts'
  → 12 pass, 0 fail
pg-test-lock.sh sh -c '... bun test packages/module-ee/test/integration/services/billing-sweeper.test.ts'
  → 63 pass, 0 fail
pg-test-lock.sh sh -c '... bun test .../org-drain.test.ts .../storage-entitlement.test.ts .../repair-account.test.ts'
  → 33 pass, 0 fail
bunx tsc --noEmit -p packages/module-ee   → clean
bunx eslint <touched files>               → clean
bun run verify:env-docs                   → ✓ (106 schema vars, 7 from the module schema)
bun run verify:compose-defaults           → ✓
```

## Notes for the orchestrator

- Based on `fix/issue-1326`; PR targets that branch.
- No migration, no new env var. `.env.example` and `docs/ENV.md` update the description of
  `EE_RECONCILIATION_BATCH_SIZE` only — no key added or renamed.
- **Deploy note**: this is a boot-refusing validation change. A deployment that has
  `EE_RECONCILIATION_BATCH_SIZE` set such that it plus `EE_RECONCILIATION_REPLAY_WINDOW`
  exceeds 1000 will fail to start until the value is lowered. Both variables are unset in
  prod as far as this branch can tell (defaults 100 + 200 = 300), so this should be a no-op —
  worth a pre-flight `grep EE_RECONCILIATION` on the prod `.env` all the same.
- No overlap with #1330, #1370 or #1329. `usage-recorder.ts` is touched only for the moved
  constant and one comment; `billing-sweeper.ts` is not touched at all.


pierrecabriere:

Issue #1349, opened 2026-09-10
db: the Better Auth schema check validates the Drizzle object, never the database


pierrecabriere:
## Summary

The Better Auth schema check diffs two **TypeScript objects**; it never reads the database. A migration that missed a column the TS schema declares would boot clean and fail as a Postgres `42703` on the first `/oauth2/token`.

## Mechanism

`findDrizzleSchemaProblems` (`@better-auth/drizzle-adapter/dist/index.mjs:32-34`, `:601`) diffs `getExpectedSchema(options)` against `introspectDrizzleSchema(schema)`. `diffSchema` (`@better-auth/core/dist/db/schema-diff.mjs:29-56`) emits exactly three finding kinds — `missing-table`, `missing-column`, `unexpected-required-column` — all computed from the TS object. It never reads `information_schema`.

The comment at `packages/db/src/schema/oidc.ts:24-27` —

> a field the provider writes with no column here raises `SchemaMismatchError` and rejects auth traffic

— is accurate about **TS drift** and is easy to misread as covering the database.

There is no DB-drift gate anywhere in `bun run check` either.

## Why it matters now

Migration 0057 was verified by hand for this audit and misses nothing. But this is the same class as the recorded "prod: 2 index manquants" incident, and the index gap in the companion issue is invisible to the check for the same reason: `diffSchema` does not look at indexes at all.

## Proposed fix

Add a boot-time or CI-time comparison against the live database for the Better Auth tables — column presence, nullability, and the indexes upstream declares. A clean boot on a database missing a column is a production auth outage with no warning.

Issue #1348, opened 2026-09-10
db(oidc): eight FK columns better-auth declares indexed have no index — every logout seq-scans both token tables


pierrecabriere:
## Summary

better-auth 1.7.3 declares `index: true` on eight FK columns across the OAuth tables. None of them is indexed, in TypeScript or in SQL — and two are `ON DELETE` targets, so **every logout sequential-scans both token tables end to end**.

## The gap

Upstream declarations (`@better-auth/oauth-provider/dist/authorize-9whjxVLJ.mjs`): `oauthAccessToken.clientId` `:4017`, `.sessionId` `:4027`, `.userId` `:4036`, `.refreshId` `:4062`; `oauthRefreshToken.clientId` `:3922`, `.sessionId` `:3932`, `.userId` `:3941`; `oauthConsent.clientId` `:4090`, `.userId` `:4099`.

Repo side:

| table | indexed | declared `index: true` upstream |
|---|---|---|
| `oauth_access_tokens` | `authorization_code_id` | `client_id`, `session_id`, `user_id`, `refresh_id` |
| `oauth_refresh_tokens` | `authorization_code_id` | `client_id`, `session_id`, `user_id` |
| `oauth_consents` | **none** | `client_id`, `user_id` |

`packages/db/src/schema/oidc.ts:236-241`, `:268`, `:271-287`, `:186-195`. Verified across all 61 migrations: the only indexes ever created on these tables are `idx_oauth_clients_org`, `idx_oauth_clients_app` and the two `*_auth_code` indexes 0057 adds.

## Cost

Postgres does not auto-index referencing columns. `session_id` / `user_id` are FK columns with `ON DELETE set null` / `cascade`, so Postgres seq-scans the referencing table for every parent row deleted:

- every **logout** scans `oauth_access_tokens` and `oauth_refresh_tokens` end to end
- every **user deletion** likewise
- `oauth_consents` is scanned by `(client_id, user_id)` on every `/authorize`

On an instance with a large token table this is an O(rows)-per-call amplification path reachable from `/sign-out` — rate-limited per IP, but the work is real.

## Why nothing caught it

`diffSchema` emits only `missing-table`, `missing-column` and `unexpected-required-column` — it does not check indexes. See the companion issue: the boot check reads the TypeScript objects, not the database.

## Proposed fix

Add the eight indexes in one migration, and consider a check that upstream `index: true` declarations are honoured.

Issue #1012, opened 2026-07-27
OIDC: admin-registered OAuth clients cannot use loopback redirect URIs in production (DCR can)


pierrecabriere:
Split out of #953, whose repro step 2 ("register an OAuth client with a loopback redirect URI") fails in production for a reason unrelated to that issue's subject.

## Problem

The two client-registration paths disagree on loopback redirect URIs.

- **Dynamic Client Registration** (`POST /api/auth/oauth2/register`, what Codex / Claude Code use) is validated by `@better-auth/oauth-provider`, which allows `http://` when the host is loopback (RFC 8252 §7.3). Works everywhere.
- **Admin CRUD** (`POST /api/oauth/clients`, the dashboard path) goes through `isValidRedirectUri()` in `apps/api/src/modules/oidc/services/redirect-uri.ts`, which allows `http://` only when `isDevEnvironment()` is true — i.e. only when `APP_URL` itself points at localhost.

So on any real deployment (`APP_URL` = https), registering a native/CLI client with `http://127.0.0.1/callback` through the admin API returns 400, even though the exact same client registered via DCR is accepted and completes the flow.

## Secondary detail

`LOCALHOST_HOSTS` (`apps/api/src/services/redirect-validation.ts`) is `{"localhost", "127.0.0.1"}`. It omits the IPv6 loopback `[::1]`, which RFC 8252 §7.3 puts on equal footing, and covers only the single IPv4 literal rather than `127.0.0.0/8`.

## Why it matters

Any native or CLI OAuth client is loopback-only by construction. An operator who registers such a client by hand (rather than letting it self-register) hits a wall with no obvious cause, and the error message ("redirectUri scheme or host is not allowed") does not hint that the DCR path would have accepted it.

## Suggested direction

Align the admin path with RFC 8252: accept `http://` when the host is loopback (`localhost`, `127.0.0.0/8`, `[::1]`) regardless of environment, keeping the current https-or-nothing rule for every other host. The SSRF concern that motivates the current gate does not apply to a redirect URI — it is a browser navigation target, not a URL the server fetches.

Not urgent: no shipped client is blocked, since the ones that need loopback all self-register.

pierrecabriere:
PR ouverte : #1015

[...]

PR #1385, opened 2026-09-10
fix(deps): collapse the duplicate @better-auth/core and guard the single-instance invariant

### LINKED ISSUE #1347: auth: two @better-auth/core copies remain, one direct import is still live, and nothing guards the invariant
## Summary

Two `@better-auth/core@1.7.3` instances are still installed, and nothing prevents a new direct `@better-auth/core/*` import from re-opening the marker-identity bug #1295 fixed. The invariant established by that commit is unguarded, and one such import is still live.

## Mechanism

Two instances differing only in the resolved `jose` peer:

- `apps/api/node_modules/@better-auth/core` → `.bun/@better-auth+core@1.7.3+0ac89928…` (jose 6.2.12) — `apps/api`'s direct dependency, `apps/api/package.json:34`, pinned `jose: ^6.2.12` at `:47`
- `better-auth`, `@better-auth/oauth-provider`, `@better-auth/drizzle-adapter` → `.bun/@better-auth+core@1.7.3+53b163a2…` (jose 6.2.7)

Probed live from `apps/api`:

```
SAME verifyJwsAccessToken identity?:      false
SAME isInsufficientScopeError identity?:  false
core-minted err recognised via facade?:   false   ← the #1295 bug, on demand
facade-minted err recognised via facade?: true
```

Still importing directly:

- `apps/api/src/modules/oidc/services/enduser-token.ts:21` — `verifyJwsAccessToken` from `@better-auth/core/oauth2`. `better-auth/oauth2` re-exports it (`dist/oauth2/index.mjs` line 4), so this is a one-line specifier fix.
- `apps/api/src/modules/oidc/services/redirect-uri.ts:37` — `isLoopbackHost`.

Impact of the remaining import is bounded: `verify.mjs`'s module state is JWKS caches, and the catch at `enduser-token.ts:138-145` is fail-closed (`return null`). The defect is that instance linkage is install-order dependent — CI green and prod red on the same lockfile — and nothing checks it.

## Proposed fix

1. Fix the two import specifiers to go through `better-auth/*`.
2. Add an ESLint `no-restricted-imports` rule forbidding `@better-auth/core/*` outside the façade.
3. Collapse the duplicate: either drop `@better-auth/core` from `apps/api`'s direct dependencies, or add a `jose` override so both resolve to one copy.

## Verified correct alongside

Exactly one `better-auth` instance (`+54492ba0…`), shared by `apps/api` and `@better-auth/oauth-provider` — the façade route is architecturally stable, and both core instances link identical `better-call`, `kysely` and `zod`, which is why `instanceof APIError` kept working.

pierrecabriere:
Closes #1347

## Root cause

`@better-auth/core` carries `jose` as a **real peer**, so the version skew between `apps/api` (`jose: ^6.2.12`, `apps/api/package.json:47`, set by #1289) and better-auth's own transitive pin (6.2.7, `bun.lock`) made bun materialise two peer instances of `@better-auth/core@1.7.3`. Which copy a consumer links is decided by the peer graph, not by any line of our code — and `oauth2/verify.mjs` marks insufficient-scope errors in a **module-level WeakSet**, private to a copy. That is the bug #1295 fixed, one import site at a time (`apps/api/src/modules/mcp/router.ts:39-54`), leaving the invariant itself unguarded and one direct import still live at `apps/api/src/modules/oidc/services/enduser-token.ts:21`.

Reproduced in a clean worktree off `main`: `node_modules/.bun/` held `@better-auth+core@1.7.3+0ac89928…` (jose 6.2.12, linked by `apps/api`) and `+53b163a2…` (jose 6.2.7, linked by `better-auth` and `@better-auth/oauth-provider`).

## Fix

Three layers, smallest first:

1. **Collapse the duplicate** — pin `jose` in the root `overrides`, next to `zod`. zod is pinned there already, which is precisely why both core copies linked an identical zod and `instanceof APIError` kept working. After the change: exactly one `@better-auth/core` and one `jose`, with `apps/api`, `better-auth` and `@better-auth/oauth-provider` on the same copy. `bun install --frozen-lockfile` is green on the committed lock.
2. **Route the live imports through the façade** — `verifyJwsAccessToken` (`enduser-token.ts`) and `encodeBasicCredentials` (two oidc integration tests) now come from `better-auth/oauth2`, which re-exports `@better-auth/core/oauth2` from the copy better-auth itself links.
3. **Guard it** — an ESLint `no-restricted-imports` ban on `@better-auth/core/*` across `apps/api/**`, plus a unit test asserting the resolution is single.

Two judgement calls worth flagging:

- **`@better-auth/core/utils/host` is carved out of the ban**, in the rule rather than with a disable at the import site. better-auth re-exports no host classifier (its own `utils/url.mjs` deliberately keeps the zod-carrying one out of the client bundle — verified against the installed dist), and `isLoopbackHost` is a pure predicate with no module state to get wrong. `oidc/services/redirect-uri.ts` needs the very predicate DCR applies; re-implementing it is the drift #1012 is about. Deep subpaths under `utils/host` stay banned. This also means `@better-auth/core` must remain a direct dependency of `apps/api` — the issue's "drop the direct dependency" option is not available.
- **The Pi SDK ban is restated, not duplicated by accident.** ESLint flat config *replaces* a rule across blocks rather than merging it (the config file says so itself, at the web seam exemption), so a new block covering `apps/api/**` would have silently dropped the Pi guard for `apps/api/src/**`. `PI_SDK_BAN` is now a shared constant carried by both blocks. Verified by positive control.

## Tests

New: `apps/api/test/unit/better-auth-single-instance.test.ts` (2 tests) — asserts `@better-auth/core/oauth2` resolves to one path from `apps/api`, from `better-auth` and from `@better-auth/oauth-provider`, and that one `jose` sits under it. The behavioural half was already covered by `apps/api/src/modules/mcp/test/integration/mcp.test.ts` (the `insufficient_scope` step-up challenge); what was unguarded is the resolution, which can split again on the next bump with our code unchanged.

**Negative control**: restored the pre-fix `package.json` + `bun.lock`, wiped `node_modules`, reinstalled → 2 core copies, and the new test fails on the exact split (`+0ac89928…` vs `+53b163a2…`). Restored, reinstalled → 2 pass.

```
TEST_TIER=0 bun test apps/api/test/unit/better-auth-single-instance.test.ts        → 2 pass 0 fail
TEST_TIER=0 bun test .../oidc/test/integration/{routes/realm-isolation,services/oauth-flows}.test.ts → 20 pass 0 fail
TEST_TIER=0 bun test .../mcp/test/integration/mcp.test.ts                          → 20 pass 0 fail
TEST_TIER=0 bun test .../oidc/test/{unit/enduser-token-verify,integration/middleware/enduser-token-auth}.test.ts → 34 pass 0 fail
TEST_TIER=0 bun test apps/api/test/unit/redirect-uri.test.ts                       → 14 pass 0 fail (with realm-isolation)
bunx tsc --noEmit -p apps/api | -p packages/db | -p packages/core                   → clean
bunx eslint <touched files>                                                        → clean
```

ESLint positive controls (probe files, deleted after): `@better-auth/core/oauth2` in `src/` → error; the same in `test/` → error; `@better-auth/core/utils/host/deep` → error; `@better-auth/core/utils/host` → allowed. Pi SDK ban still fires in both `apps/api/src/` and `packages/runner-pi/src/`.

## Notes for the orchestrator

- **`bun.lock` changed** and must ride with `package.json`. The override also lifts `@modelcontextprotocol/sdk`'s `jose` from 6.2.4 to 6.2.12 (its range `^6.1.3` is satisfied); the MCP integration suite is green. Expect a lockfile conflict with any sibling that also touches deps — re-resolve by re-running `bun install`, not by hand-merging the lock.
- No migration, no env var, no OpenAPI change. Nothing to sequence at deploy.
- **Overlap with siblings**: none. #1348 (better-auth FK indexes) and #1349 (schema check reads the Drizzle object) touch `packages/db`; this PR touches `apps/api`, the root `package.json`/lock and `eslint.config.mjs`. If a sibling adds a `no-restricted-imports` block for `apps/api/**`, it must carry `PI_SDK_BAN` and `BETTER_AUTH_CORE_BAN` — the replace-not-merge trap is documented at both blocks.

### DIFF
--- a/apps/api/src/modules/oidc/services/enduser-token.ts
+++ b/apps/api/src/modules/oidc/services/enduser-token.ts
@@ -18,7 +18,7 @@
  */
 
 import * as jose from "jose";
-import { verifyJwsAccessToken } from "@better-auth/core/oauth2";
+import { verifyJwsAccessToken } from "better-auth/oauth2";
 import { getEnv } from "@appstrate/env";
 import type { OrgRole } from "@appstrate/core/permissions";
 import { logger } from "../../../lib/logger.ts";
--- a/apps/api/src/modules/oidc/test/integration/routes/realm-isolation.test.ts
+++ b/apps/api/src/modules/oidc/test/integration/routes/realm-isolation.test.ts
@@ -22,7 +22,7 @@
  */
 
 import { describe, it, expect, beforeEach } from "bun:test";
-import { encodeBasicCredentials } from "@better-auth/core/oauth2";
+import { encodeBasicCredentials } from "better-auth/oauth2";
 import { eq } from "drizzle-orm";
 import { db } from "@appstrate/db/client";
 import { user as userTable, session as sessionTable } from "@appstrate/db/schema";
--- a/apps/api/src/modules/oidc/test/integration/services/oauth-flows.test.ts
+++ b/apps/api/src/modules/oidc/test/integration/services/oauth-flows.test.ts
@@ -52,7 +52,7 @@ import {
   restoreProtectedResources,
 } from "../../../../../lib/protected-resources.ts";
 import { getMcpOrgResourceUri, orgIdFromMcpAudience } from "../../../../../lib/audiences.ts";
-import { encodeBasicCredentials } from "@better-auth/core/oauth2";
+import { encodeBasicCredentials } from "better-auth/oauth2";
 import { decodeJwt } from "jose";
 
 // The protected-resource registry is a process-wide singleton shared with the
--- a/apps/api/test/unit/better-auth-single-instance.test.ts
+++ b/apps/api/test/unit/better-auth-single-instance.test.ts
@@ -1,0 +1,63 @@
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Regression for #1347 (and the bug #1295 fixed by hand): `@better-auth/core`
+ * must resolve to ONE copy for every consumer in `apps/api`'s import graph.
+ *
+ * `oauth2/verify.mjs` marks insufficient-scope errors in a module-level
+ * WeakSet, so `createInsufficientScopeError` (minted through `better-auth`)
+ * and `isInsufficientScopeError` (read by `@better-auth/oauth-provider` when
+ * building the step-up challenge) are only the same marker while both come
+ * from the same copy. #1295 was that invariant broken: `apps/api` pinned
+ * `jose ^6.2.12` while better-auth's transitive `jose` sat at 6.2.7, so bun
+ * materialised two peer instances of `@better-auth/core@1.7.3` and the MCP 403
+ * silently lost its `WWW-Authenticate` header.
+ *
+ * The behavioural half is covered by
+ * `src/modules/mcp/test/integration/mcp.test.ts` ("403s an authenticated
+ * caller lacking `mcp:read` with an `insufficient_scope` step-up challenge").
+ * What was unguarded — and is what this asserts — is the resolution itself,
+ * which depends on the peer graph rather than on any line of our code: it can
+ * split again on the next dependency bump, silently, with the code unchanged.
+ * `jose` is pinned in the root `overrides` (next to `zod`, which is why zod
+ * never split) so the peers cannot skew; this test is what notices if they do.
+ */
+
+import { describe, it, expect } from "bun:test";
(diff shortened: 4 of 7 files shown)


pierrecabriere:

PR #1386, opened 2026-09-10
fix(ee): cancel the duplicate subscription a concurrent checkout leaves live

### LINKED ISSUE #1325: ee(billing): concurrent checkouts create two live subscriptions, and the superseded one is never cancelled

pierrecabriere:
Closes #1325

## Root cause

`createCheckoutSession` (`packages/module-ee/src/stripe/checkout.ts:38`) gates on `planAction`, which reads `stripeSubscriptionId` / `subscriptionStatus` — state Stripe only writes once a session is **paid**. Two sessions opened before either payment both pass the guard, and if the user pays both, Stripe holds two live subscriptions.

Only one can hold the account row: the conditional UPDATE in `checkout.session.completed` (`stripe/webhooks.ts`, `attachableSubscription`) attaches the winner, and the loser fell through to `logSupersededSubscription` — a `logger.info` line. The result was a paid subscription billing the customer every month that nothing in the product named and nothing would ever cancel.

## Fix

The refused attach now reconciles instead of logging (`reconcileSupersededCheckout`): it reads the subscription that actually holds the account and, when that is a *different* one, cancels the arriving duplicate at Stripe.

- **Cancels only against positive evidence of a winner.** No billing row, or a row already carrying this very subscription, is not a duplicate — ending a paid subscription on that much would be guesswork, so those keep the log-only behaviour.
- **A refused cancellation throws**, which drops the event claim so Stripe's redelivery retries. Safe because the handler wrote nothing on this path and cancelling an already-gone subscription is a no-op.
- **Reconciliation, not reservation.** The issue offered either. A pending-checkout marker needs new state on the row (a migration) *and* still leaves reconciliation necessary for every duplicate a reservation cannot see — a second tab that loaded before the marker, a subscription created outside Checkout. Reconciliation alone closes the harm, so it is the whole fix.
- Small refactor: `subscriptions.cancel` + the "already canceled" classification (Stripe returns it as a 400 with no distinguishing `code`) lived privately in `billing/org-cancellation.ts`. Rather than copy a Stripe error-message regex into a second file, it moved to `stripe/cancel.ts` and both callers use it. Its log lines take the caller's reason as structured context instead of baking `"for a deleted org"` into the message.
- `createCheckoutSession`'s docstring claimed "a double charge is the server's to prevent" — the opposite of what the guard can do. It now states what the guard actually covers and points at where the duplicate is settled.

**Boundary (deliberate):** a duplicate created *outside* Checkout (Stripe dashboard/API) emits no `checkout.session.completed`, so it is still only logged. Auto-cancelling an operator-created subscription is a policy call, not a bug fix. Every checkout-originated duplicate is covered, whichever delivery order Stripe uses — the loser is always cancelled by its own `checkout.session.completed`.

## Tests

`packages/module-ee/test/integration/services/webhooks.test.ts` (+2 cases): a duplicate arriving at an account already holding a live subscription is cancelled at Stripe while the winner's row is untouched; an attach refused with no winner to point at cancels nothing.

```
pg-test-lock.sh … bun test packages/module-ee/test          → 436 pass, 26 skip, 0 fail
pg-test-lock.sh … bun test webhooks + checkout + post-signup → 67 pass, 0 fail
bunx tsc --noEmit -p packages/module-ee                      → clean
bunx eslint / prettier on the 5 touched files                → clean
```

Negative control: with `reconcileSupersededCheckout` reverted to the old `logSupersededSubscription` call, the new duplicate test fails (only the `GET /v1/subscriptions/…` is recorded, no `DELETE`) — the test proves the fix rather than the fixture.

## Notes for the orchestrator

- **Base is `fix/issue-1323`**, the sibling's branch. This builds directly on its `subscriptions.retrieve`-before-write: at the refusal point the subscription is already known live, which is what makes cancelling it correct.
- **Overlap:** `stripe/webhooks.ts` is shared with #1323 (base, already merged in here), #1324 and #1370. My change adds two helpers and rewrites one line inside the `checkout.session.completed` arm; #1324's fix is in the `customer.subscription.updated` arm, so a textual conflict is unlikely but both touch the same file.
- No migration, no env var, no OpenAPI/route contract change. No deploy ordering constraint.
- **Open product decision:** cancelling stops the *recurring* charge, but the duplicate's first invoice is already paid. Refunding it is a pricing policy call, not something this fix should decide — the new `logger.warn` ("Duplicate paid subscription — cancelling the one that lost the attach", with both subscription ids) is the hook for whatever policy is chosen.

### DIFF
--- a/packages/module-ee/src/billing/org-cancellation.ts
+++ b/packages/module-ee/src/billing/org-cancellation.ts
@@ -7,50 +7,12 @@
  * unconfirmed cancellation leaves a subscription charging a customer that nothing names.
  */
 
-import Stripe from "stripe";
 import { eq, isNotNull } from "drizzle-orm";
 import { getEeDb } from "../db.ts";
 import { billingAccounts, orgUsageRecords } from "../../drizzle/schema.ts";
-import { getStripe } from "../stripe/client.ts";
+import { cancelSubscription } from "../stripe/cancel.ts";
 import { logger } from "../logger.ts";
 import { deleteBillingManagers } from "./managers.ts";
-
-/**
- * Stripe's sentence for "already canceled", returned as a 400 with no distinguishing
- * `code`. Anchored: reading any other 400 as success deletes the subscription id.
- */
-const ALREADY_CANCELED_MESSAGE =
-  /^A canceled subscription can only update its cancellation_details/;
-
-/** Is this failure indistinguishable from success — a subscription Stripe no longer has? */
-function isAlreadyCanceled(err: unknown): boolean {
-  if (!(err instanceof Stripe.errors.StripeInvalidRequestError)) return false;
-  if (err.code === "resource_missing" || err.statusCode === 404) return true;
-  return err.statusCode === 400 && ALREADY_CANCELED_MESSAGE.test(err.message);
-}
-
-/** Cancel `subscriptionId`; true when it is gone, "already gone" included. Never throws. */
-async function cancelSubscription(orgId: string, subscriptionId: string): Promise<boolean> {
-  try {
-    await getStripe().subscriptions.cancel(subscriptionId);
-    logger.info("Stripe subscription canceled for a deleted org", { orgId, subscriptionId });
-    return true;
-  } catch (err) {
-    if (isAlreadyCanceled(err)) {
-      logger.info("Stripe subscription was already gone for a deleted org", {
-        orgId,
-        subscriptionId,
-      });
-      return true;
-    }
-    logger.error("Failed to cancel Stripe subscription — billing rows kept for retry", {
-      orgId,
-      subscriptionId,
-      error: err instanceof Error ? err.message : String(err),
-    });
-    return false;
-  }
-}
 
 /** Remove the org's EE-owned rows: no EE table FKs the platform, so nothing cascades. */
 async function deleteOrgBillingRows(orgId: string): Promise<void> {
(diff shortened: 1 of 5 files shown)


pierrecabriere:

Issue #1356, opened 2026-09-10
models: POST /discover buffers an unbounded body from an operator-supplied URL, and is unaudited


pierrecabriere:
## Summary

`POST /discover` buffers an unbounded response body from an operator-supplied URL, and leaves no audit trail.

## Mechanism

`apps/api/src/services/model-providers/model-listing.ts:169`:

```ts
body = await res.json();
```

Reached from `apps/api/src/routes/model-provider-credentials.ts:371-405`, which accepts an inline `api_key` plus an arbitrary `base_url_override` from any member holding `model-provider-credentials:write` (rate-limited 6/min per user, `middleware/rate-limit.ts:148`).

`fetchModelListing` (`org-models.ts:1105`) sets `AbortSignal.timeout(10_000)` but **no size cap**, and `packages/afps-shared/src/guarded-fetch.ts` has none either. `MAX_SERVED_MODELS = 1000` caps the *parsed* array — after `res.json()` has materialised the whole payload.

New surface: before #1276 nothing read this body.

## Failure scenario

A hostile or misconfigured endpoint streams hundreds of megabytes inside the 10 s window; a handful of concurrent calls OOMs the API process.

## Second defect: no audit event

The route calls no `recordAuditFromContext`, unlike the create / update / delete routes beside it (`:317`, `:503`, `:534`). An outbound probe to an arbitrary URL carrying a user-supplied key leaves no trail.

## Third, latent

`assertApiKeyProvider` (`:104`) gates on `authMode !== "api_key"` rather than on `modelDiscovery?.mode === "static"` — an api-key provider declaring static discovery would still be enumerated. No shipped provider is in that shape today (both static providers are oauth2).

## Proposed fix

Cap the response body (stream with a byte budget, refuse past it), record an audit event, and fix the `assertApiKeyProvider` predicate.

## Verified correct alongside

SSRF is properly layered: `checkEgressUrl` pre-flight, `egressGuardedFetch` with per-hop DNS and address pinning, `maxRedirects: 0` (a 3xx is refused rather than followed with the key), and the block reason never reaches the client.

Issue #1355, opened 2026-09-10
web(models): retrying a failed single model add creates a second credential and orphans the first


pierrecabriere:
## Summary

Retrying a failed **single** model add creates a second provider credential. The batch path threads the created credential id through the retry; the single path drops it.

## Mechanism

`apps/web/src/hooks/use-models.ts:186`:

```ts
const bindCredential = async (data) =>
  data.newCredential
    ? (await createCredential.mutateAsync({ body: credentialBody(data.newCredential) })).id
    : data.credentialId;
```

`submitBatch` returns `{ failedModelIds, credentialId }` and the modal reuses it (`model-form-modal.tsx:341`).

`submitOne` (`use-models.ts:218`) returns `{ failedModelIds: [data.modelId] }` — **no `credentialId`**. The modal has nothing to reuse, `data.newCredential` is still set, so the retry mints a second credential and the first is orphaned.

## Failure scenario

Credential created → model creation refused (bad model id, quota, transient 5xx) → user fixes the field and retries → a second credential now holds the same API key, and the first is referenced by nothing.

Related: a fully-failed **batch** also leaves the credential behind if the user closes the modal instead of retrying (`use-models.ts:193-210`).

## Proposed fix

Return `credentialId` from `submitOne` on failure, the same way `submitBatch` does. Consider cleaning up a credential that ends up referenced by no model.

PR #1387, opened 2026-09-10
fix(models): follow the model listing cursor instead of reporting a truncated page as success

### LINKED ISSUE #1353: models: discovery fetches one page and reports success, silently truncating paginated providers
## Summary

Model discovery fetches one page and reports success. Against a provider that paginates — Anthropic's `/v1/models` returns 20 by default with `has_more` / `last_id` — the user sees a "successful" detection that is silently truncated.

## Mechanism

`apps/api/src/services/model-providers/model-listing.ts` performs exactly one fetch (`:147`) and one `parseServedModels` (`:191`), then `return { ok: true, models }`.

Verified: zero occurrences of `has_more`, `last_id`, `after_id` or `limit=` in the file.

## Failure scenario

An endpoint serving 21 models across two pages returns `ok: true` with 20. The user configures models from a list that is missing entries, with no indication anything was cut.

## Proposed fix

Follow the cursor while the shape declares more, with a page cap and an overall cap that stays consistent with `MAX_SERVED_MODELS`. Surface "truncated" in the result rather than reporting plain success.

pierrecabriere:
Closes #1353

## Root cause

`listServedModels` (`apps/api/src/services/model-providers/model-listing.ts`) did exactly one `fetchModelListing` + one `parseServedModels`, then `return { ok: true, models }`. Anthropic's `/v1/models` answers 20 entries with `has_more` / `last_id`, and Google's answers a page plus `nextPageToken` — so an endpoint serving more than one page was reported as a complete listing that was silently short. `POST /api/model-provider-credentials/discover` passed that straight to the model form, which printed "N models detected" over a truncated list.

## Fix

**Follow the cursor.** The two pagination dialects are folded into the existing `listingShape(apiShape)` — it already forked Google (`{ models: [{ name }] }`) from everything else (`{ data: [{ id }] }`), and now also carries how each points at its next page: Google `nextPageToken` → `?pageToken=`, OpenAI/Anthropic `has_more` + `last_id` → `?after_id=`. `listServedModels` loops on it, deduping ids across pages.

`fetchModelListing` (`org-models.ts`) takes the cursor as a second argument and appends it to the URL its shape builds, so every page is derived from the base URL — the cursor cannot accumulate across pages and the first request is byte-identical to today's (no new query parameter is sent to endpoints that do not paginate). The per-page fetch/HTTP-verdict mapping moved into a `fetchListingPage` helper so the loop stays readable; the verdicts themselves are unchanged.

**Stop, and say so.** Reading stops at 1000 models or 10 pages. A result cut by either cap — or by a listing that declares `has_more` without publishing a cursor to follow — carries `truncated: true` and logs a warning, instead of passing for a whole listing. `truncated` is a new required field on the `/discover` 200 response (documented; `detect:breaking` classifies it non-breaking), and the model form prints a line under the detected-count when it is set.

Design note: the caps are anti-runaway guards, not the normal path. `truncated` exists so the degenerate case is visible rather than silent — which is the actual complaint in the issue, and would otherwise just move from "20 of 21" to "1000 of 1001".

## Tests

`apps/api/test/integration/routes/model-provider-credentials-discover.test.ts` — the stub gains a request counter and three endpoints: an Anthropic-shaped listing paged on `has_more`/`last_id`, a Gemini-shaped one paged on `nextPageToken`, and a cursor that never ends. Four new cases pin that both pages come back, that a one-page listing still spends exactly one request, that a complete paginated listing spends exactly one request per page and no more, and that the page cap reports `truncated: true`.

```
TEST_TIER=0 bun test apps/api/test/integration/routes/model-provider-credentials-discover.test.ts   → 17 pass / 0 fail
TEST_TIER=0 bun test .../routes/model-provider-credentials-discover.test.ts .../services/model-discovery.test.ts → 26 pass / 0 fail
TEST_TIER=0 bun test apps/api/test/unit/model-listing.test.ts                                        → 27 pass / 0 fail
(from apps/web) bun test src/components/test/model-form-modal.test.tsx                               → 20 pass / 0 fail
(from apps/web) bun test src/lib/test/model-discovery.test.ts src/components/model-form/test/model-pick-list.test.tsx → 13 pass / 0 fail
(from apps/web) bun test src/locales/test/locale-keys.test.ts --timeout 60000                         →  9 pass / 0 fail
bun run verify:openapi → ALL CHECKS PASSED   bun run verify:api-types → up to date   bun run detect:breaking → 0 breaking, 1 non-breaking
bunx tsc --noEmit -p apps/api → clean        bunx tsc --noEmit -p apps/web → clean
```

Negative control: with `MAX_LISTING_PAGES` forced to 1 (the old one-page behaviour), the three pagination cases fail — the tests exercise the loop, not the stub.

## Notes for the orchestrator

- No migration, no env var. `apps/web/src/api/schema.d.ts` is regenerated output (`bun run generate:api`), committed.
- The OpenAPI baseline was **not** regenerated: the added response field is non-breaking and `detect:breaking` passes against the existing baseline.
- **Overlap**: `apps/api/src/routes/model-provider-credentials.ts` (the `/discover` handler — two added lines) also belongs to #1356, and `apps/web/src/components/model-form-modal.tsx` (one `setDiscovery` call site) also belongs to #1355. Both edits are a few lines and additive; expect a trivial conflict at most.
- Two pre-existing harness quirks met while verifying, unrelated to this change: `apps/api/test/integration/services/model-discovery.test.ts` never boots the app, so it only passes alongside a file that does; and `apps/web/src/components/test/model-form-modal.test.tsx` needs `bun test` run from `apps/web` (the `@/` alias does not resolve from the repo root).

### DIFF
--- a/apps/api/src/openapi/paths/model-provider-credentials.ts
+++ b/apps/api/src/openapi/paths/model-provider-credentials.ts
@@ -328,7 +328,7 @@ export const modelProviderCredentialsPaths = {
       tags: ["Model Provider Credentials"],
       summary: "Enumerate the models an endpoint serves",
       description:
-        "Asks an endpoint once for its model listing (`GET <base_url>/models`) and returns the ids it serves, each described with a context window, max output tokens, input modalities and reasoning support. Those come from the listing body itself when the server publishes them per entry (vLLM `max_model_len`, Mistral `capabilities`, OpenRouter `context_length` / `architecture` / `supported_parameters`, LM Studio `max_context_length`) — read from the response already in hand, nothing else is requested — and from the vendored pricing catalog otherwise; `source` says which described a given model. `label` always comes from the catalog. Unlike `POST /{id}/refresh-models` this works BEFORE a credential exists — the operator supplies `provider_id` + `api_key` inline — and it **persists nothing**: no credential is created, no `available_model_ids` is written. Per-token cost is deliberately never returned: an endpoint serving a vendor's model id is not billed at the vendor's rate. Only providers with `authMode: api_key` are accepted — a subscription (OAuth) token is never read or spent to enumerate models. Rate limited to 6 requests per minute.",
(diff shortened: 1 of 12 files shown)


pierrecabriere:

Issue #1344, opened 2026-09-10
security(connect): releaseJti runs after a real upstream DCR registration, making the unauthenticated start route a spam amplifier


pierrecabriere:
## Summary

The hosted connect portal releases its single-use JTI for **every** 4xx, including the auto-DCR path that has already registered a real client at the third-party authorization server. Combined with an unauthenticated, unrate-limited route, that is an unbounded DCR-spam amplifier attributable to the deployment.

Introduced by #1302, not yet deployed.

## Mechanism

`apps/api/src/routes/integrations.ts:965`:

```ts
if (err instanceof ApiError && err.status >= 400 && err.status < 500) {
  logger.warn("Hosted connect OAuth begin refused", {...});
  await releaseJti(claims.jti);
  return c.html(popupHtmlError(err.message, {}), err.status as ContentfulStatusCode);
}
```

But the auto-DCR branch performs a real registration at the third-party AS (`apps/api/src/services/integration-connections.ts:1625`) and then throws a 403 when the AS returns a confidential client — the code's own comment at `:1652` says *"The upstream registration is abandoned unused."*

`packages/core/src/connect-session.ts:106-107` warns explicitly: *"Never call this after a partial success."*

## Failure scenario

A holder of a 10-minute `connect_url` loops:

```
GET /api/integrations/connect/start?token=…
```

That route is unauthenticated (`apps/api/src/lib/auth-pipeline.ts:449`) and has **no rate limiter**. Each iteration is 3 outbound HTTPS calls plus one orphan client registration at the third party.

## Proposed fix

Release only on the failure that provably precedes egress — the "administrator must register…" 403 — not the whole 4xx class. Add a rate limiter to the unauthenticated start route regardless.

PR #1388, opened 2026-09-10
fix(connect): address the hosted portal's error completions to their package

### LINKED ISSUE #1346: connect: the portal's error completion is context-less and hijacks unrelated connect cards

pierrecabriere:
Closes #1346

## Root cause

`apps/api/src/routes/integrations.ts` — every `/connect/start` failure past
`readConnectToken` passed `{}` as its completion detail while `claims.package_id`
sat a few lines above: the vanished auth, the spent jti, the unconnectable auth,
and both `strategy.begin` outcomes.

A completion that identifies nothing is delivered to **every** waiting surface by
contract — `packages/core/src/connect-handshake.ts:233`, the permissive tail
written for the pages that resolved neither a package nor a state — and both
carriers fan out (`BroadcastChannel` + `postMessage`). So a failing Gmail link
flipped an open ClickUp card into an error naming Gmail: the exact hijack
`packageId` was introduced to prevent.

## Fix

One local `completionDetail = { packageId: claims.package_id }`, minted right
after the claims are read, applied to the five sites past that point. `packageId`
is also the *only* identifier a hosted-connect surface holds — its OAuth state is
minted later, past the redirect — so this is what makes these pages reach the
card that opened the link, and nothing else.

The issue proposed one line at the `begin` refusal. The other four sites are the
same mechanism in the same handler; fixing one would have left four live hijacks,
so all five move together and the repeated literal is bound once.

The two pages *before* the token resolves anything (`Missing connect token`,
`This connect link is invalid or expired.`) keep `{}` — they are the case the
permissive tail exists for. The predicate itself is untouched.

## Tests

- `apps/api/test/integration/routes/integrations-hosted-connect.test.ts` — new
  describe (issue #1346), 4 cases reading the popup's inlined completion payload
  as a browser would: the OAuth-begin refusal, the spent link and the vanished
  integration each name their package; the two pre-claims pages stay
  context-less.
- `packages/core/test/connect-handshake.test.ts` — the "payloads the platform
  emits" matrix pinned the *old* sender (it listed five now-addressed messages as
  context-less). Split into the two that still are, plus a new case asserting the
  rest reach the hosted card and the SPA popup and neither another package's card
  nor a state-only one.

```
TEST_TIER=0 bun test apps/api/test/integration/routes/integrations-hosted-connect.test.ts   → 19 pass, 0 fail
TEST_TIER=0 bun test packages/core/test/connect-handshake.test.ts                           → 35 pass, 0 fail
bunx tsc --noEmit -p apps/api ; -p packages/core                                            → clean
bunx eslint / prettier --check on the 3 touched files                                       → clean
```

Negative control: with `completionDetail` forced back to `{}`, the three
addressed-completion cases fail and the pre-claims one stays green — the shape of
the bug.

## Notes for the orchestrator

- Based on `fix/issue-1345`; targets that branch. It changed the *message* at
  these render sites, this changes the *detail* — same call, different argument,
  no textual conflict expected once #1345 lands.
- No migration, no env var, no OpenAPI contract change (statuses and response
  media types are untouched; the popup body is not schema-described).
- Siblings #1352 and #1344 are untouched. #1344 (`releaseJti` after upstream DCR)
  concerns the same `catch` block in `/connect/start`; this PR only changes the
  second argument of the `popupHtmlError` call inside it, so the overlap is one
  adjacent line.

### DIFF
--- a/apps/api/src/routes/integrations.ts
+++ b/apps/api/src/routes/integrations.ts
@@ -935,6 +935,17 @@ export function createIntegrationsRouter() {
     if (!claims) return c.html(popupHtmlError("This connect link is invalid or expired.", {}), 410);
     const scope = scopeFromClaims(claims);
     const actor = actorFromClaims(claims);
+    // From here on the claims name the integration this link was minted for, so
+    // every completion this handler emits must carry it (issue #1346). A
+    // completion that identifies NOTHING is for everyone by contract
+    // (`completionMatches` — the permissive tail exists for the two pages above,
+    // which resolved neither a package nor a state), and both carriers fan out,
+    // so a context-less error from a failing Gmail link drove an unrelated
+    // ClickUp card into an error naming Gmail. `packageId` is also the only
+    // identifier a hosted-connect surface holds — the OAuth `state` is minted
+    // later, past the redirect — so it is what makes these pages reach the card
+    // that opened them and nothing else.
+    const completionDetail = { packageId: claims.package_id };
     // Resolve the integration BEFORE consuming the jti — if the auth no longer
     // exists, the capability token stays unburned so the caller can retry once
     // the integration is back, rather than being forced to re-mint.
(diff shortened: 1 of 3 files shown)


pierrecabriere:

PR #1390, opened 2026-09-10
fix(cli): keep a malformed syncSpaces from bricking every profile

### LINKED ISSUE #1363: cli: a malformed syncSpaces value in one profile bricks every command for every profile, logout included

pierrecabriere:
Closes #1363

## Root cause

`readSyncSpaces` threw from inside `readConfig`'s per-profile loop (`apps/cli/src/lib/config.ts:137` → `:150-156` before this PR). `readConfig` is the first call of `login`, `logout`, `whoami`, `token`, `api`, `openapi` and every `org` / `space` subcommand, none of which wraps it. A single hand-edited typo in one profile — `syncSpaces = "spc_abc"`, the natural TOML mistake for a flag whose help says "repeatable" — therefore failed every command for every profile, and `logout` could not remove the offending section either. Skipping the row instead was not an option: the profile would read as "not configured", `logout` would answer "already signed out" and leave the bad section in place.

## Fix

Per-profile failure, deferred to the one reader.

- `Profile.syncSpaces` becomes `string[] | InvalidSyncSpaces`; `readSyncSpaces` returns `{ invalid: rawValue }` instead of throwing, so every profile still loads and every command that does not read the list keeps working — `logout` included.
- New `syncSpaceIds(profileName, profile)` is where the malformed case turns into the error, with the same message as before (profile + key + file + example). `selectedSpaces` in `apps/cli/src/commands/skills.ts` reads through it, so `skills sync` — the only consumer — is the only command a typo stops, and it fails loudly rather than silently syncing every reachable space.
- `writeConfig` unwraps the marker back to the raw value, so an unrelated write (`space switch` on another profile) round-trips the user's typo instead of quietly deleting it. Deleting it would widen the next sync from a narrowed list to every space with no message.

The union means the compiler, not a convention, forces the malformed case to be handled: exactly one call site had to narrow.

## Tests

`apps/cli/test/config.test.ts` — the old "rejects malformed selections" test is replaced by three that assert the new contract: every profile stays loadable (across four malformed shapes) while `syncSpaceIds` throws for the offending one; `logout`'s `deleteProfile` removes the bad section; an unrelated `updateProfile` writes the malformed value back verbatim.
`apps/cli/test/skills-multi-context.test.ts` — end-to-end guard that `skills sync` exits with `Invalid syncSpaces for profile "default"` rather than falling through to the every-space default.

```
cd apps/cli && TEST_TIER=0 bun test test/config.test.ts test/skills-multi-context.test.ts   → 45 pass, 0 fail
cd apps/cli && TEST_TIER=0 bun test test/login-org.test.ts test/skills-command.test.ts \
  test/spaces.test.ts test/org-command.test.ts test/space-command.test.ts test/logout.test.ts → 168 pass, 0 fail
bunx tsc --noEmit -p apps/cli   → clean
bunx eslint / prettier on the four touched files → clean
```

## Notes for the orchestrator

- Based on `fix/issue-1362`; PR targets that branch. The overlap with #1362 is `apps/cli/src/commands/skills.ts`: two lines inside `selectedSpaces` (`profile.syncSpaces` → `syncSpaceIds(...)`), a different function from the `reachableSpaces` #1362 added.
- The exported `InvalidSyncSpaces` type and `syncSpaceIds` are new public members of `apps/cli/src/lib/config.ts` — internal to the CLI app, not `@appstrate/core`, so no lockstep or publish implication.

### DIFF
--- a/apps/cli/src/commands/skills.ts
+++ b/apps/cli/src/commands/skills.ts
@@ -10,7 +10,7 @@
  */
 
 import { mapWithConcurrency } from "@appstrate/core/map-with-concurrency";
-import { resolveActiveProfile, type Profile } from "../lib/config.ts";
+import { resolveActiveProfile, syncSpaceIds, type Profile } from "../lib/config.ts";
 import { ApiError } from "../lib/api.ts";
 import { listSpaces, resolveSpaceRef, type Space } from "../lib/spaces.ts";
 import { DEFAULT_IO, type CommandIO } from "../lib/io.ts";
@@ -660,13 +660,14 @@ async function selectedSpaces(
     }
     return [...new Set(chosen.map((space) => space.id))];
   }
-  if (!profile.syncSpaces) return spaces.filter(suppliesSkills).map((space) => space.id);
+  const configured = syncSpaceIds(profileName, profile);
+  if (!configured) return spaces.filter(suppliesSkills).map((space) => space.id);
   // A stored list outlives the grants it was written against. An id this
   // profile no longer reaches is dropped with a note, not a failure: losing
   // access is a decision elsewhere, and it must not break the other spaces.
   const listed = new Map(spaces.map((space) => [space.id, space]));
   const kept: string[] = [];
-  for (const id of profile.syncSpaces) {
+  for (const id of configured) {
     const space = listed.get(id);
     if (space && suppliesSkills(space)) {
       kept.push(id);
--- a/apps/cli/src/lib/config.ts
+++ b/apps/cli/src/lib/config.ts
@@ -32,7 +32,41 @@ export interface Profile {
   email: string;
   orgId?: string;
   spaceId?: string;
-  syncSpaces?: string[];
+  syncSpaces?: string[] | InvalidSyncSpaces;
+}
+
+/**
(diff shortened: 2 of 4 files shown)


pierrecabriere:

Issue #1368, opened 2026-09-10
view-as: identity-derived capabilities are not previewable, so a preview shows more than the role does


pierrecabriere:
## Summary

"View as role" cannot preview capabilities that are gated on **identity** rather than on role, so a preview shows more than the previewed role really sees. This is by design on the server — the question is what the UI should do about it.

## Mechanism

The doctrine is explicit at `apps/api/src/lib/view-as.ts:7`: a persona is a pure restriction of the caller's own session, `c.get("user")` stays real, every step is a refusal. So anything gated on "you created this" survives the preview:

- **Runs** — `runs:read` means "the runs I launched" (`apps/api/src/lib/run-visibility.ts:70`). An owner previewing `viewer` sees **their own** runs; a real viewer sees none.
- **Files** — deleting is preset admin **or the file's creator** (`apps/api/src/services/files.ts:328`). The delete control stays on files the previewer created.
- **Integration connections** — `isConnectionOwnedBy(c, user?.id)` (`apps/web/src/pages/integration-detail.tsx:1221`), same effect.

## Why it is worth fixing anyway

The feature's promise is "see what this role sees". A preview that shows a populated Runs list where a real viewer sees an empty one, and offers Delete on files a real viewer cannot touch, is misleading in exactly the situation it exists for — checking a role before handing it to someone.

## Options

- Suppress identity-derived capabilities while a persona is active (the banner already marks the state), and label the suppression so it is not read as a bug.
- Or state the limitation in the preview banner: "runs and files you created remain visible".

Either is better than the current silent divergence. Related: the stale-permissions issue on the same feature.

pierrecabriere:
Shipping option 2 (state the limitation) in #1415: the banner now carries one muted line — *« Ce que vous avez créé reste visible — vos runs, vos fichiers, vos connexions. Ce rôle en voit moins. »* — under the role sentence, for every persona shape.

Option 1 (suppress identity-derived capabilities while a persona is active) is deliberately **not** in that PR. It is a product decision, not a bug fix: the SPA would have to hide data the server legitimately returned, one rule per surface (runs list, file Delete/Keep, connection ownership), and it cannot be made exact — the client cannot tell "returned because I own it" from "returned because the role may read it" without a new server signal on each of those payloads. If that is wanted, it deserves its own issue with that signal specced.

[...]

PR #1391, opened 2026-09-10
fix(rbac): withhold locked input values from a summary agent read

### LINKED ISSUE #1338: security(rbac): the runner preset reads the space's locked input values

pierrecabriere:
Closes #1338

## Root cause

`apps/api/src/routes/agent-detail-handler.ts:189-193` emitted the space's stored
input layer in full — `values` and `locked_fields` — to every caller, with the
comment stating the choice was unconditional. A summary read (`agents:run`
without `agents:read`, the `runner` preset added by #1307) therefore received the
editor's value behind every locked field. `locked_fields` exists so the launcher
cannot *set* those fields (the run refuses one with 400 `locked_input_field`);
the summary let it *read* them.

## Fix

One conditional in the detail serializer: a summary read gets
`withoutLockedFields(storedValues, lockedFields)` instead of `storedValues`. The
helper is the existing `@appstrate/core/input-resolution` one that the launch
form, the scheduler and the input parser already share — no new helper.

The locked field NAMES stay in `locked_fields`, deliberately: they are what tells
the launch form to render the field as imposed (`AgentInputForm`'s `locked`
state) instead of asking for it, and the form already handles a locked field with
no value behind it (it renders `—`). Design choice: filter the values, not the
names — withholding the names would make the runner's form prompt for a field the
run would then refuse.

Scope checked and deliberately left alone: the agent LIST projection emits no
stored values, and `GET /api/spaces/{spaceId}/packages/{scope}/{name}/run-config`
— the other reader of `space_packages.input_settings` — is guarded by
`requirePermission("agents", "read")`, which a runner does not hold.

The OpenAPI description of `AgentDetail.input` gained the carve-out sentence
(same style as the `dependencies` members), so `apps/web/src/api/schema.d.ts` was
regenerated. Description-only: `detect:breaking` reports no change against the
baseline.

## Tests

`apps/api/test/integration/routes/runner-preset.test.ts` — the existing suite
asserted the leaking shape (`values: { tone: "concise" }` for the runner), so it
is updated rather than duplicated. The fixture now stores one locked value
(distinctive, `ADMIN-ONLY-HOUSE-STYLE`) next to one unlocked one, which proves
both halves: the unlocked stored default still reaches the runner, the locked one
does not — asserted on the parsed body, on `Object.hasOwn`, and on the raw
response text. The operator control asserts the editor's read still carries it,
and the `X-View-As` case covers the previewed persona.

- `set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/integration/routes/runner-preset.test.ts` → 12 pass, 0 fail (92 expects)
- negative control (fix reverted, test kept) → 2 fail, confirming the assertions bite
- `... bun test apps/api/test/integration/routes/{agents,spaces-end-user-space-scope,run-input-resolution}.test.ts` → 57 pass, 0 fail
- `bunx tsc --noEmit -p apps/api` → clean; `bunx tsc --noEmit -p apps/web` → clean
- `bun run verify:openapi` → ALL CHECKS PASSED; `bun run verify:api-types` → up to date after `bun run generate:api`; `bun run detect:breaking` → 0 breaking, 0 non-breaking
- `bunx eslint` + `bunx prettier --check` on the touched files → clean

## Notes for the orchestrator

- Based on `fix/issue-1313`; PR targets that branch. No overlap with it — #1313
  changed `apps/api/src/middleware/guards.ts` (`pinnedSpaceScopeGuard`), this
  changes only the detail projection. The one shared file is
  `runner-preset.test.ts`, which #1313 did not touch (its test landed in
  `spaces-end-user-space-scope.test.ts`).
- Siblings #1341 / #1339 / #1368 untouched.

### DIFF
--- a/apps/api/src/openapi/schemas.ts
+++ b/apps/api/src/openapi/schemas.ts
@@ -621,7 +621,7 @@ export const schemas = {
           },
         ],
         description:
-          "AFPS schema wrapper for the agent's parameters, plus the per-space stored values and field locks. Resolution order at launch: author default (JSON Schema `default`) < stored value (`values`) < schedule value < caller input. A field named in `locked_fields` is not asked at launch and a caller that sets it is refused with 400 `locked_input_field`.",
+          "AFPS schema wrapper for the agent's parameters, plus the per-space stored values and field locks. Resolution order at launch: author default (JSON Schema `default`) < stored value (`values`) < schedule value < caller input. A field named in `locked_fields` is not asked at launch and a caller that sets it is refused with 400 `locked_input_field`. A summary read (`agents:run` without `agents:read`) still receives every locked field's NAME, but `values` carries no entry for one — a field the launcher cannot set is not one it reads the stored value of.",
       },
       output: {
         type: "object",
--- a/apps/api/src/routes/agent-detail-handler.ts
+++ b/apps/api/src/routes/agent-detail-handler.ts
@@ -23,6 +23,7 @@ import { getLastRun, getRunningRunsForPackage } from "../services/state/runs.ts"
 import { getInstalledPackageSettings } from "../services/space-packages.ts";
 import { resolveRunTimeout } from "../services/run-limits.ts";
 import { isToolsWildcard, parseManifestIntegrations } from "@appstrate/core/dependencies";
+import { withoutLockedFields } from "@appstrate/core/input-resolution";
 import { parseScopedName } from "@appstrate/core/naming";
 import { getItemId } from "./packages.ts";
 import { notFound } from "../lib/errors.ts";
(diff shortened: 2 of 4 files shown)


pierrecabriere:

PR #1392, opened 2026-09-10
fix(module-ee): widen cost_credits to bigint so an overflow cannot wedge the sweep

### LINKED ISSUE #1330: ee(billing): cost_credits is an int4 whose overflow throws inside the sweep transaction, wedging billing fleet-wide

pierrecabriere:
Closes #1330

## Root cause

`ee_usage_records.cost_credits` was declared `integer` (`packages/module-ee/drizzle/schema.ts:101`) while holding a **cumulative** credit total per usage context. The durable `unattributed` bucket is one row per org that never resets, so that total grows for the lifetime of the deployment and an int4 ceiling of 2 147 483 647 credits (~$2.15M) is reachable.

The overflow was not one org's problem. `billLedgerRows` writes the column as `round((cost_usd + delta) * 1000)::int` and derives the org debit from a `RETURNING` expression over the same cast (`packages/module-ee/src/billing/usage-recorder.ts:432` and `:436`) — both **inside** the transaction that claims the ledger rows. An `integer out of range` there rolls the whole pass back: no row claimed, no watermark advanced. Billing stops fleet-wide, on every tick, until someone finds the offending row.

## Fix

- **EE migration 0007** (`0007_bigint_cost_credits.sql`): `ALTER TABLE ee_usage_records ALTER COLUMN cost_credits SET DATA TYPE bigint`. Pure DDL, no data repair — the USING conversion is implicit and lossless. int4 => int8 is a table rewrite, so it lands now, while every stored value still fits in four bytes; the same migration after an overflow is an incident.
- **Schema**: `bigint("cost_credits", { mode: "number" })`. `mode: "number"` keeps the TS type a plain number and keeps every existing `record.costCredits` assertion working; a JS number is exact to 2^53 credits (~$9e12), far past any real total.
- **Sweep**: both casts become `::bigint`. Widening the column alone would not have helped — `round(numeric)::int` overflows at 2^31 regardless of the column's width, and that cast is the actual throw site (see the negative control below). The `RETURNING` delta already goes through `Number()`, which is exactly what postgres.js needs for an int8 it renders as text.

Deliberately **not** widened: `ee_billing_accounts.credits_used`. It is reset to 0 on every renewal (`src/stripe/webhooks.ts:418`), so it is period-scoped rather than unbounded — a different question from the one this issue raises.

No refactor: three source lines plus the migration. The rest is comments and tests.

## Tests

Two regressions, one per half of the fix.

- `packages/module-ee/test/integration/services/billing-sweeper.test.ts` — seeds an `unattributed` bucket already at 3 000 000 000 credits ($3M cumulative), sweeps one $0.05 row, and asserts the pass commits, only the 50-credit delta is debited, and the record lands at 3 000 000 050.
- `packages/module-ee/test/integration/services/migrate.test.ts` — a `0006 → 0007` upgrade case: the column is `integer` before and `bigint` after, a pre-existing 2 000 000 000-credit row survives intact, a cumulative past the int4 ceiling then accumulates, and re-applying is a no-op. The file's journal-count assertion moves 7 → 8, which is what proves the migration is self-applied at `init()`.
- `packages/module-ee/test/helpers/seed.ts` — `seedUsageRecord` gains `costUsd`, defaulted to the whole-dollar equivalent of `costCredits`, so a seeded row satisfies the invariant the sweep reads (`cost_credits === round(cost_usd * CREDITS_PER_DOLLAR)`) instead of leaving the column at 0. Three existing callers are unaffected (they sum or delete the rows).

Commands (from the worktree root, serialised on the shared Postgres):

```
pg-test-lock.sh sh -c 'set -a && . ./.env && set +a && bun test packages/module-ee/test/integration/services/{billing-sweeper,migrate,repair-account,post-signup,org-drain}.test.ts'
```

- `billing-sweeper` + `migrate` → **72 pass, 0 fail**
- `billing-sweeper` + `repair-account` + `post-signup` + `org-drain` → **104 pass, 0 fail**
- `bunx tsc --noEmit -p packages/module-ee` → clean
- `bunx eslint` + `bunx prettier --check` on the five touched TS files → clean
- `bun run verify:no-migration-dml` → green (69 files, 2 trees)

**Negative control**: reverting just the two casts to `::int` makes the new sweeper test fail with the production symptom, `PostgresError: integer out of range`, on the exact `insert … on conflict … returning` statement. Restored and re-run green afterwards.

## Notes for the orchestrator

- **Migration**: EE **0007** (the index reserved for this issue), `packages/module-ee/drizzle/migrations/0007_bigint_cost_credits.sql`, with `meta/0007_snapshot.json` and the `_journal.json` entry written by hand (`db:generate` needs a TTY). The snapshot diffs against 0006 in exactly three places: `id`, `prevId`, and `cost_credits.type`. `when` is a real current timestamp, strictly greater than 0006's — a `when` at or below the last applied `created_at` is silently skipped by the migrator.
- **Deploy ordering**: self-applied at module `init()` from the folder, like every EE migration. It is a table rewrite (`ACCESS EXCLUSIVE` for its duration), but `ee_usage_records` is small at present, so it is not a lock-window concern today. No env var, no platform migration, nothing to run by hand.
- **Base**: built on `fix/issue-1328`; the chain's own commits are untouched.
- **Overlap**: `packages/module-ee/src/billing/repair-account.ts` reads `SUM(cost_credits)` and is **sibling #1329's** territory — deliberately not touched. The SUM already returned a wider type than `credits_used` (`SUM(int4)` is `int8`), so widening the column to `bigint` changes nothing about that assignment cast. `usage-recorder.ts` is also touched by the #1326/#1328 chain, but in `billing-sweeper.ts` / `env.ts` and the batch-size logic — not in `billLedgerRows`, so no textual overlap.
- **Open decision**: none.

### DIFF
--- a/packages/module-ee/drizzle/migrations/0007_bigint_cost_credits.sql
+++ b/packages/module-ee/drizzle/migrations/0007_bigint_cost_credits.sql
@@ -1,0 +1,18 @@
+-- Widen the cumulative credit total so it cannot overflow mid-sweep.
+--
+--   ee_usage_records.cost_credits — integer => bigint.
+--
+-- `cost_credits` is CUMULATIVE per usage context, and the `unattributed` bucket
+-- is one row per org that never resets, so an `integer` ceiling of 2 147 483 647
+-- credits (~$2.15M) is reachable. The sweep writes it as
+-- `round((cost_usd + delta) * 1000)` and derives the org debit from a RETURNING
+-- expression over the same column, both INSIDE the transaction that claims the
+-- ledger rows: an `integer out of range` there rolls the whole pass back, so a
+-- single overflowing row stops billing for EVERY tenant on every tick.
+--
+-- Cheap now, an incident later: int4 => int8 is a table rewrite, and this widens
+-- a column while every value in it still fits in four bytes.
+--
+-- No DML: the USING conversion is implicit and lossless (every int4 is an int8).
+-- Re-runnable — re-applying the same type is a no-op.
+ALTER TABLE "ee_usage_records" ALTER COLUMN "cost_credits" SET DATA TYPE bigint;
--- a/packages/module-ee/drizzle/migrations/meta/0007_snapshot.json
+++ b/packages/module-ee/drizzle/migrations/meta/0007_snapshot.json
@@ -1,0 +1,459 @@
+{
+  "id": "5f1ec2b5-804b-4f15-a818-4d4ab6053882",
+  "prevId": "9af6f70f-1556-45ac-b6a7-57f9d9986e7a",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.ee_billing_accounts": {
+      "name": "ee_billing_accounts",
+      "schema": "",
+      "columns": {
+        "org_id": {
+          "name": "org_id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "stripe_customer_id": {
+          "name": "stripe_customer_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "stripe_subscription_id": {
+          "name": "stripe_subscription_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "plan_id": {
+          "name": "plan_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "'free'"
+        },
+        "credits_used": {
+          "name": "credits_used",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "default": 0
+        },
+        "credit_quota": {
+          "name": "credit_quota",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "default": 0
+        },
+        "period_end": {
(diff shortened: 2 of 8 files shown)


pierrecabriere:

Issue #1371, opened 2026-09-10
rbac: DELETE /api/roles/:id 500s on a concurrent assignment instead of returning 409


pierrecabriere:
## Summary

`DELETE /api/roles/:id` counts assignments without a lock and issues the delete with no `.catch`, so a concurrent assignment turns the intended 409 `role_in_use` into an uncaught Postgres `23503` and a 500.

## Mechanism

`apps/api/src/services/space-roles.ts:226-263` — the pre-count is lock-free and the `db.delete` is unguarded. A racing `POST /api/spaces/:id/members {custom_role_id}` between the two produces the foreign-key violation.

The non-racing path is correct — executed, clean 409 with counts.

## Proposed fix

Catch `23503` and map it to the same 409 the pre-count produces, or take the count and the delete in one transaction with the appropriate row lock.

adityawaghamare04:
### 🔀 Pull Request Created
I have opened a Pull Request with the verified solution patch: [https://github.com/appstrate/appstrate/pull/1374](https://github.com/appstrate/appstrate/pull/1374)

### Proposed Solution & Patch
## 🛠️ Proposed Solution (by Aditya Waghamare)

### Analysis
`DELETE /api/roles/:id` in `apps/api/src/services/space-roles.ts` performs a lock-free pre-count of assignments followed by an unguarded `db.delete`. If a concurrent request (`POST /api/spaces/:id/members`) inserts an assignment in the window between the check and the delete, Postgres throws a foreign key violation (`23503`), resulting in an unhandled rejection and a 500 error instead of the intended 409 `role_in_use`.

### Fix
Wrap the deletion in a database transaction with a row-level lock (`FOR UPDATE`) or catch the Postgres foreign-key violation code (`23503`) and throw/return a 409 `role_in_use` error response matching the pre-count validation branch.

### Implementation
```typescript
// apps/api/src/services/space-roles.ts

export async function deleteSpaceRole(roleId: string): Promise<void> {
  try {
    await db.transaction(async (trx) => {
      // Re-verify assignment count with pessimistic locking inside transaction
      const assignments = await trx
        .select()
        .from(spaceMembers)
        .where(eq(spaceMembers.customRoleId, roleId))
        .for('update');

      if (assignments.length > 0) {
        throw new HttpError(409, 'role_in_use', 'Cannot delete role with active assignments');
      }

      await trx
        .delete(spaceRoles)
        .where(eq(spaceRoles.id, roleId));
    });
[...]

Issue #1334, opened 2026-09-10
rbac: the custom_roles feature gate is fail-open on member assignment and fail-closed on cleanup


pierrecabriere:
## Summary

`assertCustomRolesFeature()` guards `POST/PATCH/DELETE /api/roles` and the view-as preview, but **not** member assignment. With the feature off, existing bundles still grant, the role picker still offers them, and a brand-new member can be assigned one — while the verbs that would clean them up are refused.

## Mechanism

The gate lives at `apps/api/src/services/space-roles.ts:59`. `assignmentColumns` (`apps/api/src/services/space-members.ts:265`), reached from `routes/spaces.ts:463` and `:488`, does not call it.

## Executed

With `custom_roles` off — the EE → OSS downgrade case, leftover `space_roles` rows:

```
GET  /api/spaces/:id/roles                              → 200, bundles offered
POST /api/spaces/:id/members {custom_role_id: srl_…}    → 201, assigned to a new member
PATCH  /api/roles/:id                                   → 403 feature_unavailable
DELETE /api/roles/:id                                   → 403 feature_unavailable
```

The verb that **propagates** is open; the verbs that **clean up** are blocked. That is the wrong way round.

## Proposed fix

Apply `assertCustomRolesFeature()` on assignment, and decide explicitly what an OSS downgrade does with rows that already carry a custom role — resolve them, or refuse them, but say which. Today the answer differs per verb.

Issue #1333, opened 2026-09-10
security(rbac): nothing bounds the target of a space-member removal or role change


pierrecabriere:
## Summary

`assertCanGrantSpaceRole` bounds only the role being **granted**. Nothing bounds the **target** of a removal or a role change, so a holder of a narrow custom bundle can eject or demote a preset-`admin` member of the same space.

Introduced by the granular-RBAC work (#1260), not yet deployed.

## Mechanism

- `DELETE /api/spaces/:id/members/:userId` (`apps/api/src/routes/spaces.ts:521`) — the "granted" role checked is the **post-removal implicit** role. In a `closed` or `private` space that is `null`, `spacePermissions(null)` is empty, and `assertCanGrantSpaceRole` (`apps/api/src/lib/space-role-policy.ts:25`) passes trivially.
- `PATCH /api/spaces/:id/members/:userId` (`:488`) — no target check at all, in any visibility.

## Executed

Attacker holds a custom bundle `{space-members:remove, space-members:read}` in a `closed` space:

```
DELETE /api/spaces/<closed>/members/<space-admin>  → 200 {"access_after":"none"}
DELETE /api/spaces/<private>/members/<space-admin> → 200
```

With `{space-members:change-role}` only:

```
PATCH /api/spaces/<id>/members/<space-admin>  → 200   (demoted to a minimal custom role)
```

In an `open` space the removal is refused with 403 — purely because the open default happens to grant something. The protection is an accident of visibility.

## Proposed fix

Bound the target as well as the grant: a caller must hold at least what the target holds before removing or changing them, the same rule `canGrantSpaceRole` applies to the role being handed out. The `access_after` computation already loads the target's role — the check has the data it needs.

PR #1393, opened 2026-09-10
fix(roles): report what a custom bundle grants, and name what it only spells

### LINKED ISSUE #1343: rbac: a custom bundle reports permissions it does not grant, and cannot be edited as stored

pierrecabriere:
Closes #1343

## Root cause

`toWire` returned `space_roles.permissions` verbatim (`apps/api/src/services/space-roles.ts:82`) while resolution narrowed the same array through `knownSpaceLevelPermissions()` (`apps/api/src/lib/space-role.ts:63`). One stored array, two answers: `GET /api/roles` listed a permission no holder was granted, and the array it listed was one `normalizePermissions` refused — so a bundle degraded by a module leaving `MODULES` (or a fleet whose replicas disagree on it) could not be edited as it was read.

## Fix

The narrowing happens **once** — `partitionSpacePermissions(stored)` in `apps/api/src/lib/permissions.ts` — and both sides read it:

- `spacePermissions` (enforcement) takes `.granted`, replacing its own copy of the loop.
- `toWire` reports `permissions` = `.granted`, so what an admin reads is what a holder is granted **and** is an array `PATCH` accepts back.
- What the row spells and this deployment cannot name is **not hidden**: it is a new required response field `unavailable_permissions` (always `[]` for a preset). Filtering alone would have removed the only signal an operator has that a bundle is degraded — and would have made the SPA's "Unavailable" panel (added one PR ago in #1288) dead code.

The stored row is deliberately left alone: a rename issued against a replica that lost a module must not erase the org's intent for the replicas that kept it. Sending back a `permissions` array is what drops a ghost from the row — the same explicit act the SPA's Remove button used to perform, now performed by saving.

The second-order effect the issue raises is **kept as-is and documented at the decision site** (`spacePermissions`): a bundle whose every string became unknown grants nothing and does **not** fall back to an open space's default role. Falling back would hand a holder *more* than their bundle names on the very replica that understands it least.

SPA: the editor no longer re-derives the unavailable set from the vocabulary (a second answer to a question the server now answers); `apps/web/src/lib/role-permissions.ts` is deleted, the Remove control goes with the derivation, and the panel's hint says what saving does to them (en + fr).

## Tests

- `apps/api/test/integration/routes/roles.test.ts` — new case `separates what a bundle grants from what it merely spells, and round-trips`: seeds a row holding `ghostmodule:write` straight to the table (the write route refuses it), asserts the split on a partly- and a fully-unknown bundle, asserts a preset reports `[]`, then PATCHes the listed array back (200, ghost dropped) **and** its permitted-twin refusal — PATCHing the unknown string is still a 400, so the round-trip comes from the projection agreeing, not from validation going soft.
- Negative control: reverting only the `toWire` line makes the new case fail (`1 fail`), and the fix restores it.
- `apps/web/src/pages/test/role-editor-unavailable-permissions.test.tsx` rewritten for the new panel; fixtures in `assignable-space-roles.test.tsx` and `view-as-entry-points.test.tsx` carry the new field.

```
  apps/api/test/integration/routes/roles.test.ts apps/api/test/unit/permissions.test.ts \
  apps/web/src/pages/test/role-editor-unavailable-permissions.test.tsx \
  apps/web/src/hooks/test/assignable-space-roles.test.tsx \
  apps/web/src/pages/test/view-as-entry-points.test.tsx     → 64 pass, 0 fail
… TEST_TIER=0 bun test apps/api/test/unit/lib/view-as-callsites.test.ts \
  apps/api/test/unit/lib/deferred-assignment-authority.test.ts \
  apps/api/test/integration/routes/space-member-management.test.ts → included above, 0 fail
… TEST_TIER=0 bun test --timeout 60000 apps/web/src/locales/test/locale-keys.test.ts → 9 pass, 0 fail
bun run verify:api-types        → up to date (after `bun run generate:api`, committed)
bun run detect:breaking         → PASSED — 0 breaking, 2 non-breaking
bunx knip                       → no unused/unresolved issues
bunx eslint / prettier on every touched file → clean
```

`locale-keys.test.ts` fails at the default 5 s timeout under the parallel fleet load; it passes at `--timeout 60000`.

## Notes for the orchestrator

- Based on `fix/issue-1342` and reuses its `normalizePermissions` unchanged (write-time bound + dedupe stays the only writer-side rule).
- **Wire contract grows**: `RoleObject.unavailable_permissions` is a new *required* response field on `GET /api/roles`, `GET /api/spaces/{id}/assignable-roles`, `POST /api/roles`, `PATCH /api/roles/{id}`. `detect:breaking` classes it non-breaking, so `openapi/baseline.json` is untouched; `apps/web/src/api/schema.d.ts` is regenerated and committed. Any other consumer constructing a `RoleObject` literal must add the field.
- **Overlap**: none with #1334 / #1371 / #1333. #1334 gates the `custom_roles` feature and #1371 touches `DELETE /api/roles/:id`; this PR touches neither `hasCustomRoles`/`assertCustomRolesFeature` nor `deleteSpaceRole` (its `toWire(row)` return picks up the new projection with no code change).
- Deleted file: `apps/web/src/lib/role-permissions.ts` (its only consumer and its only test were rewired here).


pierrecabriere:

Issue #1332, opened 2026-09-10
ee(stripe): the live contract suite missed this bump, cannot cover a Dependabot bump, and never touches Invoice


pierrecabriere:
## Summary

The live Stripe contract suite added by #1303 did not cover the 18.5.0 → 22.6.1 bump, cannot cover a future Dependabot bump, duplicates the `apiVersion` literal in a file the typechecker never sees, and never touches `Invoice` — the surface carrying the *other* post-basil relocation.

## 1. It did not run against this bump, and cannot run against the next

Ordering: `3f2876129` (the bump) merged 2026-09-09 10:28 UTC; `0fbf1c414` (the workflow) at 12:18 UTC. 18.5.0 → 22.6.1 was **never** run against live Stripe.

Going forward: `.github/workflows/stripe-live.yml:24-27,75` reads `${{ secrets.STRIPE_LIVE_SECRET_KEY }}` on a `pull_request` event. GitHub does not expose repository secrets to runs triggered by Dependabot PRs — those see only the Dependabot secret store. So on a Dependabot `stripe` bump the variable is empty, `describe.skipIf(!SECRET_KEY)` skips all 26 tests, and the job is **green**. The workflow's own comment claims the opposite:

> A Dependabot bump of `stripe` edits packages/module-ee/package.json, so it lands here without anyone remembering to label it.

## 2. The apiVersion literal sits outside the typecheck

`packages/module-ee/test/live/stripe-contract.test.ts:118`:

```ts
stripe = new Stripe(SECRET_KEY!, { apiVersion: "2026-08-26.dahlia" });
```

`packages/module-ee/tsconfig.json` has `"include": ["src", "drizzle"]` — `test` is absent, and module-ee is the outlier here (core, db, env, emails, afps-shared, afps-runtime, shared-types, runner-pi, mcp-transport and apps/cli all include it).

`apiVersion?: LatestApiVersion` is a *literal* type, so a mismatched literal is a hard `TS2322` — that error is exactly what forced `src/stripe/client.ts:11` to change in this bump. The test file gets no such push: at the next bump, `tsc` drags `client.ts` forward, the test literal stays behind, and the suite goes green having validated the shapes of an API version production no longer speaks.

Note: adding `test` to the program is not a one-line fix — it currently yields 9 pre-existing errors elsewhere in the test tree.

## 3. Invoice has no coverage in either suite

[...]

PR #1394, opened 2026-09-10
fix(ee): narrow customer.source.expiring to an actual card

### LINKED ISSUE #1331: ee(stripe): 'as Stripe.Card' renders "undefined/ed" in the card-expiring email, and two casts blind the 22.x relocations

pierrecabriere:
Closes #1331

## Root cause

`packages/module-ee/src/stripe/webhooks.ts` (the `customer.source.expiring` arm) cast `event.data.object` to `Stripe.Card`. In stripe 22.6.1 that payload is a `CustomerSource` — `Account | BankAccount | Card | Source` (`stripe/cjs/resources/Events.d.ts:977`, `CustomerSources.d.ts:5`). Only `Card` carries `exp_month` / `exp_year`, so for a legacy `Source` the template computed `String(undefined).padStart(2, "0")` + `String(undefined).slice(-2)` and mailed the customer `"undefined/ed"` as their card expiry.

Two neighbouring casts in the `customer.subscription.updated` arm were unnecessary in 22.6.1 and actively harmful: `event.data.previous_attributes` is typed `Partial<Subscription>` upstream, so `as { previous_attributes?: Record<string, unknown> }` and `previousAttributes.items as { data?: … }` erased the compiler's ability to flag a relocation of `items[].price.id` — exactly what an SDK major moves.

## Fix

- Narrow on `source.object !== "card"` and `break` with one debug line. Nothing is lost: the other three members have no expiry to announce, so there is no email to send. With the payload correctly typed, `source.last4 ?? "????"` was dead (`Card.last4: string`, non-nullable) and the `??` is gone.
- Drop both `previous_attributes` casts. `previousAttributes.items?.data[0]?.price.id` now typechecks against the real `ApiList<SubscriptionItem>`, so the next relocation is a compile error instead of a silent `undefined`.

No behaviour change for the card path — same email, same values.

## Tests

`packages/module-ee/test/integration/services/webhook-emails.test.ts` gains a `customer.source.expiring` block (two cases: a `card` payload renders `03/27` and contains no `"undefined"`; a `source` payload sends nothing). The file's captured-email harness now records `html` alongside `to`/`subject`, which is what lets the expiry be asserted at all.

```
pg-test-lock.sh … bun test packages/module-ee/test/integration/services/webhooks.test.ts \
                            packages/module-ee/test/integration/services/webhook-emails.test.ts
→ 44 pass, 0 fail, 119 expect() calls
bunx tsc --noEmit -p packages/module-ee → clean
bunx eslint + prettier on both touched files → clean
```

Negative control: with the `as Stripe.Card` cast restored, the non-card case fails (`9 pass / 1 fail`) — the email goes out with `"undefined/ed"`. Reverted before committing.

## Notes for the orchestrator

- Based on `fix/issue-1324`, so this PR carries that chain's five commits too. Merge order: #1323 → #1325 → #1324 → this.
- Overlap with the sibling chain is confined to `webhooks.ts`; the chain touched the checkout and subscription arms, this touches the `customer.source.expiring` arm and the two casts in `customer.subscription.updated`. No shared lines.
- No migration, no env var, no OpenAPI change. Nothing to order at deploy.
- For #1332 (live contract suite): the `CustomerSource` union is the kind of shape a live contract test would pin — the non-card branch here is asserted only against a synthetic payload.

### DIFF
--- a/packages/module-ee/src/stripe/webhooks.ts
+++ b/packages/module-ee/src/stripe/webhooks.ts
@@ -572,8 +572,7 @@ async function processEvent(event: Stripe.Event): Promise<void> {
       }
       const { orgId } = subMetadata;
 
-      const previousAttributes = (event.data as { previous_attributes?: Record<string, unknown> })
-        .previous_attributes;
+      const previousAttributes = event.data.previous_attributes;
       const itemPeriodEnd = subscriptionPeriodEnd(subscription);
       const newPlan = planForSubscription(subscription, plans);
 
@@ -644,9 +643,7 @@ async function processEvent(event: Stripe.Event): Promise<void> {
         "items" in previousAttributes &&
         !subscription.cancel_at_period_end
       ) {
-        const previousPriceId = (
-          previousAttributes.items as { data?: Array<{ price?: { id?: string } }> }
-        )?.data?.[0]?.price?.id;
+        const previousPriceId = previousAttributes.items?.data[0]?.price.id;
         const oldPlan = Object.values(plans).find((p) => p?.stripePriceId === previousPriceId);
 
         if (oldPlan && oldPlan.id !== newPlan.id) {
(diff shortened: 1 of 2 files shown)


pierrecabriere:

Issue #1361, opened 2026-09-10
cli: self-update resolves 'latest' within one release page and can silently downgrade users


pierrecabriere:
## Summary

`appstrate self-update` picks the highest semver **within one page** of releases and returns on the first page containing any `v*` candidate. After enough non-platform releases accumulate, a hotfix cut for an older line becomes "latest" and every user is silently downgraded — and pinned there.

## Mechanism

`apps/cli/src/lib/self-update.ts:386-405` — the loop `return`s on the first page holding a candidate, so `candidates.sort(compareSemver)` never sees page 2. `RELEASES_PAGE_SIZE = 30` (`:47`).

The comment at `:382` states the exact invariant the code does not hold:

> a hotfix for an older line published after a newer release must not become latest

`test/self-update.test.ts:211` only covers the same-page case. `compareSemver` itself (`:208-256`) is correct, including the §11.4 numeric-vs-alphanumeric rule.

## Failure scenario

`v1.1.0` ships. ≥29 `cli@` / `core@` / `afps-shared@` / dependabot releases accumulate. A `v1.0.1` hotfix is cut for the old line → page 1 holds only `v1.0.1` → every `appstrate self-update` downgrades 1.1.0 → 1.0.1.

## Related: the two install channels disagree

`scripts/bootstrap.sh:145-167` takes the **first `v*` in creation order** within a page (BSD `sort` has no `-V`, acknowledged at `:141`), while `self-update.ts:402` takes the highest semver. After a late-published hotfix, `curl … | bash` installs `v1.0.1` and the next `self-update` immediately replaces it.

Also in that resolver: `grep -oE '"(tag_name|draft|prerelease)": …'` matches inside a release **body**, so notes quoting `"prerelease": true` inject a phantom field the awk state machine mis-pairs; and `[ -z "$fields" ] && return 1` aborts the whole 5-page walk on one transient curl failure, surfacing it as "No platform v* release found" rather than as a network error.

## Proposed fix

Walk every page before sorting (or ask the API for the newest `v*` directly — see #1271, which proposes a signed channel manifest and would resolve both channels at once). Make the two channels share one resolution rule.

Issue #1321, opened 2026-09-10
cli: the @napi-rs/keyring 2.0 bump activates dead error markers — hard refusal instead of file fallback, and logout can leave a plaintext token


pierrecabriere:
## Summary

`@napi-rs/keyring` 2.0.0's breaking change is that `getPassword()` and `deletePassword()` now **throw** where v1 swallowed the error. That activates two `catch` blocks that were previously dead — and their error-string matchers do not match any string the real binary emits, so every store failure classifies as `"broken"`, which hard-refuses and **skips the plaintext file fallback**.

Bump merged in `ebb3764d2` (#1299); no call site was adapted.

## Mechanism

`apps/cli/src/lib/keyring.ts:139`:

```ts
const MISSING_BACKEND_MARKERS = ["Platform secure storage failure", "No storage"];
```

`classifyKeyringError` returns `"broken"` for anything not containing a marker, and `"broken"` calls `refuseBrokenKeyring(...)` — a hard throw before the file fallback at `:334` and `:360`.

`strings` on the installed `keyring.darwin-arm64.node@2.0.0` in this checkout:

```
"Couldn't access platform storage:
Platform failure:
```

Neither `"Platform secure storage failure"` nor `"No storage"` exists in the binary (the only `no storage` present is `"no storage space"`, lowercase, an unrelated io string). `"No matching entry"` is absent too.

## Failure scenarios

**1. CI runner / stripped container with no secret-service daemon.** Under v1, `appstrate api …` read `null` from the keyring and fell through to `~/.config/appstrate/credentials.json` — the documented reason the fallback exists. Under v2 the read throws, classifies `"broken"`, and the command dies with *"the OS keyring is installed but not serving"* unless `APPSTRATE_ALLOW_PLAINTEXT_TOKENS=1` is set.

**2. `appstrate logout` can leave a live plaintext refresh token on disk.** `deletePassword()` throws on a failed delete; the catch classifies `"broken"`; `refuseBrokenKeyring("delete", err)` throws; the final `await deleteFromFile(profile)` is never reached. On a machine in split-brain storage with the keychain locked, v1 cleaned the file store and printed "Signed out"; v2 throws and leaves the 30-day refresh token in `~/.config/appstrate/credentials.json`. Refusing a *delete* to avoid plaintext-on-disk guarantees the plaintext stays.

## The comment is wrong

`apps/cli/src/lib/keyring.ts:~132`:

[...]

PR #1396, opened 2026-09-10
fix(cli): tell a busy skills-sync lock apart from one that cannot work

### LINKED ISSUE #1364: cli: the skills-sync flock never inspects errno, and cannot load at all on Windows

pierrecabriere:
Closes #1364

## Root cause

`apps/cli/src/lib/skills-sync/lock.ts` bound only `flock`'s return value, so the poll loop read *any* non-zero as "another sync holds it" (`lock.ts:65-68` before this change). `ENOLCK` / `EOPNOTSUPP` from a mount without `flock` — NFS without lockd, some 9p / virtiofs container mounts — therefore blocked the marketplace command for the full 60 s and then threw `SyncLockBusyError`, naming a concurrent process that does not exist.

Separately, the libc candidate list (`lock.ts:35-38` before) was glibc/musl/libSystem only. On `win32` all three `dlopen`s failed and the helper threw: `skillsSyncCommand` exited 1 every time, and `logoutCommand` warned and skipped skill cleanup (credentials were still wiped, correctly).

## Fix

`bindFlock` now resolves `__error` (darwin) / `__errno_location` (glibc, musl) alongside `flock` in the same `dlopen`, and reads errno through `read.i32` immediately after a failed call. One attempt yields one of four answers:

- `EWOULDBLOCK` (35 darwin / 11 linux) → **busy**, keep polling until the deadline, then `SyncLockBusyError`. This is now the *only* road to that error.
- `EINTR` → **interrupted**, retry at once without burning a poll interval.
- anything else → **unsupported**, carrying `flock(2) failed with errno N`.
- `win32`, or every libc candidate failing to load → **unsupported** before any syscall; `resolveTryLock` no longer throws.

When the lock is unsupported `withSyncLock` **fails open**: it runs the body unlocked after writing `warning: skills sync lock unavailable (<reason>); continuing unlocked — do not run two syncs at once.` to the caller's `CommandIO`. The lock guards a narrow race (two syncs started in the same window); refusing every sync on those hosts is the larger breakage, and it is what made `skills sync` unusable on Windows and made `logout` skip skill cleanup. Both call sites now pass their `io` so the warning reaches the sink the command already owns.

**No `LockFileEx` path.** Windows takes the same degrade branch as an flock-less mount — one code path, not a Windows-specific one, and no code this project cannot verify. A real Windows binding can land the day someone can test it there.

The FFI seam is injectable (`SyncLockOptions.tryLock`, `resolveTryLock(platform)`) so the classification and the platform gate are testable on any host — DI, no `mock.module()`.

## Tests

`apps/cli/test/skills-lock.test.ts`, +6 cases (5 → 11), extending the existing file:

- errno read against **real libc**: two descriptors on one lock file → `busy`; `fd 9999` (EBADF) → `unsupported`, which is the exact bug — a non-EWOULDBLOCK errno must not enter the poll loop.
- `resolveTryLock("win32")` → `unsupported` without reaching for libc.
- unsupported → body runs, warning on stderr, and with `timeoutMs: 0` it still resolves instead of throwing "another sync is running".
- `interrupted` retries immediately (`pollMs: 60_000`, `timeoutMs: 5_000` — only an immediate retry reaches `acquired`).

```
cd apps/cli && TEST_TIER=0 bun test test/skills-lock.test.ts      → 11 pass, 0 fail (x3 runs, no flake)
cd apps/cli && TEST_TIER=0 bun test test/logout.test.ts test/skills-command.test.ts test/skills-multi-context.test.ts
                                                                  → 108 pass, 0 fail
bunx tsc --noEmit -p apps/cli                                     → clean
bunx eslint <4 touched files> ; bun run knip                      → clean (knip exit 0)
```

Negative controls, both confirmed: collapsing the classification back to "every non-zero is busy" fails the errno case, and removing the `win32` branch fails the Windows case.

## Notes for the orchestrator

- Based on `fix/issue-1363`; PR targets that branch. No overlap with the siblings — #1361 (self-update) and #1321 (`@napi-rs/keyring` markers) touch other files, and #1363's `config.ts` work is untouched here.
- Behaviour change worth calling out in the release notes: on a host where `flock` does not work, `skills sync` now runs (with a stderr warning) instead of failing after 60 s. On every host where it does work — every CI runner, macOS and Linux dev machine — the behaviour is unchanged.

### DIFF
--- a/apps/cli/src/commands/logout.ts
+++ b/apps/cli/src/commands/logout.ts
@@ -49,34 +49,39 @@ export async function logoutCommand(
   };
   let credentialsCleared = false;
   try {
-    await withSyncLock(async () => {
-      try {
-        const tokens = await loadTokens(profileName);
-        hadTokens = !!tokens;
-        const profile = await getProfile(profileName);
-        if (tokens && profile) {
-          await revokeCliRefreshToken(
-            normalizeInstance(profile.instance),
-            CLI_CLIENT_ID,
-            tokens.refreshToken,
+    await withSyncLock(
+      async () => {
+        try {
+          const tokens = await loadTokens(profileName);
+          hadTokens = !!tokens;
+          const profile = await getProfile(profileName);
+          if (tokens && profile) {
+            await revokeCliRefreshToken(
+              normalizeInstance(profile.instance),
+              CLI_CLIENT_ID,
+              tokens.refreshToken,
+            );
+          }
+        } catch (err) {
+          io.stderr.write(
+            `warning: could not revoke refresh token server-side (${formatError(err)}); continuing with local cleanup.\n`,
           );
+        } finally {
+          await clearCredentials();
+          credentialsCleared = true;
         }
-      } catch (err) {
-        io.stderr.write(
-          `warning: could not revoke refresh token server-side (${formatError(err)}); continuing with local cleanup.\n`,
-        );
-      } finally {
-        await clearCredentials();
-        credentialsCleared = true;
-      }
-      const cleanup = await cleanupProfileSkills(profileName);
-      if (cleanup.pluginReset)
-        io.stderr.write(
-          "Appstrate plugin reset. Run `claude plugin update appstrate@appstrate` and restart Claude, or start a new session with automatic plugin refresh enabled.\n",
-        );
-      for (const failure of cleanup.warnings)
-        io.stderr.write(`warning: ${failure}. Retry appstrate logout --profile ${profileName}.\n`);
-    });
+        const cleanup = await cleanupProfileSkills(profileName);
+        if (cleanup.pluginReset)
+          io.stderr.write(
+            "Appstrate plugin reset. Run `claude plugin update appstrate@appstrate` and restart Claude, or start a new session with automatic plugin refresh enabled.\n",
+          );
+        for (const failure of cleanup.warnings)
+          io.stderr.write(
+            `warning: ${failure}. Retry appstrate logout --profile ${profileName}.\n`,
+          );
+      },
+      { io },
+    );
   } catch (err) {
     io.stderr.write(
       `warning: could not complete skills cleanup (${formatError(err)}). Retry appstrate logout --profile ${profileName}.\n`,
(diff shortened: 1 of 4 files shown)


pierrecabriere:

PR #1397, opened 2026-09-10
fix(db): index the OAuth foreign-key columns better-auth declares indexed

### LINKED ISSUE #1348: db(oidc): eight FK columns better-auth declares indexed have no index — every logout seq-scans both token tables

pierrecabriere:
Closes #1348

## Root cause

`@better-auth/oauth-provider` 1.7.3 declares `index: true` on ten FK columns
(`apps/api/node_modules/@better-auth/oauth-provider/dist/authorize-9whjxVLJ.mjs:3672`,
`:3922-3941`, `:4017-4062`, `:4090-4099`). The OIDC tables are hand-written Drizzle
(`packages/db/src/schema/oidc.ts`), not generated by the Better Auth CLI, and nothing
reconciles the two — so none of the ten index was ever declared or created.
`oauth_consents` had no index at all.

PostgreSQL indexes a *referenced* column, never the *referencing* one, so each parent
delete proved "no child points here" with a sequential scan: a logout deletes a
`session` row and both token tables carry `session_id ON DELETE set null`; a user
deletion cascades into all four tables on `user_id`; a client deletion into three on
`client_id`; a refresh-token deletion into `oauth_access_tokens` on `refresh_id`.
`/authorize` read `oauth_consents` by `(client_id, user_id, reference_id)` unindexed
on every request.

## Fix

`packages/db/src/schema/oidc.ts` gains the ten `index()` declarations, and migration
`0062_oauth_fk_indexes.sql` creates them. DDL only — no row written
(`verify:no-migration-dml` green).

Design notes:

- **The index set is exactly the plugin's declarations, no more.** A composite
  `(client_id, user_id)` on `oauth_consents` would serve `/authorize` slightly better
  but leave the `user_id` cascade uncovered and need a third index; two single-column
  indexes cover both cascades and let Postgres BitmapAnd the `/authorize` predicate.
- **Ten, not the nine the issue enumerates.** `oauth_clients.user_id` is declared
  `index: true` upstream (`:3672`), is an `ON DELETE cascade` target of the same user
  deletion, and is unindexed. Leaving it out would have made the invariant — and the
  test below — untrue by one column.
- `CREATE INDEX CONCURRENTLY` is not available: drizzle runs a migration file in one
  transaction. Plain `CREATE INDEX` takes a SHARE lock; these tables hold short-lived
  tokens so the build is seconds, but the migration header says so for operators.
- No boot-time check added — that is #1349's territory (`diffSchema` checks columns,
  not indexes). The guard here is a test, which needs no runtime surface.

## Tests

`apps/api/src/modules/oidc/test/unit/better-auth-declared-indexes.test.ts` (new, 102
lines) reads `index: true` off the **live** plugin objects returned by
`oidcBetterAuthPlugins()` and asserts each declared column leads an index on the
matching Drizzle table — so a plugin upgrade that indexes a new column fails here
rather than passing silently. A companion assertion pins the set of models carrying
declarations, so a plugin that stops exposing `schema` cannot make the per-model
checks vacuous.

- `set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/src/modules/oidc/test/unit/ packages/db/test/` → **136 pass, 2 skip, 0 fail** (17 files)
- Negative control: removing `idx_oauth_consents_user` and `idx_oauth_access_tokens_session` from the schema → **2 fail**, naming the missing columns. Restored.
- Full migration chain applied on PGlite (all 62 files) then `pg_indexes` queried: all ten new indexes present, none duplicated.
- `bunx tsc --noEmit -p apps/api` → clean; `-p packages/db` → clean.
- `bun run verify:no-migration-dml` → green (69 files, 2 trees).
- `bunx eslint` + `bunx prettier --write` on the touched files → clean.

## Notes for the orchestrator

- **Migration index 0062** (platform), as reserved. `_journal.json` carries one entry
  `idx: 62`, `when: 1788888549265`; the snapshot was hand-built from
  `0060_snapshot.json` (diff against it shows index additions only) with `prevId` =
  0060's `id`. **0061 is absent** — a sibling adds it in parallel, so `_journal.json`
  ordering and `0062_snapshot.json`'s `prevId` need reconciling once 0061 lands.
- No env var, no OpenAPI change, no API contract change.
- Deploy ordering: none. The migration is additive and converges from either end
  (`IF NOT EXISTS` on every statement).
- Overlap with siblings: none in code. #1349 covers the boot check that would have
  caught this class; this PR deliberately leaves `diffSchema` alone.
- Based on `fix/issue-1347`; its `jose` override + `better-auth/*` facade routing are
  untouched, and `bun.lock` is not modified.


pierrecabriere:

Issue #1351, opened 2026-09-10
oidc: CIMD/DCR registration discards the declared scope and persists the full self-service ceiling


pierrecabriere:
## Summary

CIMD and DCR registration discard whatever `scope` the client declares and persist the full self-service ceiling instead, so a client is granted scopes it never asked for. Bounded — it can never exceed `getSelfServiceScopes()` — but declared narrowing is currently cosmetic, and the in-tree tests bless it.

## Mechanism

`@better-auth/oauth-provider/dist/authorize-9whjxVLJ.mjs:1934-1937` overwrites the persisted scope with `resolveClientRegistrationScopes(opts)` for both `dynamic` and `clientMetadataDocument` registrations.

A client publishing `scope: "openid"` is stored with `[openid, profile, email, offline_access, mcp:read, mcp:invoke]`.

The ceiling that still applies is `apps/api/src/modules/oidc/auth/scopes.ts:93-95`; core action scopes are rejected. So there is no escalation.

Related, and worth deciding at the same time: the consent screen advertises `mcp:read` / `mcp:invoke` (`apps/api/src/modules/oidc/pages/consent.ts:45-46`), but for instance-level tokens the scope claim is **not** a permission ceiling (`claims.ts:60` returns empty; `auth-pipeline.ts:150-163` writes no `scopeCeiling`). The minted token drives anything the user's live role allows through `invoke_operation`. Confined to the MCP resource and capped by role — not an escalation, but the consent text under-describes reach.

## Note

`markClientSelfService` (`apps/api/src/modules/oidc/services/oauth-admin.ts:426-432`) **lost** its CIMD scope backfill in #1287. That removal is *correct* given the upstream behaviour above, and it closes the old CIMD scope-widening bypass — but it is unexplained in the commit, and the recorded note "CIMD 1st hit = invalid_scope" will tempt someone to restore it. Restoring it would reintroduce a client-controlled scope path.

## Proposed fix

Either honour the declared scope as an upper bound intersected with the ceiling, or state plainly in the docs and consent copy that registration-time scope is advisory. Add a test pinning whichever is chosen, so the backfill is not re-added.

Issue #1350, opened 2026-09-10
docs(migration): 0057's pre-flight query under-counts the clients the 1.7.3 bump breaks, then the migration destroys the evidence


pierrecabriere:
## Summary

The operator pre-flight query documented for migration 0057 under-counts the clients the 1.7.3 bump breaks, and the migration then drops the column needed to enumerate them.

## Mechanism

`scripts/migration/README.md:101-104` counts:

```sql
WHERE token_endpoint_auth_method IS NULL AND "public" = false
```

But `public` was **nullable** (`649277242^:packages/db/src/schema/oidc.ts:104`), and the fold at `0057_oauth_provider_1_7_3.sql:192` requires `"public" IS NOT NULL` — so rows with **both** columns NULL are silently skipped by both, and then `public` / `type` are dropped (`:193-194`), destroying the evidence.

Those rows behave identically to the counted ones: `@better-auth/oauth-provider/dist/utils-oOcdCCVw.mjs:640` reads `client.tokenEndpointAuthMethod ?? "client_secret_basic"` and `:647` then demands a secret, so they get `invalid_client`, unrecoverably.

## Second defect: the comment blames the wrong step

`0057:179-192` says such a client is answered `invalid_client` *"the moment the fold below writes that value in"*. It isn't — a stored `NULL` already reads as `client_secret_basic` under 1.7.3, fold or no fold. **The break arrives with the code, not with the migration.** An operator who rolls the new image first — the normal order, since migrations run at boot — has already broken those clients *and* destroyed the column needed to find them.

## Proposed fix

- Pre-flight becomes `WHERE token_endpoint_auth_method IS NULL`, no `public` predicate.
- The instruction "run BEFORE this migration" becomes "before deploying the 1.7.3 image".
- The count should be preserved as a saved result, not as a comment in a file that deletes its own input.

Fail direction is safe: NULL → `client_secret_basic`, never → `none`. No confidential-to-public downgrade.

Issue #1319, opened 2026-09-10
scripts: operator script 0008 is destructive if re-run, guarded only by a comment


pierrecabriere:
## Summary

`scripts/migration/0008-org-viewer-to-guest.sql` is safe exactly once. Its step-4 capture predicate is **permanent**, not window-scoped, so a second run after the new app is live widens OAuth signup grants.

## Mechanism

`scripts/migration/0008-org-viewer-to-guest.sql:93-95` and step 4 (`:155-159`) capture on:

```sql
signup_role = 'guest' AND signup_space_assignments = '[]'::jsonb
```

After the deploy that is no longer a "rows the rename left empty" predicate — it re-matches any client an admin **deliberately** left with no space assignments, and hands it `viewer` rows in every space created since.

The file's own header and `0012-org-invitation-history-viewer-to-guest.sql`'s preamble both say so; 0012 exists precisely so the follow-up fix is not shipped as "re-run 0008".

## Proposed fix

Make the constraint mechanical rather than documentary — a one-shot marker the script writes and refuses to run past, or a `WHERE updated_at < :cutoff` bound captured at the start of the window. A destructive-if-repeated operator script whose only guard is a comment is a trap for whoever runs the next deploy.

## Ops note

For the current window: run 0008 exactly once, inside the maintenance window, and never again.

Issue #1318, opened 2026-09-10
scripts: migration 0010 refuses after a single EE boot, and its refusal message invites data loss


pierrecabriere:
## Summary

`scripts/migration/0010-ee-tables-into-platform-db.ts` refuses to run once the EE module has booted **even once** against the target, and its refusal message invites exactly the wrong recovery. It also applies the module's migrations *before* copying the rows, so every DML migration in the module's history is a no-op for the moved data.

## Defect 1 — the refusal cannot tell a prior copy from a single boot

`scripts/migration/0010-ee-tables-into-platform-db.ts:373-380`:

```
refusing: the target already holds ee_* rows (ee_billing_cursor).
This script copies, it does not merge — re-running it would double-count billed usage.
If the move already ran, nothing is left to do; otherwise empty the target first.
```

But `packages/module-ee/src/index.ts:139` calls `ensureCursorSeeded` at `init()`, which INSERTs one row into `ee_billing_cursor`. A Coolify restart, a health probe, or an operator starting the stack before the maintenance window is enough.

**Failure**: the operator reads the first branch — "the move already ran, nothing is left to do" — and abandons the source `cloud_*` database holding every billing account, Stripe customer/subscription id and claim. Total loss of billing state while Stripe keeps charging.

None of the 7 cases in `scripts/test/migration-0010-ee-tables-into-platform-db.test.ts` covers a cursor-seeded target.

## Defect 2 — migrations applied before the copy

`0010:404-406`:

```ts
await applyModuleMigrations(targetUrl);
await target.begin(async (tx) => {
  for (const plan of plans) await copyTable(src, tx, plan);
});
```

The script also never reads the source's migration level — it infers shape from column presence only, despite the header asserting "the expected production source is at 0003".

**Failure**: a source below 0003 still has free-plan rows carrying `subscription_status = 'canceled' | 'incomplete_expired'` — exactly what `0003_normalize_free_subscription_status.sql` exists to fix. They land verbatim; `packages/module-ee/src/billing/quota-check.ts:126-131` then rejects every positive-quote operation for those orgs with `402 subscription_blocked`. Free-tier orgs locked out, after a script that printed "copied — every table matches".

## Proposed fix

1. Distinguish "seeded by a boot" from "already copied": check for rows in the *data* tables, or record a copy marker, and never emit a message whose first branch is a data-loss instruction.
2. Copy first, then run the module's migrations over the copied rows.
[...]

PR #1398, opened 2026-09-10
fix(oidc): refuse to boot while the self-service confinement fold is pending

### LINKED ISSUE #1317: security(oidc): migration 0057 disables the self-service confinement until operator script 0011 runs, and nothing enforces it
## Summary

Migration `0057_oauth_provider_1_7_3.sql` moves the self-service discriminator from `metadata.selfService` to a new column with `DEFAULT false`, which turns an **active security control off** for every already-registered DCR/CIMD client. Turning it back on is a manual out-of-band script (`scripts/migration/0011-oauth-clients-self-service-fold.sql`), and **nothing enforces that it ran**.

## Mechanism

`packages/db/drizzle/0057_oauth_provider_1_7_3.sql:210`:

```sql
ALTER TABLE "oauth_clients" ADD COLUMN IF NOT EXISTS "self_service" boolean DEFAULT false NOT NULL;
```

`apps/api/src/modules/oidc/auth/guards.ts:620-627` now reads that column:

```ts
return row?.selfService === true;
```

The pre-bump guard read `JSON.parse(row.metadata).selfService` (`649277242^:guards.ts:702-703`) and was working. `metadata` survives the migration, so the fold was left out of it deliberately (0057:203-209 says so) — but the reasoning was made on "is this a fold or a data repair", not on "does this control fail open in the interval".

## Impact during the window

`isSelfServiceClient()` returning `false` makes the `/oauth2/token` before-hook skip the confinement branch (`guards.ts:743-744`). Such a client can then pass `resource=${APP_URL}` — the broad platform audience — and because `APP_URL` is not a protected-resource URI, `apps/api/src/lib/protected-resources.ts:184-191` computes `boundToResource === false` and lets it through on every REST route with the connecting user's full authority, instead of being caged to one org's MCP endpoint.

Registration is unauthenticated (`plugins.ts:262-263`), so anyone could have planted such a client before the upgrade. The ceiling that still applies is `getSelfServiceScopes()` (`apps/api/src/modules/oidc/auth/scopes.ts:93-95`) — identity scopes plus module end-user-grantable scopes — not the resource confinement, which is off.

## Proposed fix

Add a boot-time assertion that fails the health gate:

```sql
SELECT count(*) FROM oauth_clients
WHERE self_service = false AND metadata->>'selfService' = 'true';
```

Non-zero ⇒ 0011 has not run. Migrations otherwise apply automatically at boot, so the operator currently gets no prompt at all: the window is "deploy → whenever someone remembers", and it is silent on both ends.

## Verified alongside

- The fold key `selfService` matches exactly what the old code wrote, so 0011 is **not** a silent no-op.
- New clients are stamped `true` synchronously at creation (`oauth-admin.ts:426-432`), so only legacy rows depend on the script.
- 0011 is idempotent, can never flip `true` → `false`, and aborts cleanly inside `BEGIN` if run before 0057.

pierrecabriere:
Closes #1317

## Root cause

`packages/db/drizzle/0057_oauth_provider_1_7_3.sql:210` adds
`oauth_clients.self_service boolean DEFAULT false NOT NULL`, and
`apps/api/src/modules/oidc/auth/guards.ts:626` now reads that column instead of
the `metadata` JSON key it replaced. Every already-registered DCR/CIMD client
therefore lands `false` — i.e. reads as operator-provisioned — so the
`/oauth2/token` before-hook skips the single-protected-resource confinement
(`guards.ts:743-744`) and such a client may bind its token to `APP_URL`, the
broad platform audience, acting across the whole REST API with the connecting
user's full authority. Registration is unauthenticated (`plugins.ts:262-263`).

Restoring the flag is `scripts/migration/0011-oauth-clients-self-service-fold.sql`
and **nothing enforced that it ran**. Migrations apply automatically at boot and
the script does not, so the window was "deploy → whenever someone remembers",
silent on both ends, on a control that fails **open**.

## Fix

`assertSelfServiceFoldApplied()` in `apps/api/src/lib/boot.ts`, called from
`bootCritical` immediately after `assertOAuthResourceColumnsPresent()` — i.e.
after migrations have applied. It counts the rows 0011 would fold and, if any
remain, throws naming the script. Detection only; it never repairs
(`NO_TRANSITIONAL_CODE` §5).

Design notes:

- **Refuse, not warn** — same trade, same reasoning, as the
  `assertOAuthResourceColumnsPresent` refusal it sits beside. A log line on a
  control that is off is what the issue is about.
- **The fold stays out of 0057.** §2 is explicit: "a fold whose source
  **survives** is ordinary data repair and is not exempt" — `metadata` survives
  the file, so 0057's author was right and `verify:no-migration-dml` stays
  green. Editing an already-merged migration would also be silently skipped by
  any DB that has applied it (the migrator keys on a watermark, not a hash).
- **The predicate is exactly the script's `WHERE`**, `pg_input_is_valid`
  included. A row whose `metadata` is not JSON is one the fold deliberately
  skips and leaves `false`; counting it here would refuse a boot that running
  the script could never clear. Every condition this throws on is one 0011
  removes.
- **Invisible on a fresh install** — nothing writes that JSON key any more
  (`markClientSelfService` stamps the column and only the column,
  `oauth-admin.ts:426-432`), so the state exists solely on pre-bump rows.
- **§4 is respected explicitly**: this is a retirement guard, it records its
  removal condition (delete it, its tests and the README note in the PR that
  records `0011` as run in `scripts/migration/README.md`'s log), and the doc
  comment says so.

`scripts/migration/README.md` step 3 and the 0011 log row now state that the API
refuses to boot in between and that this is the intended sequence; the script's
own header cross-references the gate.

## Tests

- `apps/api/test/unit/oauth-self-service-fold-boot-gate.test.ts` (new, 6 tests) —
  injected count: passes at 0, refuses above it, names the script, reports the
  count, propagates a failing probe, and asserts the named script still carries
  the predicate the gate refuses on (a drift between the two would make a
  successful run fail to clear the refusal).
- `apps/api/test/integration/db/oauth-client-self-service-fold-migration.test.ts`
  (+1 test) — the **live** probe against the seed that is already exactly the
  state it detects: refuses before the script, passes after, with the
  unparseable-metadata row still `false` and still not keeping boot down.

```
bunx tsc --noEmit -p apps/api                                        → clean
TEST_TIER=0 bun test apps/api/test/unit/oauth-self-service-fold-boot-gate.test.ts
                                                                     → 6 pass / 0 fail
TEST_TIER=0 bun test apps/api/test/integration/db/oauth-client-self-service-fold-migration.test.ts
                                                                     → 3 pass / 0 fail
TEST_TIER=0 bun test apps/api/test/unit/oauth-resources-drift.test.ts \
                     apps/api/test/integration/db/oauth-resources-probe.test.ts
                                                                     → 7 pass / 0 fail
bun run verify:no-migration-dml                                      → green (68 files, 2 trees)
bunx eslint + prettier on the touched files                          → clean
```

Both backends the live probe runs on have `pg_input_is_valid` (PostgreSQL 16.8,
and the tier-0 PGlite build — verified by the integration run above, which is
PGlite).

## Notes for the orchestrator

- **No migration, no env var.** Platform migration indices untouched.
- **Deploy ordering — this is the operational point of the PR.** On the release
  that carries `0057`, a deployment that has ever accepted a self-registered
  client will now come up only far enough to apply the migrations and then exit,
  crash-looping under Coolify until
  `scripts/migration/0011-oauth-clients-self-service-fold.sql` is run against the
  database. That is intended and documented in `scripts/migration/README.md`
  step 3. Production is at `0055` (`git ls-tree v1.0.0-beta.57 packages/db/drizzle/`),
  so `0057` has not run there yet and the sequence is available cleanly.
- **Assumption stated**: I did not edit `0057` itself. Beyond the §2 rule above,
  a dev/staging DB that has already applied it would never see the change —
  drizzle's postgres-js migrator applies by `max(created_at)` watermark, not by
  hash — which is the same silent failure class the issue reports. Any such DB
  simply hits the new boot refusal and is cleared by running 0011, like
  production.
- **Overlap with #1350**: both concern `0057` / `scripts/migration/README.md`,
  but disjoint hunks — #1350 is the pre-flight query at README lines 101-104 and
  the `public`/`application_type` fold comment at `0057:179-192`; this PR touches
  README step 3 (line ~117) and the log row, and does not modify `0057` at all.
  No conflict expected for the agent basing on this branch.
- No overlap with #1351, #1319, #1318.


pierrecabriere:

Issue #1358, opened 2026-09-10
models: duplicate org_models rows are not refused, and per-model spend reporting splits across them


pierrecabriere:
## Summary

`POST /api/models` accepts duplicate `(orgId, credentialId, modelId)` rows. Because `llm_usage.model` stores the **row id**, per-model spend reporting then splits across the duplicates and under-reports each one.

## Mechanism

`packages/db/src/schema/organizations.ts:449-499` has no unique constraint on `(orgId, credentialId, modelId)` or on `label`. `createOrgModel` (`apps/api/src/services/org-models.ts:399-436`) performs no existence check.

Only the **label** is deduped (`deriveModelLabel:380` → `dedupeLabel`), and that read is in a separate transaction from the insert — a TOCTOU under the multi-add path's N sequential POSTs.

`seedOrgModelsForCredential` **does** guard (`org-models.ts:527-533`); `POST /models` does not.

## Impact

- The picker shows "GPT-5" and "GPT-5 (2)".
- `llm_usage.model` stores the row id (`metering.ts:328`), so spend reporting splits across duplicates and under-reports each.
- Deleting the credential (FK `ON DELETE RESTRICT`) requires finding every copy.

#1276's multi-add makes this easier to hit.

## Proposed fix

Add the unique constraint, or an existence check inside the same transaction as the insert, and decide what the API returns on a duplicate (409, or idempotent 200 on the existing row).

Issue #1357, opened 2026-09-10
codex: gpt-6-astra is Pro-plan-only and is the first featured model for every user


pierrecabriere:
## Summary

`gpt-6-astra` is documented as recommended for **Pro plans and above**, and is now the **first** featured model for every Codex user, with no plan awareness. A Plus subscriber is offered it in the top slot and finds out at the first run.

## Mechanism

`packages/module-codex/src/index.ts:229`:

```ts
featuredModels: ["gpt-6-astra", "gpt-5.6-sol", …]
```

Because `modelDiscovery: { mode: "static" }`, the served set is `resolveCatalogBackedCandidates` = `(definition ∩ catalog)` (`model-selection.ts:245-253`) with no plan awareness — and compliance forbids enumerating a subscription, so it cannot acquire any.

CHANGELOG lines 819-823 record the Pro-plan constraint.

## Why this is inconsistent

The same commit deliberately kept `gpt-5.3-codex-spark` — also Pro-only — **out** of `featuredModels`, with a comment recording that judgement. The only thing that differs for `gpt-6-astra` is that it happens to be present in `openai.json`.

## Proposed fix

Either apply the same judgement (drop it from `featuredModels`, keep it selectable), or surface the plan requirement in the picker so the failure is legible before the run rather than after it.

Issue #1354, opened 2026-09-10
models: a discovered model freezes catalog-derived — sometimes foreign-vendor — capabilities as permanent overrides


pierrecabriere:
## Summary

A model added through **discovery** has catalog-derived capabilities written as **explicit** `org_models` overrides — including capabilities looked up from a *different vendor's* catalog. Unlike a catalog-origin row, those values are then pinned forever and the weekly refresh cannot correct them.

## Mechanism

`describeServedModel` (`apps/api/src/services/model-providers/model-metadata.ts:37-58`) fills missing fields from the provider's own catalog, then from **any** catalog by exact id, then from any catalog after stripping one `<vendor>/` prefix — and reports `source: "catalog"`.

`discoveredRows` (`apps/web/src/lib/model-source.ts:92-105`) carries that `source` onto the row, but `rowToEntry` (`apps/web/src/lib/model-form-payload.ts:227-245`) branches only on `origin`: for `origin === "discover"` it ships `contextWindow`, `maxTokens`, `input` and `reasoning` unconditionally.

## Failure scenario

An operator points a custom endpoint at a self-hosted vLLM serving `deepseek-ai/DeepSeek-V3` with a 32k context. Discovery strips the vendor prefix, finds `DeepSeek-V3` in the together-ai catalog, and the row is saved with `contextWindow: 128000` as an explicit override. Every run then sends prompts the endpoint will reject or truncate, and the row is also labelled with the foreign vendor's label. The weekly catalog refresh cannot fix it.

## Proposed fix

Branch on `source` as well as `origin`: a value that came from a catalog should be stored as `null` and re-resolved, exactly as a catalog-origin row does. A cross-vendor prefix-stripped match probably should not populate capabilities at all — at minimum it should not be presented as authoritative.

PR #1399, opened 2026-09-10
fix(models): bound and audit the /discover probe

### LINKED ISSUE #1356: models: POST /discover buffers an unbounded body from an operator-supplied URL, and is unaudited

pierrecabriere:
Closes #1356

## Root cause

Three distinct defects on `POST /api/model-provider-credentials/discover`, which probes an **operator-supplied** URL with a user-supplied key:

1. **Unbounded body.** `fetchListingPage` (`apps/api/src/services/model-providers/model-listing.ts`) called `res.json()` on the response. `fetchModelListing` (`apps/api/src/services/org-models.ts:1119`) sets `AbortSignal.timeout(10_000)` and `egressGuardedFetch` bounds *where* the request goes — nothing bounded *how many bytes* came back. `MAX_SERVED_MODELS` caps the parsed array, i.e. after the whole payload is materialised.
2. **No audit trail.** The route recorded nothing, unlike the create / update / delete routes beside it (`model-provider-credentials.ts:318`, `:504`, `:535`).
3. **Wrong gate (latent).** `assertApiKeyProvider` (`:114`) refused on `authMode !== "api_key"`, but the declaration that keeps a credential off the listing path is `modelDiscovery: { mode: "static" }` — the predicate `discoverAvailableModels` (`model-discovery.ts:83`) branches on and the registry enforces at boot for oauth2 (`registry.ts:75`).

## Fix

- **Body cap** — `readBoundedJson` reads each page under a 4 MB budget, cancelling the stream the moment it crosses and returning `BAD_RESPONSE`. Budget spent on the bytes read, **not** on a declared `content-length` (an untrusted endpoint controls that header). Placed in the shared page fetch rather than the route, so `POST /{id}/refresh-models` gets the same protection. 4 MB is roughly an order of magnitude above a 1000-model page carrying full metadata.
- **Audit** — `model_provider_credential.discovered` with provider, endpoint reached, outcome, model count, truncation. Never the key. `resource_id` is null on the inline form (no credential exists yet); the `credential_id` form names the credential. A request refused before the probe (unknown provider, static list, both forms at once) is not audited — no key was spent.
- **Gate** — `assertApiKeyProvider` → `assertEnumerableProvider`, refusing on `modelDiscovery?.mode === "static"`. Behaviour for every shipped provider is unchanged (both static providers are oauth2); an api-key provider declaring a static list is now refused instead of enumerated.

No new helper was invented where one existed: the repo has bounded readers in `runtime-pi/sidecar`, `packages/connect` and `runtime-pi/mcp`, none reachable from `apps/api` and none of them a shared export — promoting one into `@appstrate/afps-shared` would drag an npm publish lockstep into a security fix, so the reader is local to the file that owns the page fetch.

## Tests

`apps/api/test/integration/routes/model-provider-credentials-discover.test.ts` — three cases, each verified to **fail without its fix** (negative control run: 3 fail / 17 pass with the fixes reverted):

- *refuses a listing whose body streams past the size budget* — stub streams ~6 MB of **well-formed** JSON, chunked so no `content-length` declares it; parsing would succeed, only the budget refuses it.
- *refuses an api-key provider that declares a static model list* — synthetic provider, asserts no listing request was spent.
- *records the probe in the audit trail, without the key*.

```
TEST_TIER=0 bun test apps/api/test/integration/routes/model-provider-credentials-discover.test.ts   → 20 pass / 0 fail
TEST_TIER=0 bun test .../model-listing.test.ts .../model-discovery.test.ts .../model-provider-credentials.test.ts → 63 pass / 0 fail
bunx tsc --noEmit -p apps/api → clean;  -p apps/web → clean
bun run verify:openapi → ALL CHECKS PASSED
bun run verify:api-types → up to date (regenerated: description-only change)
bun run detect:breaking → PASSED — 0 breaking, 1 non-breaking (the base's `truncated` field)
```

## Notes for the orchestrator

- Based on `fix/issue-1353`; the body cap sits in that branch's `fetchListingPage`, so it protects every page of a paginated listing, not just the first.
- `apps/web/src/api/schema.d.ts` regenerated (one line — the `/discover` description).
- **Overlap:** none with #1354 / #1355 / #1358 / #1357. The audit row is a new write on a route the siblings' issues do not touch.
- The 4 MB per-page budget is a judgement call, not a product decision — a single constant (`MAX_LISTING_BODY_BYTES`) if it ever needs tuning.


pierrecabriere:

Issue #1336, opened 2026-09-10
security(schedules): GET /api/schedules/:id/runs is guarded on schedules:read alone, bypassing the API-key run ceiling


pierrecabriere:
## Summary

`GET /api/schedules/:id/runs` is guarded on `schedules:read` alone. Every other run-list surface moved to `requireRunsRead` with the `runs:read` / `runs:read-all` split (#1307); this one did not, so a credential with no run scope at all reads the full enriched run projection.

## Mechanism

`apps/api/src/routes/schedules.ts:573`:

```ts
router.get("/schedules/:id/runs", requirePermission("schedules", "read"), async (c) => {
```

Rows are still narrowed by `runVisibilityFilter`, so this is a **credential-ceiling bypass**, not a cross-principal leak.

## Executed

API key with `scopes: ["schedules:read"]` — a legal, grantable scope set:

```
GET /api/runs                    → 403
GET /api/schedules/{id}/runs     → 200
```

…returning the full `enrichedRunColumns` projection: `input`, `result`, `checkpoint`, `error`, `metadata`, `context_snapshot`, `resolved_connections`, `cost`, `token_usage`.

## Proposed fix

Add `requireRunsRead` alongside `schedules:read` on this route, matching `lib/run-visibility.ts:58`'s stated contract that it is "the guard on every surface `runs:read` opens".

Issue #1335, opened 2026-09-10
security(runs): PATCH /api/runs/:runId/sink/extend applies no run visibility


pierrecabriere:
## Summary

`PATCH /api/runs/:runId/sink/extend` applies no run visibility. A member who cannot even *see* another principal's run can mutate it — pushing out its sink expiry and bumping its heartbeat, which defeats the stall watchdog — and the 200-vs-404 difference is a liveness oracle over the whole space's runs.

Introduced alongside the `runs:read` / `runs:read-all` split (#1307), not yet deployed.

## Mechanism

`apps/api/src/routes/runs-remote.ts:397-448`. The guards are `requirePermission("agents", "run")` plus an org/space `WHERE`. The handler's comment claims *"the handler checks ownership explicitly"* — it checks **tenancy**.

## Executed

User B holding the **`runner`** preset, against user A's run:

```
GET   /api/runs/<A's run>                                   → 404
PATCH /api/runs/<A's run>/sink/extend {"ttl_seconds":3600}   → 200 {"id":…,"expiresAt":…}
```

The write pushes out `sink_expires_at` and bumps `last_heartbeat_at`, so `services/run-watchdog.ts` stops finalising a stalled run.

No test covers this route's isolation.

## Proposed fix

Run the same predicate every other run surface uses — `assertRunVisible` / `runVisibilityFilter` from `apps/api/src/lib/run-visibility.ts`. That module exists precisely so ownership is not re-derived per route; this is one of the two places that still does.

PR #1400, opened 2026-09-10
fix(files): cap a creator's file lifecycle right at the credential ceiling

### LINKED ISSUE #1315: security(files): the creator branch of delete/keep bypasses the API-key credential ceiling

pierrecabriere:
Closes #1315

## Root cause

`getFileCapabilities` (`apps/api/src/services/files.ts:358`, pre-fix `:351`) computed
`manage = isCreator || canManage`, and the two lifecycle routes
(`apps/api/src/routes/files.ts` `DELETE /files/:id`, `POST /files/:id/keep`) deliberately
carry no `files:*` guard at layer 1 — an end-user cleaning up its own upload holds no org
permission. Under API-key auth the actor is the key's **creator** (RBAC spec §7.1), so
`isCreator` was true for every file that user ever produced: a key minted with
`scopes: ["agents:run"]` could `DELETE` any of them and `POST …/keep` to pin them against
the retention GC. `permissions` (role ∩ scopes) could not stop it, because the creator arm
is an ownership right and never passes through that intersection.

## Fix

The credential's own **scope ceiling** now caps the creator arm.

- `credentialAdmits(c, permission)` (`apps/api/src/lib/permissions.ts`) — the ceiling
  question, sibling to `callerPermissions`'s grant question. `c.get("scopeCeiling")` is
  `undefined` for a cookie session (unbounded, ownership right intact) and is the scope list
  for an API key / OIDC token (kept only when it names `files:delete`).
- `getFileCapabilities(doc, actor, { visible, canManage, creatorCanManage })` —
  `manage = canManage || (isCreator && creatorCanManage)`. Both lifecycle inputs default to
  **false**, so a caller that does not state its authority gets none.
- `getFileForActor` / `listFilesForActor` take `opts.creatorCanManage` (optional,
  fail-closed). The file router passes `fileLifecycleCeiling(c)` on **every** resolution, so
  the DTO's `capabilities` and the two enforcement points can never disagree; the read-only
  surfaces (MCP `resources/read`, the `appfile://` run-input resolver, the chat attachment
  resolver) pass `false` explicitly — they expose no lifecycle affordance.
- Reused the sibling's `isFileCreator` helper (it is what `isCreator` already reads through).

Nothing widens: the grant arm (`files:delete`) is untouched, and `download` / `metadata` /
`preview` are untouched — a scope-limited key still reads its creator's files with
`files:read`. The two 403 messages now name the scope instead of saying "creator or admin".

**Deliberate narrowings** (both are the fix, not collateral):
- An API key whose scopes omit `files:delete` no longer manages its creator's files. A key
  held by an `operator`/`runner`/`viewer` can never carry that scope (`validateScopes`
  filters to the creator's own authority), so those roles lose *headless* file cleanup;
  their cookie sessions are unaffected.
- `files:delete` is not in `OIDC_ALLOWED_SCOPES`, so an end-user JWT never carries the
  creator arm. End-user cleanup runs through an API key that holds the scope — the
  integration test covers exactly that, plus the negative arm.

Docs: `docs/architecture/FILES.md` capability matrix gains the ceiling footnote; the two
OpenAPI descriptions state the rule; `apps/web/src/api/schema.d.ts` regenerated.

## Tests

- `apps/api/test/unit/services/files.test.ts` — new `it("the creator arm is capped by the
  credential ceiling, the grant arm is not")`; updated the existing creator case to state
  its unbounded credential.
- `apps/api/test/integration/services/files.test.ts` — new `it("an API key does not inherit
  its creator's ownership right over their files")`: `files:read`-only key → `GET` 200 with
  `capabilities.delete/keep` **false**, `DELETE` 403, `keep` 403, `expires_at` intact;
  `files:read + files:delete` key → 200 / 204. The end-user test gains a negative arm (a
  `files:read`-only key impersonating the file's end-user creator → 403).

- `set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/integration/services/files.test.ts apps/api/test/unit/services/files.test.ts` → **76 pass, 0 fail**
- `… bun test apps/api/test/integration/services/files-hardening-e2e.test.ts apps/api/test/integration/modules/mcp-files.test.ts apps/api/test/integration/routes/files-preview.test.ts` → **56 pass, 0 fail**
- `… bun test apps/api/test/integration/services/files-storage-matrix.test.ts apps/api/test/integration/routes/runs-read-isolation.test.ts` → **21 pass, 6 skip, 0 fail**
- `bunx tsc --noEmit -p apps/api` → clean; `bunx tsc --noEmit -p apps/web` → clean
- `bun run verify:openapi` → ALL CHECKS PASSED; `bun run generate:api` (committed) → `bun run verify:api-types` green
- `bunx eslint` + `bunx prettier --write` on every touched file

## Notes for the orchestrator

- **Base**: `fix/issue-1314`. Both branches edit `apps/api/src/services/files.ts`,
  `apps/api/test/integration/services/files.test.ts`, `docs/architecture/FILES.md`,
  `apps/web/src/api/schema.d.ts` and `apps/api/src/openapi/paths/files.ts` — stacked, not
  conflicting.
- **No CHANGELOG entry** — matching `fix/issue-1314`, which adds none either; a single
  consolidated entry on the merged branch is cleaner than four conflicting ones. This one
  is worth an `### Security` bullet: it is live in production.
- Overlap with siblings #1335 / #1336: none (neither touches files).
- `apps/web/src/api/schema.d.ts` moved by 2 lines (OpenAPI descriptions only). If a sibling
  also regenerates it, re-run `bun run generate:api` after the merge rather than resolving
  by hand.

### DIFF
--- a/apps/api/src/lib/permissions.ts
+++ b/apps/api/src/lib/permissions.ts
@@ -434,6 +434,26 @@ export function callerPermissions(c: Context<AppEnv>): ReadonlySet<string> {
 }
 
 /**
+ * Does the CREDENTIAL this request arrived on admit `permission` at all?
+ *
+ * The ceiling question, not the grant question {@link callerPermissions}
+ * answers. An API key's scope list and a token's scope claim cap whatever
+ * their creator holds ({@link effectivePermissions}); a cookie session carries
+ * no ceiling and admits everything its role does.
+ *
+ * It exists for the one authorization in the platform that is NOT a role grant
+ * and therefore never passes through that intersection: a file's own creator
+ * manages their file (`keep` / `delete`) without holding `files:delete`. That
+ * ownership right must still not outreach the credential the request arrived
+ * on — a key minted with `agents:run` does not inherit its creator's files
+ * (RBAC spec §7.1).
+ */
+export function credentialAdmits(c: Context<AppEnv>, permission: Permission): boolean {
+  const ceiling = c.get("scopeCeiling");
+  return ceiling === undefined || ceiling.has(permission);
+}
+
+/**
  * Org-level grants of `role` as an org LISTING shows them, sorted. No space
  * half: the space slice is answered per space by `GET /api/spaces`.
  */
--- a/apps/api/src/modules/mcp/package-file-tools.ts
+++ b/apps/api/src/modules/mcp/package-file-tools.ts
@@ -100,7 +100,11 @@ async function readPackageFileBytes(
 ): Promise<PackageFileBytes> {
   const fileId = parseFileUri(uri);
   if (!fileId) throw new McpError(ErrorCode.InvalidParams, `Not a file URI: ${uri}`);
-  const resolved = await getFileForActor(ctx.scope, ctx.actor, fileId, ctx.permissions);
+  // Read-only surface: MCP exposes no `keep` / `delete` affordance, so a
+  // creator's ownership right never applies through it.
+  const resolved = await getFileForActor(ctx.scope, ctx.actor, fileId, ctx.permissions, {
+    creatorCanManage: false,
+  });
   if (!resolved) throw new McpError(ErrorCode.InvalidParams, `File not found: ${uri}`);
   if (!resolved.capabilities.download) {
     throw new McpError(ErrorCode.InvalidParams, `File is not downloadable: ${uri}`);
--- a/apps/api/src/modules/mcp/tools.ts
+++ b/apps/api/src/modules/mcp/tools.ts
@@ -1317,7 +1317,12 @@ export function buildFileResourceProvider(ctx: McpToolContext): AppstrateResourc
         throw new McpError(ErrorCode.InvalidParams, `Not a file resource URI: ${uri}`);
       }
 
-      const resolved = await getFileForActor(ctx.scope, ctx.actor, docId, ctx.permissions);
+      // Read-only surface: MCP exposes no `keep` / `delete` affordance, so a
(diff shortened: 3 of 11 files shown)


pierrecabriere:

Issue #1327, opened 2026-09-10
pricing: the weekly catalog refresh silently re-prices existing customers and zero-rates removed models


pierrecabriere:
## Summary

Catalog-backed org models store `cost = NULL` deliberately and resolve the price live on every call, all the way down to the EE credit debit. So the weekly automated pricing refresh silently changes **what existing customers pay** — and silently drops others to zero — with no audit event, no notification, and no per-org pin.

`ba8a38a99` (#1277) is an unreviewed bot commit that contains exactly such a change.

## The chain

`resolveCatalogDefaults` returns `entry.cost` (`apps/api/src/services/org-models.ts:731`) → `computeCostUsd(usage, inputs.resolved.cost ?? null)` (`apps/api/src/services/llm-proxy/metering.ts:296`) → `llm_usage.cost_usd` → module-ee credit debit.

Rows are created with `cost = NULL` at `org-models.ts:536-540`, on purpose, so they track the catalog.

## What `ba8a38a99` actually changed

Diffed `apps/api/src/data/pricing/*.json` across `ba8a38a99^..ba8a38a99`:

```
mistral   magistral-medium-latest   output 5   →  7.5          (+50 %)
mistral   magistral-small-latest    in 0.5→0.15, out 1.5→0.6
mistral   ministral-14b-{2512,latest}, ministral-3b-{2512,latest}, … (cacheRead added)

google-ai gemini-2.5-flash-native-audio-latest             REMOVED
google-ai gemini-2.5-flash-native-audio-preview-09-2025    REMOVED
google-ai gemini-2.5-flash-native-audio-preview-12-2025    REMOVED
google-ai gemini-3.1-flash-live-preview                    REMOVED
```

**Re-pricing**: a customer on `magistral-medium-latest` has their output rate raised 50 % the moment the next release deploys.

**Zeroing**: `resolveCatalogDefaults` returns `{}` for an unknown id → `cost` undefined → `computeCostUsd(usage, null)` → `cost_usd: 0`. Any org model on one of those four ids now bills **zero credits**. There is a warn (`services/pricing-provenance.ts:89`) but it is deduped per `(orgId, model, status)` behind a cap and nothing gates on it.

`pricingStatus` stays `priced` in both directions, so nothing logs the transition either way.

## Proposed fix

- Make a price delta on a **model an org actually uses** a blocking review item, not a line in a bot PR's diff. The PR template already asks reviewers to spot-check prices; that check has no teeth and no list of affected orgs.
[...]

PR #1401, opened 2026-09-10
fix(module-ee): refuse to resume a billing watermark the sweeper abandoned

### LINKED ISSUE #1370: ee(billing): re-enabling the module after a disabled window back-bills the whole gap and leaves orgs inconsistent

pierrecabriere:
Closes #1370

## Root cause

`packages/module-ee/src/index.ts` `init()` seeded or read `ee_billing_cursor` with no age check (`ensureCursorSeeded`, `src/billing/usage-recorder.ts:139`). Removing the module from `MODULES` stops the sweep but not the platform's `llm_usage` appends, so the watermark outlives the window and the first tick after re-enabling claims every row of it and debits the lot against the organizations' **current** quotas — fleet-wide soft-cap overshoots and quota-warning emails, and irreversible: `ee_billed_llm_usage` is never purged, so those rows cannot be un-billed by re-reading them.

A second, latent problem the guard exposed: `ee_billing_cursor.updated_at` only moved when a pass had rows to advance over (`usage-recorder.ts`, the `rows.length === 0` early return), so on a deployment quiet enough to read an empty ledger region it meant "the last row was billed", not "the sweep last looked".

## Fix

- **`updated_at` now means "the sweep last confirmed this watermark"** — stamped on the caught-up path too, and surfaced on `CursorSeedResult`. Without this a quiet deployment ages into a refusal with no gap behind it.
- **`assertCursorResumable`** (`src/billing/billing-sweeper.ts`), called from `init()` between the seed and `startBillingSweeper()`. Throws — a fatal boot — and the message *is* the runbook: forgive the gap with `DELETE FROM ee_billing_cursor;` (the next boot re-seeds watermark **and** `floor_id` at the settled frontier, exactly as the original cutover, so the gap is never read again), or bill it by setting `EE_RECONCILIATION_MAX_GAP_SECONDS` above the printed age (`0` resumes over anything). It also points at `repair:account` for the orgs the window stranded.
- **The predicate is two-part on purpose**: the sweep was absent longer than `EE_RECONCILIATION_MAX_GAP_SECONDS` (new, default 86400) **and** the backlog exceeds one tick's drain capacity (`MAX_DRAIN_ITERATIONS × EE_RECONCILIATION_BATCH_SIZE`). Age alone would refuse a platform that was merely shut down for a week (no sweep ran, but no usage accrued either); backlog alone would refuse a sweep that is honestly behind while ticking, which needs capacity, not an operator. A head-of-line stall trips neither — it pins `settledFrontier()` where it pins the watermark, so the backlog stays ~0. `EE_RECONCILIATION_INTERVAL_SECONDS=0` (metering paused, per #1326) skips the check entirely; the gap is judged at the boot that un-pauses it.
- **README** now documents all three effects of a disabled window, including the two that are *not* guarded (see the open decision below).

Rejected the issue's literal "compare the watermark's age against a threshold": age alone bricks the boot of any platform that was simply switched off, and `updated_at` did not carry the meaning the check needs.

## Tests

`packages/module-ee/test/integration/services/billing-sweeper.test.ts` — new `describe("billing cursor — resuming after the sweeper was absent")`, 8 cases: the refusal, the runbook content of its message, and the four shapes that must still boot (shut-down platform, behind-but-ticking sweep, paused metering, `MAX_GAP_SECONDS=0`, freshly seeded cutover), plus the caught-up stamp. Four pre-existing `toEqual` assertions on `CursorSeedResult` became `toMatchObject` (the shape gained `updatedAt`).

```
pg-test-lock.sh … bun test packages/module-ee/test/integration/services/billing-sweeper.test.ts   → 72 pass, 0 fail
pg-test-lock.sh … bun test packages/module-ee                                                     → 444 pass, 26 skip, 0 fail
bunx tsc --noEmit -p packages/module-ee                                                           → clean
bunx eslint <touched files>                                                                       → clean
bun run verify:env-docs / verify:compose-defaults / verify:module-isolation /
    verify:module-sql-boundary / verify:license-boundary / verify:no-migration-dml                 → all green
```

## Notes for the orchestrator

- **New env var** `EE_RECONCILIATION_MAX_GAP_SECONDS` (optional, default `86400`): `packages/module-ee/src/env.ts`, `.env.example`, `docs/ENV.md`, plus the `docker-compose.yml` module pass-through block (`verify:compose-defaults` fails without that last one). No migration, no platform env change.
- **Deploy ordering**: this can refuse a boot. A production deployment whose sweep has genuinely been off for over a day and is more than 5000 rows behind will not start until the operator picks a path. Worth checking `SELECT updated_at, last_llm_usage_id FROM ee_billing_cursor;` against `settledFrontier` before rolling it out; prod has been running EE continuously, so the expected answer is "fresh".
- **Overlap**: none with #1329 (`repair:account` transactionality) or #1327 (catalog re-pricing). Stacked on #1330; touches `usage-recorder.ts`, `env.ts` and `billing-sweeper.ts`, all already merged from #1326/#1328/#1330 in the base.

## Open decision (deliberately not shipped)

The issue also asks to **reconcile missing `ee_billing_accounts` rows on the same pass**. That is not implementable inside the module's platform contract: EE reads the platform only through `ctx.services` and `getOrgOwnerEmails` / `getOrgMembers` (`src/platform-org-queries.ts`), and **nothing in `ModuleInitContext` enumerates organizations** (`packages/core/src/module.ts:1233`) — a fleet-wide backfill would need a new core query. It is also a commercial call: auto-provisioning free-tier accounts for orgs an operator never saw is a policy, not a repair. Today those orgs are refused at `beforeUsage` (`no_account` → `402`) and named by the sweep at `error` level, and `bun run repair:account` fixes each one; that is now written down in the README instead of being silently absent. If the fleet-wide backfill is wanted, it needs (a) a core `listOrganizations`-shaped query and (b) a decision on free-tier minting.

The third effect — usage of orgs **deleted** during the window cascaded away unbilled — is unrecoverable after the fact by construction. Documented, not fixable.


pierrecabriere:

Issue #1263, opened 2026-09-04
Hosted Connect portal returns 502 when the space has no OAuth client registered


otarbes:
## Problem

`GET /api/integrations/connect/start` returns **502** with "Could not start the connection. Please try again." when the space has no OAuth client registered for the integration's auth. Retrying never helps: the condition is a permanent configuration gap, not a transient fault.

The same condition is reported correctly on the programmatic surface:

```
POST /api/integrations/{packageId}/auths/{authKey}/connect/oauth2
403 forbidden
"Administrator must register OAuth client credentials for '@appstrate/gmail-mcp' auth 'primary' before connection"
```

That message names the action to take. The hosted portal hides it behind a generic gateway error.

## Why it matters

This is the default path for a user who creates their first space and connects an integration there. OAuth clients are stored per space (`spaceId` is on the row returned by `createIntegrationOAuthClient`), so a brand-new space has none until one is registered. The first thing that user sees is a Cloudflare-style 502.

## Secondary effect: the link is burned

`consumeJti` runs before the `strategy.begin` call, so the single-use token is spent by the failed attempt. A second click on the same link returns "This connect link has already been used." — a second misleading message for the same root cause.

## Root cause

`apps/api/src/routes/integrations.ts`, `/connect/start`:

```ts
try {
  result = await strategy.begin(/* … */);
} catch (err) {
  logger.error("Hosted connect OAuth begin failed", { /* … */ });
  return c.html(popupHtmlError("Could not start the connection. Please try again.", {}), 502);
}
```

The catch is documented as covering transient or structural faults (OAuth client removed between mint and click, provider discovery error, network). A missing-client configuration error falls into the same net and inherits a "please try again" wording that does not apply.

## Reproduction

1. Create a new space.
2. Activate an OAuth-based integration in it, without registering an OAuth client for that space.
3. `POST /api/integrations/{packageId}/auths/primary/connect/session`, open the returned `connect_url`.
4. 502. `POST .../connect/oauth2` on the same space returns the explicit 403.

## Suggested fix

Distinguish configuration errors from transient ones in that catch, and render the actionable message the API already produces ("register OAuth client credentials for this auth"). Ideally also skip burning the jti when the failure is known to be permanent, so a retry after registering the client can reuse the link.

PR #1404, opened 2026-09-10
fix(connect): keep the connect link burned once a refusal cost an upstream registration

### LINKED ISSUE #1344: security(connect): releaseJti runs after a real upstream DCR registration, making the unauthenticated start route a spam amplifier

pierrecabriere:
Closes #1344

## Root cause

`apps/api/src/routes/integrations.ts` — the `begin` catch on `GET /connect/start` released the single-use jti for the whole 4xx class, on the premise stated in its own comment: *"nothing was minted on the strength of this click"*.

That premise holds only for an auth whose OAuth client is pre-registered. For an **auto-provisioned** (DCR/CIMD) auth, `OAuth2Strategy.begin` → `ensureIntegrationOAuthClient` (`apps/api/src/services/integration-connections.ts:1625`) discovers the authorization server and POSTs an RFC 7591 **registration** to it — and the 403 that lands in this catch is raised precisely when that registration came back unusable (a confidential client where a public one was asked for), leaving an orphan client upstream: *"The upstream registration is abandoned unused"* (`:1651`). Releasing the jti let one 10-minute `connect_url` replay discovery + registration on **every** click of an unauthenticated, unlimited route.

## Fix

Two changes, both in the route.

1. **Release only when the refusal provably precedes any egress** — the same criterion #1352's new `releaseJti` site uses, so the two are consistent. The discriminator is the auth's kind, not the error's identity, because the auto-provisioned path is the *only* egress-before-refusal path in `begin`: on a classic auth `ensureIntegrationOAuthClient` early-returns a plain row lookup, and every `ApiError` 4xx `begin` can raise is thrown before `initiateIntegrationOAuth` — the first line that talks to anyone (everything past it throws `OAuthCallbackError`, which takes the 502 branch and keeps the burn). So `if (!usesAutoProvisionedClient(manifest, auth)) await releaseJti(claims.jti)` is exact, and it reuses the helper the route already imports rather than adding an error-code whitelist that could only approximate the same predicate.

   The end-user UX the release exists for is preserved where it was real: a classic "administrator must register a client" 403 still leaves the link reusable, so a retry after the admin acts needs no re-mint (existing test, unchanged).

2. **Rate-limit the route per IP** (`rateLimitByIp(60)`), as the issue asks. It carries no session, and every refusal that hands the jti back leaves the link replayable for its whole TTL — one click is a manifest load plus client lookups, so an unlimited public entry point is an amplifier on its own. 60/min is far above what a human clicking a popup needs. A 429 renders the platform's standard problem+json, not a popup page: this limit answers abuse, not a flow failure.

OpenAPI: the 400 and 403 descriptions asserted "The link stays reusable" unconditionally — amended to state the auto-provisioned exception; `429` added. `apps/web/src/api/schema.d.ts` regenerated.

## Tests

`apps/api/test/integration/routes/integrations-hosted-connect.test.ts` — one new case, the mirror of the existing "keeps the link reusable" one: a remote MCP (auto-DCR) refusal must answer 410 "already been used" on the second click.

```
  apps/api/test/integration/routes/integrations-hosted-connect.test.ts \
  apps/api/test/integration/routes/integrations-connect-scope-catalog.test.ts
→ 30 pass, 0 fail (156 expect() calls)

bun run verify:api-types        → up to date
bun run detect:breaking         → 0 breaking, 1 non-breaking (new 429)
```

Negative control: with the guard reverted to the unconditional `releaseJti`, the new test fails `Expected: 410, Received: 403`.

## Notes for the orchestrator

- Based on `fix/issue-1352`; no overlap with the earlier links in the chain beyond the same handler. The comment #1263 left on case 1 ("Hand the jti back either way…") is the claim this issue falsifies, so it was rewritten rather than left contradicting the code.
- Behaviour change worth naming: the undecryptable-`client_secret` 403 and the malformed-oauth2-auth 400 now also burn the link **on an auto-provisioned auth only** (classic auths are unchanged). Both are operator misconfigurations with no upstream cost on the classic path, which is why the classic path keeps its reusability.
- The rate limiter shares the process-global limiter cache in tests; the API suite issues ~20 `/connect/start` requests total, well under the 60/min bucket.

### DIFF
--- a/apps/api/src/openapi/paths/integrations.ts
+++ b/apps/api/src/openapi/paths/integrations.ts
@@ -1011,13 +1011,14 @@ export const integrationsPaths = {
         "302": { description: "Redirect to the provider OAuth screen or the hosted form." },
         "400": {
           description:
-            "Missing token, or the oauth2 auth declares neither an issuer nor explicit endpoints (HTML error page). The link stays reusable.",
+            "Missing token, or the oauth2 auth declares neither an issuer nor explicit endpoints (HTML error page). The link stays reusable — except on an auth that auto-provisions its client (DCR/CIMD), where every refusal burns it.",
         },
         "403": {
           description:
-            "The space has no OAuth client registered for this auth and none could be auto-provisioned; the page says the failure is permanent and to ask an administrator, while the operator-facing detail naming the exact remedy stays on the server log — this route carries no session (HTML error page). The link stays reusable so a retry after the administrator registers a client needs no re-mint.",
+            "The space has no OAuth client registered for this auth and none could be auto-provisioned; the page says the failure is permanent and to ask an administrator, while the operator-facing detail naming the exact remedy stays on the server log — this route carries no session (HTML error page). For an auth whose client is pre-registered the link stays reusable, so a retry after the administrator registers one needs no re-mint. For an auth that auto-provisions its client at the authorization server (DCR/CIMD) the link is burned: reaching this refusal means a registration was already attempted upstream, and a reusable link would replay it on every click.",
         },
         "410": { description: "Invalid, expired, or already-used token (HTML error page)." },
+        "429": { $ref: "#/components/responses/RateLimited" },
         "500": {
           description: "Integration cannot be connected / unexpected failure (HTML error page).",
         },
--- a/apps/api/src/routes/integrations.ts
+++ b/apps/api/src/routes/integrations.ts
@@ -68,6 +68,7 @@ import { setOffsetLinkHeader } from "../lib/pagination-link.ts";
 import { popupHtmlClose, popupHtmlError } from "../lib/oauth-popup-html.ts";
 import { normalizeOAuthErrorCode, oauthDiagnosticSuffix } from "../lib/oauth-error-diagnostic.ts";
 import { requirePermission } from "../middleware/require-permission.ts";
+import { rateLimitByIp } from "../middleware/rate-limit.ts";
 import { getActor, type Actor } from "../lib/actor.ts";
 import { getSpaceScope } from "../lib/scope.ts";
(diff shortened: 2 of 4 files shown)


pierrecabriere:

Issue #1372, opened 2026-09-10
oidc: runs:read-all is unrequestable by any OAuth client, silently narrowing admins over MCP


pierrecabriere:
## Summary

`runs:read-all` cannot be requested by any OAuth client, including one acting for an org admin, because it is absent from the provider's global scope vocabulary — so an admin driving the platform over MCP-over-OAuth is silently narrowed to their own runs.

## Mechanism

`apps/api/src/modules/oidc/auth/scopes.ts:42-88` and `plugins.ts:225`. Excluding `runs:read-all` from the **end-user** allowlist is correct. But `APPSTRATE_BUILTIN_SCOPES` is also the provider's global vocabulary, so no client can request it at all — while `claims.ts:73-76` would happily grant it if it could be asked for.

The CLI is unaffected (`actor_type: "user"` → `deferOrgResolution`, no scope ceiling — verified).

## Failure scenario

An org admin connects Claude Code over MCP-over-OAuth and finds `GET /api/runs` returns only the runs they launched themselves, with no explanation and no way to widen it.

## Proposed fix

Separate the end-user allowlist from the provider's vocabulary, so a scope can be requestable by a dashboard-user client while remaining ungrantable to an end-user token.

PR #1405, opened 2026-09-10
fix(oidc): make the consent screen honest about an instance token's reach, and pin registration-time scope as advisory

### LINKED ISSUE #1351: oidc: CIMD/DCR registration discards the declared scope and persists the full self-service ceiling

pierrecabriere:
Closes #1351

## Root cause

Two separate findings, one decision each.

1. `persistOAuthClientRegistration` (`@better-auth/oauth-provider` 1.7.3, `dist/authorize-9whjxVLJ.mjs:1936`) overwrites the persisted `scope` with `resolveClientRegistrationScopes(opts)` for `dynamic` and `clientMetadataDocument` registrations, so a declared `scope: "openid"` is stored as the whole self-service ceiling and `/authorize` (`:5558`, `client.scopes` is the ceiling) will hand it `mcp:invoke`.
2. The consent screen lists the requested scopes as the grant (`pages/consent.ts`), but for an `instance` client `scopesToPermissions` returns an empty set for `actor_type: "user"` (`auth/claims.ts:60`) and `lib/auth-pipeline.ts:150-163` writes no `scopeCeiling` — the request runs on the user's live org role.

## Fix

**(1) is not a bug — it is upstream's documented contract**, and the fix is to say so. `clientRegistrationAllowedScopes`' own docblock states the registration `scope` "does not narrow the persisted capability set or grant those scopes to a user"; the persisted set is an operator-approved capability set a later user authorization steps up within. Intersecting the declaration back in was rejected because **the declaration is client-controlled**: a registrant that wants the ceiling just declares it (DCR re-registers, CIMD edits its own document), so narrowing buys zero confinement — while it breaks every MCP client that publishes a minimal `scope` and then requests what the protected-resource metadata advertises (the `claude mcp add` path). What bounds a self-service client stays: the ceiling, the single-audience rule at `/oauth2/token`, consent, and the caller's live role.

Documented at the three seams the issue names — `auth/plugins.ts` where the options are set, the module README (new *Self-service client registration (DCR / CIMD)* section), and `markClientSelfService`, whose CIMD scope backfill was removed unexplained in #1287 and whose recorded symptom ("CIMD 1st hit = invalid_scope") invites restoring it. Pinned by a test asserting the stamp writes `level`/`self_service` and leaves `scopes` untouched.

**(2) is a real defect and is fixed.** `renderConsentPage` now takes a required `clientLevel` and, for `instance` only, renders what the token actually reaches. `org`/`space` tokens do carry the scope claim as a ceiling, so the notice would be false there; making it required rather than optional forces a new level to decide.

No behaviour change to any OAuth flow: no route contract, no scope vocabulary, no token shape moved.

## Tests

- `apps/api/src/modules/oidc/test/integration/services/dcr-cimd.test.ts` — new: *"stamps level and self-service WITHOUT rewriting the scope set"*. **Negative control run**: re-adding a `scopes:` write inside `markClientSelfService` turns it red on the sentinel assertion (`0 pass, 1 fail`), so the pin bites.
- `apps/api/src/modules/oidc/test/unit/pages.test.ts` — new: the reach notice renders for `instance`, and `it.each(["org","space"])` that it does not; 4 existing consent call sites updated for the required prop.

```
TEST_TIER=0 bun test .../unit/pages.test.ts .../unit/self-service-scopes.test.ts   → 32 pass / 0 fail
TEST_TIER=0 bun test .../services/dcr-cimd.test.ts                                 → 25 pass / 0 fail
TEST_TIER=0 bun test .../oauth-enduser-pages .../realm-isolation .../oauth-flows .../loopback-redirect
                                                                                   → 75 pass / 0 fail
TEST_TIER=0 bun test apps/api/test/unit/oauth-scope-labels.test.ts                 →  3 pass / 0 fail
bunx tsc --noEmit -p apps/api                                                      → clean
bunx eslint <touched files>                                                        → clean
```

## Notes for the orchestrator

- No migration, no env var, no OpenAPI change (the consent screen is server-rendered HTML, not a documented response body) — nothing to sequence at deploy.
- **Open decision, deliberately not taken**: making the consented scope a real ceiling for instance tokens (minting `scopeCeiling` from the claim) would also constrain the CLI and the dashboard SPA, which are instance clients too. That is a product decision; the PR ships the honest-copy half and records the trade-off in the README.
- **Overlap with #1372** (next in the chain, builds on this branch): it touches the same scope vocabulary area (`auth/scopes.ts`). This PR does not modify `auth/scopes.ts` at all — only `plugins.ts`, `oauth-admin.ts`, `pages/consent.ts`, `pages/layout.ts`, `routes.ts`, the README and two test files — so the surfaces should not collide.
- The `.notice` CSS lives in `pages/layout.ts` alongside the other consent-only classes (`.client`, `ul.scopes`), matching how that file is already organised.

### DIFF
--- a/apps/api/src/modules/oidc/README.md
+++ b/apps/api/src/modules/oidc/README.md
@@ -356,6 +356,24 @@ A public client is one whose `token_endpoint_auth_method` is `"none"` — that i
 - **PKCE is mandatory** for them: `isPKCERequired` returns "pkce is required for public clients" whenever `token_endpoint_auth_method` is `"none"`, whatever the client's own `require_pkce` says.
 - **Loopback ports are flexible** (RFC 8252 §7.3): the provider matches the requested redirect against the registered ones through `stripLoopbackRedirectPort`, so a client registered on `http://127.0.0.1/callback` matches `http://127.0.0.1:63785/callback` at runtime. It port-flexes `http:` loopback IP literals and the bare name `localhost`, and nothing else. Our registration gate (`services/redirect-uri.ts`, `isLoopbackHost`) is deliberately wider — it also admits `*.localhost` subdomains — so a `tenant.localhost` redirect registers but is matched port-exactly.
 
+## Self-service client registration (DCR / CIMD)
+
+Generic MCP clients onboard with no operator step, either by posting an RFC 7591 body to `/api/auth/oauth2/register` (DCR) or by identifying with a `client_id` URL that serves a Client ID Metadata Document (CIMD). Both paths land on `auth/plugins.ts` and are bounded by `getSelfServiceScopes()` — identity scopes plus every module scope opted in via `endUserGrantable: true`. A core action scope (`agents:run`, `llm-proxy:call`, …) is refused with `invalid_scope` at registration. Every such client is stamped `level = "instance"`, `self_service = true` by `markClientSelfService`, which is what confines its tokens to exactly one protected-resource audience at mint.
+
+### Registration-time `scope` is advisory
+
+A `scope` declared at registration is **validated** against the ceiling and then **discarded**: `persistOAuthClientRegistration` writes the whole ceiling to the row, so a client that publishes `scope: "openid"` is still able to request `mcp:invoke` at `/authorize`. This is Better Auth's documented contract — the persisted set is an operator-approved _capability_ set that a later user authorization steps up within, and a registration request "is not an authorization grant".
+
+The module keeps that behaviour rather than intersecting the declaration back in, for one reason: the declaration is client-controlled. A registrant that wanted the whole ceiling would simply declare it (DCR re-registers; CIMD edits its own document), so narrowing buys no confinement — while it would break every MCP client that publishes a minimal `scope` and then requests what the protected-resource metadata advertises. What actually bounds a self-service client is the ceiling itself, the single-audience rule at `/oauth2/token`, the consent screen, and the caller's live org role.
+
+Corollary: `markClientSelfService` writes `level` and `self_service` and nothing else. A scope backfill there would hand the client control of the ceiling `/authorize` enforces; it was removed in #1287 and must not come back.
+
+### What the consent screen promises
+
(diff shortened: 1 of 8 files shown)


pierrecabriere:

PR #1406, opened 2026-09-10
test(ee): make the Stripe contract suite catch an SDK bump, and cover Invoice

### LINKED ISSUE #1332: ee(stripe): the live contract suite missed this bump, cannot cover a Dependabot bump, and never touches Invoice

pierrecabriere:
Closes #1332

## Root cause

One mechanism behind all five points: **nothing could tell a fixture from a fiction.**

- `packages/module-ee/tsconfig.json` had `"include": ["src", "drizzle"]` — the module was the only workspace omitting `test`. So the API-version literal restated at `test/live/stripe-contract.test.ts:118` and every fixture in `test/helpers/stripe.ts` (all typed `Record<string, unknown>`) sat outside the typecheck program. An SDK bump drags `src/stripe/client.ts` forward via the `LatestApiVersion` literal type and leaves the test tree behind.
- Invoice had no fixture at all. `parent.subscription_details.subscription` — the post-basil replacement for the removed `invoice.subscription`, the single path `webhooks.ts:461` and `:743` hang budget allocation and dunning on — was hand-spelled at eleven call sites and disprovable at none.
- `.github/workflows/stripe-live.yml` gated the suite on `pull_request` + `secrets.STRIPE_LIVE_SECRET_KEY`. Dependabot-triggered runs read the *Dependabot* secret store, not repository secrets, so on a `stripe` bump the variable is empty, `describe.skipIf` skips all 26 tests, and the job is green. The workflow's own comment claimed the opposite.

## Fix

**One pin, in the program.** `STRIPE_API_VERSION` is exported from `src/stripe/client.ts` with `satisfies Stripe.LatestApiVersion` (a string *literal* type — the next bump is a hard `TS2322`). The live suite imports it instead of restating it.

**Typed fixtures — the half that needs no key.** `Fixture<T>` in `test/helpers/stripe.ts` is a deep-partial of the real SDK type: a fixture may omit any field production never reads, but every field it carries must exist with a compatible type. `test` joins the tsconfig, which required repairing the 9 pre-existing type errors the issue noted (`server.port`, an optional hook inside `Parameters<…>`, postgres.js `RowList` destructuring in the migration-chain tests). Negative control: renaming `current_period_end` → `current_period_endd`, or `subscription_details` → `subscription_detailsX`, both fail `tsc`.

Fixture fields are written as **direct properties**, never conditional spreads — TypeScript exempts spread properties from excess-property checking, which silently reopens the same hole. `JSON.stringify` drops the `undefined` ones, so each mocked test still sends only what it exercises.

**Invoice.** `invoiceEventObject()` is the one place the invoice payload is written; the 11 hand-built payloads in `webhooks.test.ts` and `webhook-emails.test.ts` now go through it. `fullInvoiceFixture()` (every optional field set) is drift-checked against a real invoice fetched from `subscription.latest_invoice`.

**Both halves share their read-path lists.** `SUBSCRIPTION_READS` / `INVOICE_READS` / `valueAtPath` move to the helper. The direction that needs no key — *the fixture carries every path production reads*, which the all-optional `Fixture<T>` structurally cannot catch — becomes `test/unit/stripe-fixtures.test.ts` and runs everywhere. The live suite keeps the direction only Stripe can answer.

**Webhook payload version.** New live assertion: every *enabled* webhook endpoint on the account must pin `STRIPE_API_VERSION`. The SDK pin governs what a `retrieve` returns; a webhook payload is rendered at the endpoint's configured version, and `subscriptionPeriodEnd` is applied to `event.data.object` too (`webhooks.ts:470ff`). `api_version: null` (account default) counts as a mismatch — it is pinned to nothing this repo can see.

**CI.** Added `push: branches: [main]` on the same paths: the *merge* of a Dependabot bump runs the suite with repository secrets. No second job, no new secret store. The `pull_request` trigger stays as a best-effort pre-merge signal, and the skip is now tolerated only where the secret cannot reach the run — on `push` / `schedule` / `workflow_dispatch` a missing key **fails** the job instead of skipping quietly.

### Running the live half

```sh
cd appstrate
docker compose -f test/setup/docker-compose.test.yml up -d --wait   # module-ee declares postgres: true
set -a && . ./.env && set +a
STRIPE_LIVE_SECRET_KEY=sk_test_… bun test packages/module-ee/test/live
```

The account needs two active recurring prices and a chargeable card (`pm_card_visa` is attached by the suite). A non-`sk_test_` key is refused outright. The key-free half needs none of this: `bunx tsc --noEmit -p packages/module-ee` and `bun test packages/module-ee/test/unit/stripe-fixtures.test.ts`.

## Tests

- `bunx tsc --noEmit -p packages/module-ee` → clean, now **including `test/`**.
- `bun test packages/module-ee/test` (full module, under the shared-Postgres lock) → **456 pass, 28 skip, 0 fail** (the 28 are the live suite skipping without a key).
- New `packages/module-ee/test/unit/stripe-fixtures.test.ts` → 16 pass, key-free. Negative control: dropping `attempt_count` from the invoice builder turns it red (1 fail).
- `bunx eslint` on the 8 touched TS files → clean. `bun run verify:workflows` (actionlint) → no findings. `bun run verify:license-boundary` → clean. `bunx knip --workspace packages/module-ee` → exit 0 (the one configuration hint predates this branch).

## Notes for the orchestrator

- **Base** is `fix/issue-1331`, per assignment.
- No migration, no env var, no OpenAPI change, no deploy ordering.
- **Overlap**: `test/integration/services/webhooks.test.ts` and `webhook-emails.test.ts` — 11 invoice payload literals rewritten. Sibling **#1370** may add tests to the first file; conflicts there would be textual, not semantic.
- **Expect the new webhook-endpoint assertion to be the suite's first real finding.** If the Stripe test account's endpoints are left at the account default (or an older pinned version), that test fails on the first keyed run. That is the bug it was written to surface, not a flake — the fix is to pin the endpoint's version in the Stripe dashboard.
- **Ops, optional**: adding `STRIPE_LIVE_SECRET_KEY` to the repository's *Dependabot* secret store would additionally give the pre-merge signal on Dependabot PRs. Not required — the `push` trigger covers the merge either way, which is why it was not made a hard dependency.

### DIFF
--- a/.github/workflows/stripe-live.yml
+++ b/.github/workflows/stripe-live.yml
@@ -7,9 +7,20 @@
 # when a fixture claims a field the live API does not have, or when Stripe stops
 # behaving the way the module bets it does.
 #
-# Without `STRIPE_LIVE_SECRET_KEY` the suite skips itself and this job is green
-# — the same opt-in shape as `scripts/conformance/probes.ts`. That covers fork
-# PRs, which never receive secrets.
+# Without `STRIPE_LIVE_SECRET_KEY` the suite skips itself — the same opt-in
+# shape as `scripts/conformance/probes.ts`. On a pull_request that skip is
+# allowed and this job stays green, because two kinds of PR structurally cannot
+# see repository secrets:
+#
+#   * fork PRs, which never receive them; and
+#   * DEPENDABOT PRs, which receive the Dependabot secret store instead. This is
+#     the reason the `pull_request` trigger alone cannot cover an SDK bump: the
+#     variable is empty, `describe.skipIf` skips every test, and the job reports
+#     green having checked nothing.
+#
+# The `push` trigger below is the half that does cover it. On push, schedule and
+# workflow_dispatch the repository secrets ARE available, so a missing key there
+# is a real failure and this job says so rather than skipping quietly.
 name: Stripe live contract
 
 on:
@@ -18,9 +29,16 @@ on:
   schedule:
     - cron: "0 7 * * 1"
   workflow_dispatch:
-  # Anything that touches the billing module or its Stripe dependency. A
-  # Dependabot bump of `stripe` edits packages/module-ee/package.json, so it
-  # lands here without anyone remembering to label it.
+  # The guaranteed run. A Dependabot bump of `stripe` edits
+  # packages/module-ee/package.json, so its MERGE lands here — with repository
+  # secrets — even though its PR run could not see them.
+  push:
+    branches:
+      - main
+    paths:
+      - "packages/module-ee/**"
+      - ".github/workflows/stripe-live.yml"
+  # Best-effort pre-merge signal for human PRs touching the billing module.
   pull_request:
     paths:
       - "packages/module-ee/**"
@@ -61,12 +79,22 @@ jobs:
       - name: Start test infrastructure
         run: docker compose -f test/setup/docker-compose.test.yml up -d --wait
 
+      # A skipped suite reporting green is the exact failure this job exists to
+      # prevent, so the skip is tolerated only where the secret CANNOT reach the
+      # run: a pull_request from a fork or from Dependabot.
       - name: Report coverage state
         run: |
-          if [ -z "${STRIPE_LIVE_SECRET_KEY}" ]; then
-            echo "::notice::STRIPE_LIVE_SECRET_KEY is not set — the live contract suite will skip itself."
+          if [ -n "${STRIPE_LIVE_SECRET_KEY}" ]; then
+            exit 0
           fi
+          if [ "${EVENT_NAME}" = "pull_request" ]; then
+            echo "::warning::STRIPE_LIVE_SECRET_KEY is not visible to this run — the contract suite will skip itself. Expected on a fork PR, and on a Dependabot PR (those read the Dependabot secret store, not repository secrets). The push-to-main run covers the merge."
+            exit 0
+          fi
+          echo "::error::STRIPE_LIVE_SECRET_KEY is empty on a ${EVENT_NAME} run, where repository secrets ARE available — the secret is missing, and the contract suite would skip itself while reporting green."
+          exit 1
         env:
+          EVENT_NAME: ${{ github.event_name }}
           STRIPE_LIVE_SECRET_KEY: ${{ secrets.STRIPE_LIVE_SECRET_KEY }}
 
       - name: Run Stripe live contract suite
--- a/packages/module-ee/README.md
+++ b/packages/module-ee/README.md
@@ -607,17 +607,47 @@ therefore returning `null` in every test that touched it. The confirmation
 e-mail silently fell back to today's date, and no test noticed, because the
 fixture and the assertions were written from the same stale belief.
 
-`test/live/stripe-contract.test.ts` is the other half. It runs against real
-Stripe in **test mode** and checks the two things the mock structurally cannot:
+Two things close that gap, and they are worth separating because only one of
+them needs a Stripe key.
 
-- **Shape** — every key path a fixture claims must exist on the live object.
-  Extra live fields are ignored (the fixtures are deliberately minimal); invented
-  ones fail. This is what turns the next relocation into a red mock rather than a
-  quiet production lie.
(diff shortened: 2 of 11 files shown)


pierrecabriere:

Issue #1366, opened 2026-09-10
auth: the distributed lock has no renewal, so a critical section past its 45s TTL still double-spends the refresh token


pierrecabriere:
## Summary

The distributed lock has no renewal, so a critical section longer than the 45 s TTL still permits the double-spend the deduped-refresh work targets.

## Mechanism

`apps/api/src/lib/distributed-lock.ts:67-75` polls at 100 ms until `deadline`; the holder's lock is `SET … EX 45 NX` with no watchdog.

A holder wedged past 45 s — exchange plus `reReadFreshness`'s DB read under pressure — has its lock lapse. The waiter acquires and runs `doRefresh()` concurrently, both spend the rotating `refresh_token`, the loser writes a dead one, and the credential is flagged `needsReconnection`.

The release script is compare-and-delete on `lockId`, so the holder at least cannot delete the new owner's lock.

Pre-existing; #1279 narrowed the window but did not close it.

## Proposed fix

Renew the lock while the critical section runs (a watchdog refreshing the TTL), or fence the write so a lapsed holder's update is rejected.

PR #1407, opened 2026-09-10
fix(auth): let a forced refresh adopt an in-flight proactive exchange

### LINKED ISSUE #1365: auth: the per-credential refresh chain puts 401 recovery behind head-of-line blocking on the run inference path
## Summary

The per-credential refresh chain added by the stability audit (#1279) correctly closes the no-Redis concurrency hole, but it puts the 401-recovery path behind head-of-line blocking on the sidecar's hot path. A busy credential can stall an agent run for over two minutes where it previously could not.

## Mechanism

`apps/api/src/lib/deduped-refresh.ts:93-118`. The chaining itself is correct — the `refreshChains.get(key) === tail` guard was traced through interleaved settle orders with no leak or stale-tail bug, and `tail = promise.catch(() => {})` correctly prevents both cascade and unhandled rejection.

The cost is admitted in-code at `:38-46`: a forced refresh chained behind a proactive one pays ~50 s acquire + 30 s exchange **twice** — *"the honest worst case is ≈160 s"*.

`forceRefreshOAuthModelProviderToken` is called from `resolveOAuthTokenForSidecar` (`apps/api/src/services/model-providers/token-resolver.ts:129`) — the run's inference path.

Chain depth stays ~2-3 in practice (same-kind callers still collapse into one flight), so this is a tail-latency regression, not an unbounded queue.

## Proposed fix

Give the forced path its own deadline, or let a caller that only needs *a* valid token adopt the in-flight proactive refresh's result instead of queueing a second exchange behind it.

pierrecabriere:
Closes #1365

## Root cause

`apps/api/src/lib/deduped-refresh.ts:94` partitions the singleflight on `force`, so a forced (401-recovery) caller can never *share* an overlapping proactive flight — it chains behind it in `refreshChains` (`:98-99`) and then runs a full second flight of its own. Each hop costs up to `REFRESH_LOCK_ACQUIRE_TIMEOUT_MS` (~50 s, `:47`) plus the 30 s exchange timeout in `@appstrate/connect`, so the forced path paid ~160 s — admitted in-code at `:38-46`.

That path is the run inference hot path: `resolveOAuthTokenForSidecar` → `forceRefreshOAuthModelProviderToken` (`apps/api/src/services/model-providers/token-resolver.ts:129,174`), reached from the sidecar's `POST /internal/oauth-token/:credentialId/refresh` (`apps/api/src/routes/internal.ts:457`).

## Fix

The partition itself is correct and stays: a forced caller must not be handed a value the flight produced by *short-circuiting on freshness*, because that is the very token that just 401'd it. But a flight that actually **exchanged** minted a token upstream — by construction not the stored one the caller 401'd on — and that is exactly what the forced caller asked for.

So a flight now resolves to `FlightOutcome<T> = { value, exchanged }`, recording which branch answered it. A forced caller with no `key:force` flight to join awaits the in-flight proactive one and adopts its value when `exchanged` is true, saving a whole lock-wait + exchange hop. A freshness short-circuit or a failure is not adoptable and re-enters with the caller's own flight — by which time the chain tail has cleared, so it starts immediately rather than queueing. Re-entry cannot storm the upstream: the first re-entrant publishes the `key:force` flight the rest collapse into (covered by a test).

Same-kind joins, the chain, the Redis lock and every failure semantic are untouched — all six pre-existing tests pass unmodified.

Considered and rejected: the issue's other suggestion, a separate deadline for the forced path. A deadline that fires can only throw, which turns a slow recovery into a failed run; adoption removes the wasted hop instead of capping it.

## Tests

`apps/api/test/unit/deduped-refresh.test.ts` — three new cases: adoption of an in-flight proactive exchange, no inheritance of a failed proactive flight, and no upstream storm when several forced callers re-enter after a non-adoptable outcome.

```
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/unit/deduped-refresh.test.ts
→ 9 pass, 0 fail (6 pre-existing + 3 new)
bunx tsc --noEmit -p apps/api → clean
bunx eslint / prettier --check on both touched files → clean
```

Negative control: with `apps/api/src/lib/deduped-refresh.ts` reverted to `HEAD` and the new tests in place, the adoption test fails with `Expected: "EXCHANGED" / Received: "SECOND-EXCHANGE"` — i.e. it directly observes the extra exchange this PR removes.

## Notes for the orchestrator

- No migration, no env var, no OpenAPI/route-contract change (the internal refresh endpoint's request and response are untouched), no `@appstrate/core` export change.
- **Both inference paths checked**, per the standing rule that they must behave alike: `/api/llm-proxy` refuses oauth2-subscription models outright (`apps/api/src/services/llm-proxy/core.ts:157`) and the chat path only ever calls the *proactive* `resolveOAuthTokenForSidecar` (`apps/api/src/services/chat-platform-services.ts:95`). The sidecar's `/refresh` route is the only forced caller on a hot path, and every caller — including the background `refresh-worker` and the integration-connection path (`apps/api/src/services/integration-token-refresh.ts:131`) — funnels through this one helper, so the fix covers them all uniformly.
- **Overlap with #1366 (lock renewal): none.** That issue is `apps/api/src/lib/distributed-lock.ts`; this PR touches only `deduped-refresh.ts` and does not change how `withRedisLock` is called (same TTL, same acquire timeout, same label). A renewal fix drops in underneath cleanly.
- Not addressed, deliberately (out of scope, not a regression): the mirror-image hop, where a *proactive* caller queues behind an in-flight forced flight instead of adopting its freshly minted token. Same one-extra-exchange cost, but off the 401-recovery path this issue is about.
- Pre-existing local test noise, identical before and after this change (22 pass / 28 fail on both): the `services/` integration files seed their `test-oauth` provider through a suite-wide preload, so they fail with `Unknown providerId: test-oauth` when run as a subset. Unrelated to this diff.

### DIFF
--- a/apps/api/src/lib/deduped-refresh.ts
+++ b/apps/api/src/lib/deduped-refresh.ts
@@ -39,10 +39,11 @@ const REFRESH_LOCK_TTL_SECONDS = 45;
  * Derived from the TTL so the two cannot drift apart: a waiter gives up only
  * once the holder's lock has definitively expired (holder presumed dead),
  * never at the holder's worst-case exchange duration. Trade-off: a sidecar
- * caller queued behind a wedged holder waits up to ~50 s before proceeding —
- * and a forced flight chained behind a proactive one pays that twice, once per
- * hop (~50 s acquire + the 30 s exchange timeout in `@appstrate/connect`), so
- * the honest worst case is ≈160 s. Bounded, but not one lock wait.
+ * caller queued behind a wedged holder waits up to ~50 s before proceeding,
+ * plus the 30 s exchange timeout in `@appstrate/connect`. A forced caller that
+ * arrives while a proactive flight is already exchanging pays that once, not
+ * once per hop — it adopts that exchange's token rather than queueing a second
+ * one behind it (see {@link dedupedRefresh}).
  */
 const REFRESH_LOCK_ACQUIRE_TIMEOUT_MS = REFRESH_LOCK_TTL_SECONDS * 1_000 + 5_000;
 
@@ -72,8 +73,19 @@ interface DedupedRefreshOptions<T> {
   doRefresh: () => Promise<T>;
 }
 
+/**
+ * A flight's value plus WHICH branch produced it. A forced joiner may adopt an
+ * in-flight proactive result only when it came from the exchange — see
+ * {@link dedupedRefresh}.
+ */
+interface FlightOutcome<T> {
+  value: T;
+  /** `true` when {@link DedupedRefreshOptions.doRefresh} minted this value. */
+  exchanged: boolean;
+}
+
 /** Per-flight-key in-flight singleflight map (`key` / `key:force`). */
-const inflightRefreshes = new Map<string, Promise<unknown>>();
+const inflightRefreshes = new Map<string, Promise<FlightOutcome<unknown>>>();
 /** Tail of the serialization chain per credential `key`, never rejecting. */
 const refreshChains = new Map<string, Promise<unknown>>();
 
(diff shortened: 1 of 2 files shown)


pierrecabriere:

PR #1408, opened 2026-09-10
fix(roles): gate the custom_roles licence on propagation, not on cleanup

### LINKED ISSUE #1334: rbac: the custom_roles feature gate is fail-open on member assignment and fail-closed on cleanup

pierrecabriere:
Closes #1334

## Root cause

`assertCustomRolesFeature()` was wired to exactly three routes (`apps/api/src/routes/roles.ts:84,100,114`) and nowhere else. `assignmentColumns` (`apps/api/src/services/space-members.ts:265`) and `assertSpaceAssignmentsValid` (`apps/api/src/services/space-assignments.ts:52`) — the two places a bundle is actually GRANTED, from the space-member routes, invitations and OAuth signup policies — never asked. So with `custom_roles` off a leftover bundle could still be handed to a brand-new member, while `DELETE /api/roles/:id`, the verb that would remove it, answered 403.

The gate had been drawn around a noun ("the write routes") rather than around what those verbs do to a bundle's reach. RBAC spec §9 said who may DEFINE a bundle and said nothing about the ones already stored, which is how the two halves came to disagree.

## Fix

The licence now follows the direction of change, and §9 says so:

- **Grows a bundle's reach → licensed.** `POST/PATCH /api/roles` (unchanged), plus the two grant paths: `assignmentColumns` and `assertSpaceAssignmentsValid`. The latter covers invitations, OAuth client signup policies, and each replay of a standing policy for a new signup — the one replay that would otherwise keep propagating bundles indefinitely after a downgrade.
- **Shrinks it → never licensed.** `DELETE /api/roles/:id` loses its gate; its existing 409 `role_in_use` already stops it stranding a holder. Moving a holder onto a preset was, and stays, open.
- **Reads → unchanged**, with one correction: `GET /api/spaces/:id/roles` answers "what can I assign here", so it stops offering bundles nobody can grant. `GET /api/roles` keeps listing them — that is where a downgraded deployment finds a leftover to delete. This also fixes the picker the issue mentions with no SPA change.

One deliberate exception, documented in code and in §9: accepting an invitation authored while the feature was on still applies its custom role. That grant was made under licence, it expires on its own, and refusing it would strand an invitee who has no way to fix it.

Two small refactors the fix needed: `assertCustomRolesFeature(act)` takes `"define" | "assign"` so the 403 names the act instead of always saying "Defining"; and the two-cause 403 (`forbidden` / `feature_unavailable`), previously a local constant in `paths/roles.ts`, becomes a shared `CustomRoleFeatureForbidden` response reused by the six operations that now carry it.

## Tests

`apps/api/test/integration/routes/roles.test.ts` — the gate block is now four cases, each paired with the twin that differs by exactly one thing: authoring 403s then succeeds; granting 403s on the member route and the invitation route while the same two calls carrying a **preset** go through; the space catalogue drops the bundle while the org catalogue keeps it; a held leftover is reassigned to a preset and deleted with the feature off.

Four suites that assign or list a bundle to exercise something else now declare `setFeatureFlag("custom_roles", true)` — they passed before only because the grant asked no licence: `space-member-management`, `invitation-space-assignments`, `oidc/.../org-signup-space-assignments`.

```
  apps/api/test/integration/routes/{roles,space-member-management,invitation-space-assignments,space-role-delegation,rbac-catalog-boundaries,view-as,invitations,organizations}.test.ts \
  apps/api/src/modules/oidc/test/integration/services/{org-signup-space-assignments,orgmember-mapping}.test.ts
→ 204 pass, 0 fail, 747 expect()

bunx eslint <touched .ts>       → clean
bun run verify:api-types        → up to date (regenerated schema.d.ts committed)
bun run detect:breaking         → 0 breaking, 2 non-breaking (both from the #1343 base)
bun run verify:no-migration-dml → clean
```

## Notes for the orchestrator

- **Base is `fix/issue-1343`** (space-roles chain). No overlap with #1371 (`DELETE /api/roles/:id` 500 on concurrent assignment) beyond the fact that both touch that route: this PR only removes a middleware from its registration and does not touch `deleteSpaceRole`. No overlap with #1333.
- **No migration, no env var.**
- **Behaviour change on an OSS deployment that never had the feature**: none — no bundle can exist there, so the new refusals are unreachable. The change is only visible on an EE → OSS downgrade.
- **One fail-closed consequence worth naming at deploy time**: after a downgrade, an OAuth client whose standing `signupSpaceAssignments` names a custom role stops auto-joining new users (`OrgSignupConfigurationError`, which already names the policy as the thing to fix) instead of silently granting an unlicensed bundle at every login. That is the intended direction, but it is the one case where the fix turns a working flow into a refusal.
- `apps/web/src/api/schema.d.ts` is regenerated output, committed as required.


pierrecabriere:

PR #1409, opened 2026-09-10
feat(db): diff live-database columns, not just indexes, against the snapshot

### LINKED ISSUE #1349: db: the Better Auth schema check validates the Drizzle object, never the database

pierrecabriere:
Closes #1349

## Root cause

The issue's mechanism is accurate about upstream: `findDrizzleSchemaProblems`
(`@better-auth/drizzle-adapter`) diffs `getExpectedSchema(options)` against
`introspectDrizzleSchema(schema)`, and all three findings `diffSchema` can emit
(`missing-table`, `missing-column`, `unexpected-required-column`) are computed
from the Drizzle **TypeScript object**. It never reads `information_schema`, and
it does not look at indexes at all.

Its conclusion needed narrowing, though. `apps/api/test/unit/migration-schema-parity.test.ts`
— which predates the audit base — already closes the other link of the chain in
CI: it replays the whole journal into a throwaway PGlite and diffs `pg_catalog`
against `packages/db/src/schema/` (columns, nullability, constraints, FK
actions, indexes, enums), and `schema/oidc.ts` is in the barrel, so the Better
Auth tables are covered. A migration that misses a declared column therefore
**cannot reach main** today.

What no check in the repo could see is the case the issue's own "why it matters"
paragraph names: a **live** database that never ran `0000_init.sql`. A replay
structurally cannot model it. `scripts/check-index-drift.ts` is the only tool
pointed at production, and its documented scope was index **names only**
(`scripts/check-index-drift.ts:39` on the base) — so a missing *column* on
production, the exact 42703 the issue describes, was invisible everywhere.

## Fix

`scripts/check-index-drift.ts` now diffs columns as well as indexes, against the
same watermark snapshot, reusing every piece of plumbing already there (watermark
resolution, module-owned-table subtraction, the pure `runCheck` seam, the
`Cannot check` refusals):

- **absent column** → exit 1, named. **NOT NULL disagreement** → exit 1, named.
- **undeclared column** → reported, never fatal (pre-squash residue), matching
  how undeclared indexes are already treated.
- Both halves are reported **before** the exit code is decided, so an operator
  missing an index *and* a column sees both in one run rather than discovering
  the second after a second deploy window.
- Column **types** are not compared, and a declared table with no column row is
  skipped rather than reported as missing — `information_schema.columns` is
  privilege-filtered, so its silence about a table is not evidence. Both
  carve-outs are stated in the header and in the printed output.

**Design decision — this is deliberately NOT a boot gate**, which is the half of
the issue's "Proposed fix" I did not take. Production is known to predate the
squash (#1182 found two indexes missing there), so gating boot on this
comparison would refuse to start the platform over drift that has been live for
months. It stays an operator check run against a connection string, whose exit
code is a decision a human makes. The reasoning is written into the script header
and `packages/db/README.md` so it is not re-litigated by guesswork.

The filename is unchanged even though the script is now broader than "index
drift": `packages/db/drizzle/0044_finish_file_rename.sql` names this path in
operator guidance, and a shipped migration is immutable. The header says so.

Also fixed: the comment at `packages/db/src/schema/oidc.ts:24-27` that the issue
correctly calls "easy to misread as covering the database" now states plainly
what the adapter check does and does not do, and points at the two checks that
actually read a database.

## Tests

`scripts/test/check-index-drift.test.ts` — 5 new `runCheck` cases (missing
column; NOT NULL disagreement; index **and** column reported in one run;
undeclared column non-fatal; module-owned table ignored) plus unit cases for
`declaredColumns` and `diffColumns`.

```
set -a && . ./.env && set +a && TEST_TIER=0 bun test scripts/test/check-index-drift.test.ts
  → 24 pass, 0 fail, 53 expect()   (19 before)

set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/unit/migration-schema-parity.test.ts
  → 8 pass, 0 fail
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/unit/migration-index-parity.test.ts
  → 4 pass, 0 fail

bunx tsc --noEmit -p scripts/tsconfig.json   → clean
bunx tsc --noEmit -p packages/db             → clean
bunx tsc --noEmit -p apps/api                → clean
bunx eslint <touched files>                  → clean
bun run verify:dead-code                     → exit 0
bun run verify:no-migration-dml              → green (69 files, 2 trees)
```

## Notes for the orchestrator

- **No migration, no env var, no OpenAPI change.** Nothing in `bun run check`
  changes behaviour; `verify:dead-code` and `verify:no-migration-dml` were run
  because the diff adds exports and touches `packages/db`.
- **Pre-existing flake, not mine.** Running
  `scripts/test/check-index-drift.test.ts` + both parity tests in **one** bun
  process fails one of them with `a beforeEach/afterEach hook timed out`
  (5 s hook cap, two PGlite journal replays contending). I verified this on a
  control worktree at the unmodified base `origin/fix/issue-1348`: it fails there
  too, on a *different* file of the three. Each file passes alone on both
  branches.
- **Deploy ordering:** none. This is an operator script; nothing runs it
  automatically. Running it against production after the next deploy is likely to
  print real findings — that is the point, and the exit code is advisory to a
  human, not to a pipeline.
- **Overlap:** none with the chain so far. #1347/#1348 touched the better-auth
  façade, the OAuth FK indexes and `packages/db/src/schema/oidc.ts`'s table
  declarations; this PR only appends to that file's header comment.

### DIFF
--- a/apps/api/test/unit/migration-schema-parity.test.ts
+++ b/apps/api/test/unit/migration-schema-parity.test.ts
@@ -29,8 +29,9 @@
  *
  *   - `drizzle-kit generate` compares the schema to the snapshot. Both carry
  *     the 70-byte name, so there is no drift to report.
- *   - `scripts/check-index-drift.ts` compares INDEX NAMES ONLY, against a LIVE
- *     `DATABASE_URL`. It cannot see a constraint, and it cannot run in `check`.
+ *   - `scripts/check-index-drift.ts` compares index names and column
+ *     name/nullability against a LIVE `DATABASE_URL`. It cannot see a
+ *     constraint, and it cannot run in `check`.
  *   - `migration-index-parity.test.ts` compares indexes to the snapshot.
  *   - the whole suite passes, because the FK works perfectly. Only its NAME is
  *     wrong, and nothing addresses a constraint by name until something does.
--- a/packages/db/README.md
+++ b/packages/db/README.md
@@ -35,12 +35,12 @@ bun run db:generate   # Generate migration from schema changes
 bun run db:migrate    # Apply pending migrations
 ```
 
-## Index drift
+## Schema drift
 
 `drizzle/0000_init.sql` is a squash, and any database created before it — production — never ran
-it. An index the squash introduced therefore exists in the schema and in every fresh database while
-being absent there (issue #1182 found two). Before any `DROP INDEX`, verify the surviving index
-against the **live** database; the TS schema and `0000_init.sql` are not evidence.
+it. Anything the squash introduced therefore exists in the schema and in every fresh database while
+being absent there (issue #1182 found two indexes). Before any `DROP INDEX`, verify the surviving
+index against the **live** database; the TS schema and `0000_init.sql` are not evidence.
 
 ```sh
 DATABASE_URL=postgres://… bun scripts/check-index-drift.ts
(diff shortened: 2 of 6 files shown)


pierrecabriere:

PR #1411, opened 2026-09-10
fix(web): commit a previewed persona together with its permissions

### LINKED ISSUE #1322: web(view-as): the persona activates before its permissions are refreshed, and stays stale if the refresh fails
## Summary

Entering a role preview commits the persona (the banner renders immediately) but leaves the previous org-level permissions live in the React Query cache until a refetch lands — and **permanently** if that refetch fails. Org-level admin menus stay visible under a "Lecteur" banner.

## Mechanism

`apps/web/src/stores/view-as-store.ts:129`:

```ts
function resetScopedCache(): void {
  removeOrgScopedQueries(queryClient);
  void queryClient.refetchQueries({ queryKey: orgKeys.all });
}
```

`removeOrgScopedQueries` (`apps/web/src/lib/query-keys.ts:176`) deliberately **spares** `["orgs"]`:

```ts
qc.removeQueries({ predicate: (q) => q.queryKey[0] !== orgKeys.all[0] });
```

The reason given is that `OrgGate` blocks on it and removing it flashes the boot screen. But React Query keeps the previous `data` during a background refetch **and** on error, so `currentOrg.permissions` — which `usePermissions().can()` reads (`apps/web/src/hooks/use-permissions.ts:62`) — stays the owner's set. The `void` discards the promise: nothing observes the failure.

## Why the symptom looks the way it does

The **space** half of the cache *is* dropped. So space-level gates close correctly while org-level gates stay open — which is exactly the reported symptom: Webhooks, OAuth clients, CLI sessions, billing and the other owner/admin-only org strings remain in the menu under a Lecteur banner.

Without a network error this is a sub-second window. With `/api/orgs` failing (reproduced in Chromium with a simulated 503) it is permanent and silent.

## Not an authorization bypass

The server is not at fault. A differential test over the whole matrix `{member, guest} × {admin, builder, operator, runner, viewer}` compared what `GET /api/orgs` and `GET /api/spaces` answer for a previewed persona versus a **real user holding that role pair**: `NO DIFFERENCE across the whole matrix` — same role, same `access`, same `permissions` arrays, second open space included. Every request under the preview is still enforced server-side; only what the SPA *renders* is wrong.

## Proposed fix

The persona and its permissions must land together. Options:

- Fetch the persona's org row **before** committing the persona, and commit both atomically.
- Or commit the persona and hold `can()` in an explicit not-ready state until the refetch resolves, with a visible error state if it fails (and an exit from the preview on failure — `endPreviewIfRefused` already exists for the refusal case).

What must not happen is presenting the previous authority as the persona's.

pierrecabriere:
Closes #1322

## Root cause

`apps/web/src/stores/view-as-store.ts:129` committed the persona and then only
*refetched* the org list — `["orgs"]` is deliberately spared by
`removeOrgScopedQueries` (`apps/web/src/lib/query-keys.ts:175`) because `OrgGate`
blocks on it. React Query keeps serving the previous `data` for the whole of a
background refetch **and** on error, and the `void` discarded the promise, so
`currentOrg.permissions` — what `usePermissions().can()` reads
(`apps/web/src/hooks/use-permissions.ts:65`) — stayed the previewer's. The space
half of the cache *is* dropped, which is why space gates closed correctly while
Webhooks / OAuth clients / CLI sessions / billing stayed in the menu under a
"Lecteur" banner. Sub-second normally, permanent when `/api/orgs` fails.

Server-side enforcement was never at fault (the issue's differential test over
`{member, guest} × {admin, builder, operator, runner, viewer}` found no difference).

## Fix

Make the pairing structural instead of hoping a refetch wins the race.

- `enterViewAs(persona, orgs)` now **takes** the org listing already answered for
  that persona and writes it with `setQueryData` in the same synchronous tick as
  the commit. The two can no longer disagree, and the signature makes it
  impossible for a future call site to skip the step.
- `fetchOrgsAs(persona)` (`apps/web/src/hooks/use-org.ts`) asks `GET /api/orgs`
  with an explicit `X-View-As` header — the store does not hold the persona yet,
  which is the point. `fetchOrgs` gained the optional header argument; the
  request middleware only fills headers that are absent, so the explicit one wins.
- `ViewAsDialog.submit` loads first, disables the button while it does, and on
  failure shows `viewAs.enterFailed` and keeps the dialog open. **No persona is
  committed at all**, so a failed load leaves the admin exactly where they were —
  no rollback state to get wrong, no stale persona to strand.
- `exitViewAs` keeps its background refresh (leaving *restores* authority, so a
  stale row under-states rather than over-states, and exit must work with no
  network at all: sign-out, org switch, a refused persona) but now
  `invalidateQueries` instead of `refetchQueries`, so a refresh that fails leaves
  the row marked stale and the next observer retries. With
  `refetchOnWindowFocus: false` and `retry: 1` in the query client, the old form
  could keep the persona's reduced row for good.
- `resetScopedCache()` is gone — the two paths are deliberately no longer the
  same, and each now carries the reason inline.

No change to `OrgGate`, to `removeOrgScopedQueries`, or to `usePermissions`: the
boot-screen flash the sparing rule exists to avoid is still avoided, because the
listing is seeded rather than dropped.

New i18n key `viewAs.enterFailed` (fr + en, `settings` namespace).

## Tests

`apps/web/src/stores/test/view-as.test.tsx` — every `enterViewAs` call site now
passes the listing it commits with; the cache-reset block was rewritten for the
new contract and gained the regression:

- *"publishes the persona's own org row in the same tick as the persona"* — seeds
  `["orgs"]` with an owner row carrying `webhooks:read`, enters, and asserts the
  cached row is the persona's. Fails on `main`.
- *"empties the scoped cache on entry"* — entry no longer refetches/gambles.
- *"empties it again on exit and re-asks for the caller's own org row"* — pins
  `invalidateQueries({ queryKey: ["orgs"] })`.

```
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/web/src/stores/test/view-as.test.tsx
  → 31 pass / 0 fail
  apps/web/src/pages/test/view-as-entry-points.test.tsx \
  apps/web/src/lib/test/query-keys.test.ts \
  apps/web/src/stores/test/view-as.test.tsx \
  apps/web/src/hooks/test/assignable-space-roles.test.tsx
  → 49 pass / 0 fail
bunx tsc --noEmit -p apps/web  → clean
bunx eslint <touched files>    → clean
```

## Notes for the orchestrator

- No migration, no env var, no OpenAPI contract change (`X-View-As` is already a
  declared header parameter on `listOrganizations`).
- **Pre-existing flake, not from this branch**: `apps/web/src/locales/test/locale-keys.test.ts`
  → *"declared keys are all still referenced by the source"* times out at its 5 s
  budget under the current parallel-agent load. Verified against a clean
  `origin/main` worktree in the same conditions: it times out there too (5.3 s).
  The guard does an O(keys × namespaces) substring sweep over a multi-MB blob.
- Overlap with #1368 (identity-derived capabilities not previewable): none in
  code. That issue builds on this one — `enterViewAs` now has a second parameter,
  so a branch that adds another previewable dimension should extend the same
  "load it, then commit it with the persona" shape rather than adding a second
  refetch.
- Behaviour worth knowing: submitting a *replacement* persona that the server
  refuses will now end the preview that was running (the response middleware sees
  the refusal on the loading request) **and** show the dialog's failure message.
  End state is correct — no preview, an explicit error — but the user may see two
  toasts. Left unguarded on purpose: the entry catalog is already filtered to
  grantable roles and the trigger only renders for owner/admin, so it needs a
  concurrent permission change to happen at all.

### DIFF
--- a/apps/web/src/components/view-as-dialog.tsx
+++ b/apps/web/src/components/view-as-dialog.tsx
@@ -2,6 +2,7 @@
 
 import { useState } from "react";
 import { useTranslation } from "react-i18next";
+import { toast } from "sonner";
 import { VIEW_AS_ORG_ROLES, type ViewAsOrgRole } from "@appstrate/core/permissions";
 import { Button } from "@appstrate/ui/components/button";
 import { Field, FieldGroup } from "@appstrate/ui/components/field";
@@ -16,7 +17,7 @@ import {
 import { Modal } from "./modal";
 import { OrgRoleOptions } from "./org-role-options";
 import { RoleCatalogState } from "./role-catalog-state";
-import { useCurrentOrgId } from "../hooks/use-org";
+import { fetchOrgsAs, useCurrentOrgId } from "../hooks/use-org";
 import { useCurrentSpaceId, useSpaceSwitcher } from "../hooks/use-current-space";
 import { useSpaces } from "../hooks/use-spaces";
 import { useSpaceRoleOptions } from "../hooks/use-roles";
@@ -56,6 +57,7 @@ export function ViewAsDialog({ onClose, spaceId }: ViewAsDialogProps) {
   const [orgRole, setOrgRole] = useState<ViewAsOrgRole>("member");
   const [selectedSpaceId, setSelectedSpaceId] = useState(initialSpaceId);
   const [roleValue, setRoleValue] = useState("");
+  const [entering, setEntering] = useState(false);
 
   const inSpace = selectedSpaceId !== NO_SPACE;
   const {
@@ -73,11 +75,23 @@ export function ViewAsDialog({ onClose, spaceId }: ViewAsDialogProps) {
   const catalogUsable = rolesKnown && !rolesError && options.length > 0;
   const ready = !!orgId && (!inSpace || (!!roleOption && !!space));
 
-  const submit = () => {
-    if (!orgId || !ready) return;
+  const submit = async () => {
+    if (!orgId || !ready || entering) return;
     // Replaces any preview already running: the store commits one persona and
     // resets the cache either way.
-    enterViewAs(toViewAsPersona(orgId, orgRole, space, roleOption));
+    const persona = toViewAsPersona(orgId, orgRole, space, roleOption);
+    setEntering(true);
+    try {
+      // Load the persona's own org row BEFORE committing it: the permissions
+      // and the banner must appear together, and a load that fails must leave
+      // the admin exactly where they were rather than under a persona wearing
+      // their own authority.
+      enterViewAs(persona, await fetchOrgsAs(persona));
+    } catch {
+      setEntering(false);
+      toast.error(t("viewAs.enterFailed"));
+      return;
+    }
     // Land where the persona's role applies. Elsewhere the persona is only its
     // org role — an implicit member of open spaces — and a banner naming
     // "Lecteur dans Default" over a page answered for another space reads as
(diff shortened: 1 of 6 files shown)


pierrecabriere:

PR #1412, opened 2026-09-10
fix(roles): answer a delete/assign race with 409 role_in_use, not a 500

### LINKED ISSUE #1371: rbac: DELETE /api/roles/:id 500s on a concurrent assignment instead of returning 409

pierrecabriere:
Closes #1371

## Root cause

`apps/api/src/services/space-roles.ts` — `deleteSpaceRole()` counted the role's
holders (`space_members` rows + PENDING invitations naming it) with a lock-free
read, then issued `db.delete(spaceRoles)` with no `.catch`. `space_members.custom_role_id`
is `ON DELETE RESTRICT` (`packages/db/src/schema/spaces.ts:118`), so a
`POST /api/spaces/:id/members {custom_role_id}` landing between the count and the
delete makes Postgres raise a referential-integrity violation that nothing
handles — the intended 409 `role_in_use` becomes a 500.

## Fix

Guard the delete the way `createSpaceRole`/`updateSpaceRole` already guard their
key races (`asKeyConflict`): map the violation back to the *same* 409. The
counting and the throwing are split into `countRoleHolders()` / `roleInUse()` so
the racing path recounts and the problem body still carries `member_count` and
`pending_invitation_count` — the two fields the OpenAPI description promises and
`apps/web/src/pages/org-settings/roles.tsx:77` reads. No contract change: same
status, same code, same details shape.

Chosen over "count and delete in one transaction with `SELECT … FOR UPDATE`":
that lock does serialise the two writers, but it makes the *delete* win, so the
concurrent assignment is the side that dies on the FK. Refusing the delete is
the semantics the pre-count already states ("in use → 409"), and catching the
violation gets there without holding a lock.

The predicate for that violation already existed — privately, with its own
cause-chain walk, in `apps/api/src/routes/model-provider-credentials.ts`. First
commit promotes it to `apps/api/src/lib/db-helpers.ts` beside `isUniqueViolation`
rather than writing a third copy of the walk. It keeps covering both SQLSTATEs
(`23503` on a real Postgres, `23001` on PGlite for tier 0).

## Tests

- New `apps/api/test/unit/lib/db-helpers.test.ts` (3 `it`s) pins the promoted
  predicate: both SQLSTATEs, seen through the Drizzle wrapper, and rejecting
  neighbouring codes (`23505`, `22P02`) and non-DB values.
- `set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/unit/lib/db-helpers.test.ts apps/api/test/integration/routes/roles.test.ts` → **27 pass, 0 fail**
- `… bun test apps/api/test/integration/routes/model-provider-credentials.test.ts` → **27 pass, 0 fail** (its `credential_in_use` case is the regression guard for the helper move)
- `bunx tsc --noEmit -p apps/api` → clean. eslint + prettier clean on all four files.

The race itself is not reproduced end-to-end: tier-0 PGlite is a single
connection, so there is no second writer to interleave with.

## Notes for the orchestrator

- Stacked on `fix/issue-1334`; no overlap with #1333 (`space-members` removal).
- No migration, no env var, no OpenAPI change — nothing to order at deploy.
- The pending-invitation half of the pre-count has no FK behind it, so an
  invitation created concurrently with a delete still strands its assignment.
  That is a different mechanism from the 500 this issue reports and is left
  alone deliberately.
- An outside contributor opened #1374 against this issue; this PR does not use it.

### DIFF
--- a/apps/api/src/lib/db-helpers.ts
+++ b/apps/api/src/lib/db-helpers.ts
@@ -117,6 +117,24 @@ export function isInvalidTextRepresentation(err: unknown): boolean {
 }
 
 /**
+ * True when a DB error is a referential-integrity violation on write:
+ * PostgreSQL raises `23503` (foreign_key_violation), while PGlite (tier 0)
+ * surfaces an `ON DELETE RESTRICT` as `23001` (restrict_violation). Both mean
+ * "rows still reference this one". Walks the `cause` chain for the same reason
+ * {@link isUniqueViolation} does.
+ */
+export function isForeignKeyViolation(err: unknown): boolean {
+  let current: unknown = err;
+  for (let depth = 0; current != null && depth < 5; depth++) {
+    if (typeof current !== "object") break;
+    const code = (current as { code?: unknown }).code;
+    if (code === "23503" || code === "23001") return true;
+    current = (current as { cause?: unknown }).cause;
+  }
+  return false;
+}
+
+/**
  * True when a DB error is Postgres `23505` (unique_violation). Walks the
  * `cause` chain for the same reason {@link isInvalidTextRepresentation} does:
  * Drizzle wraps the driver error in a `DrizzleQueryError` whose own `code` is
--- a/apps/api/src/routes/model-provider-credentials.ts
+++ b/apps/api/src/routes/model-provider-credentials.ts
@@ -35,6 +35,7 @@ import type { ProviderRegistryEntry, ProviderRegistryModelEntry } from "@appstra
 import type { ModelProviderDefinition } from "@appstrate/core/module";
 import { testModelConfig } from "../services/org-models.ts";
 import { logger } from "../lib/logger.ts";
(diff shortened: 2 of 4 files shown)


pierrecabriere:

PR #1413, opened 2026-09-10
fix(models): rebind a failed single model add to the credential it minted

### LINKED ISSUE #1355: web(models): retrying a failed single model add creates a second credential and orphans the first

pierrecabriere:
Closes #1355

## Root cause

`submitOne` (`apps/web/src/hooks/use-models.ts:215-235` before this change) returned `{ failedModelIds: [data.modelId] }` with no `credentialId`, while `submitBatch` returned `{ failedModelIds, credentialId }` (line 205). The modal only stores what it is given (`model-form-modal.tsx:348`, batch path only), and the single path built its payload with `selectedCredentialId: selectedCredential?.id ?? null` (line 307).

So after "inline key minted → model refused", the re-submit still carried `newCredential` and `credentialId: ""`. `resolveCredentialBinding` (`lib/model-form-payload.ts:148`) therefore minted again: a second credential holding the same secret, the first referenced by no model. `POST /api/model-provider-credentials` always inserts — there is no dedupe by key.

## Fix

The retry contract is now the same on both paths: **every refusal that happens after the key was minted names the credential it minted**, and only a refusal *of the key itself* names none.

That contract was not assertable while the flow lived inside the hook, closed over three `mutateAsync` calls — `apps/web` has no DOM in tests (`renderToStaticMarkup` only), so an async submit cannot be driven there. It moves to `apps/web/src/lib/model-form-submit.ts` with the writes injected (`ModelFormWrites`), which is also what `apps/web/CLAUDE.md` asks for: "any code that builds a request body the SPA sends is a pure function under `lib/` with a test that pins the emitted payload". The hook keeps the pending flag and the three adapters (~20 lines). Behaviour of the batch path is unchanged — `bindCredential` returning `null` replaces the outer try/catch that used to swallow the key refusal.

Modal side:
- the single path builds its payload with `selectedCredential?.id ?? createdCredentialId` and stores `outcome.credentialId` on failure, exactly as the batch path does;
- `createdCredentialId` no longer dies with the model step. `dropListing()` cleared it, and `dropListing` is wired to `resetModelStep` — so switching between the list and the manual field (same provider, same key) stranded the minted credential and the next submit minted a second one, the reported bug by another route. It is now dropped only when the key material changes: the typed key or base URL (`dropEndpointBinding`, which is what those two `register` `onChange`s used to reach `dropListing` for), the provider (`switchProvider`), or picking/clearing a saved credential (`bindCredential`);
- the "a key will be created automatically" hint hides once a key has been minted — after that a retry rebinds, it does not create.

## Tests

`apps/web/src/lib/test/model-form-submit.test.ts` (new, 8 cases): the minted credential is reported on a refused create and on a refused edit; the retry built from that outcome creates **no** second credential and binds to the first; a refused key names none and never attempts the model; batch partial/total failure and success pin the existing behaviour.

```
cd apps/web && bun test src/lib/test/model-form-submit.test.ts   → 8 pass, 0 fail
cd apps/web && bun test                                          → 886 pass, 0 fail (87 files)
bunx tsc --noEmit -p apps/web                                    → clean
bunx eslint <4 touched files>                                    → clean
bunx knip --workspace apps/web                                   → clean
```

Negative control: reverting only `return { failedModelIds: [data.modelId], credentialId }` to the old shape turns 2 of the 8 red.

## Notes for the orchestrator

- Based on `fix/issue-1354`; PR targets that branch. No overlap with its diff (`lib/model-form-payload.ts`, `lib/model-source.ts`) — this touches `hooks/use-models.ts`, `components/model-form-modal.tsx` and a new `lib/model-form-submit.ts`.
- No migration, no env var, no OpenAPI change (frontend only) — so `verify:openapi` / `verify:api-types` were not needed.
- Sibling #1358 (duplicate `org_models` rows) is untouched.
- **Open decision, deliberately not implemented**: the issue also suggests cleaning up a credential that ends up referenced by no model. Not done — deleting a stored API key on the user's behalf is destructive and the credential is a legitimate saved key visible in the credentials list. This PR removes the mechanism that *created* the orphans; reaping pre-existing ones is a separate product call.

### DIFF
--- a/apps/web/src/components/model-form-modal.tsx
+++ b/apps/web/src/components/model-form-modal.tsx
@@ -210,7 +210,11 @@ function ModelForm({
   const [picked, setPicked] = useState<ModelPickRow[]>([]);
   /** Ids a batch could not create — re-offered instead of silently dropped. */
   const [failedModelIds, setFailedModelIds] = useState<string[]>([]);
-  /** The credential a partially failed batch already created: a retry binds to it. */
+  /**
+   * The credential a refused submission already minted from the typed key. It
+   * outlives the model step: a retry binds to it instead of minting a second
+   * one against the same secret and orphaning the first.
+   */
   const [createdCredentialId, setCreatedCredentialId] = useState<string | null>(null);
   const [modelMode, setModelMode] = useState<"list" | "manual" | null>(null);
 
@@ -231,7 +235,15 @@ function ModelForm({
   const dropListing = () => {
     setPicked([]);
     setFailedModelIds([]);
+  };
+
+  /**
+   * The endpoint and the typed key are what a credential is minted from:
+   * editing either strands whatever a refused submission already minted.
+   */
+  const dropEndpointBinding = () => {
     setCreatedCredentialId(null);
+    dropListing();
   };
 
   const resetModelStep = () => {
@@ -249,11 +261,12 @@ function ModelForm({
     dropListing();
   };
 
-  /** A typed key never coexists with a saved credential. */
+  /** A typed key never coexists with a saved credential — nor with what it minted. */
   const bindCredential = (id: string) => {
     setValue("credentialId", id);
     setValue("inlineApiKey", "");
     clearErrors("credentialId");
+    setCreatedCredentialId(null);
   };
 
   /**
@@ -266,6 +279,8 @@ function ModelForm({
     clearErrors();
     setValue("credentialId", "");
     if (!keepTypedKey) setValue("inlineApiKey", "");
+    // A minted credential is pinned to the provider it was created for.
+    setCreatedCredentialId(null);
     const provider = getProviderById(id, registry);
     if (provider) {
       setValue("apiShape", provider.apiShape);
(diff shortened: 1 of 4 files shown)


pierrecabriere:

PR #1416, opened 2026-09-10
fix(module-ee): repair an org's billing account and its debt in one transaction

### LINKED ISSUE #1329: ee(billing): repair:account is not transactional, and its guard makes an interrupted repair permanent

pierrecabriere:
Closes #1329

## Root cause

`packages/module-ee/src/billing/repair-account.ts` ran the repair as two
independent commits: `provisionBillingAccount(...)` (which opened its own
transaction) and then an `UPDATE ee_billing_accounts SET credits_used =
SUM(ee_usage_records.cost_credits)`. An interruption between them (Ctrl-C, a
pool blip, a pod eviction) left the account provisioned with the debt
unapplied, and the existence guard at the top of the function then returned
`already_provisioned` on every re-run — the org's whole orphaned debt written
off silently, by the one tool meant to recover it.

## Fix

The guard, the provision and the debt now run in **one** transaction, so a
repair either lands whole or leaves no trace. To compose them,
`provisionBillingAccount` takes the caller's `EeTx` instead of opening a
transaction of its own; `onOrgCreate` opens one, so the free-tier claim and the
account insert still commit together exactly as before. That is the whole
refactor — one signature, two call sites, no duplicated claim logic.

**The guard deliberately stays keyed on the account's existence**, not on a
`credits_used` shortfall as the issue suggested. The two states are
indistinguishable: `customer.subscription.deleted` downgrades an account to
free with `credits_used = 0`, `period_end = NULL` and
`stripe_subscription_id = NULL` (`src/stripe/webhooks.ts:560-580`) while its
historical usage records survive — byte for byte the signature of "provisioned,
debt not applied". A shortfall-keyed guard would re-charge such an org for
everything it ever spent. With the transaction above, existence is not just the
only safe test but a sufficient one: the partial state is unreachable. The
reasoning is written into the module docblock so it is not re-litigated.

## Tests

`packages/module-ee/test/integration/services/repair-account.test.ts` — one new
case: an interrupted repair (a throwing `onBeforeCommit` seam, the same idiom
`sweepLedgerBatch` already uses) leaves **no** account behind, and a re-run
repairs fully. The re-run's `creditQuota` of 5000 also pins that the rolled-back
attempt did not burn the email's non-renewable free-tier claim.

Negative control: reverting only the atomicity (provisioning in a nested
transaction of its own) makes the new case fail with an account at
`credits_used: 0, credit_quota: 5000` — exactly the state the issue describes.

```
bun test packages/module-ee/test                     → 445 pass, 26 skip, 0 fail
bun test .../repair-account.test.ts                  → 6 pass, 0 fail
bunx tsc --noEmit -p packages/module-ee              → clean
bunx eslint + prettier on the 3 touched files        → clean
```

## Notes for the orchestrator

- No migration, no env var, no OpenAPI surface touched. `repair:account`'s CLI
  contract and its `RepairOutcome` union are unchanged.
- Overlap with siblings: none. The EE sweeper chain (#1326/#1328/#1330/#1370)
  never touches `repair-account.ts` or `post-signup.ts`; #1327 is catalog
  re-pricing.
- **Test-environment caveat, not a code caveat**: the Docker daemon is down on
  this host (OrbStack's VM panics — its macOS Group Container directory is
  missing/unwritable), so the shared compose Postgres was unavailable. The
  suites above were run against a locally started PostgreSQL 17.6 on :5433 and
  Redis on :6380 matching `docker-compose.test.yml`, under the fleet's
  `pg-test-lock.sh`. Worth a re-run on the consolidated branch once Docker is
  back.
- Residual, pre-existing and out of scope: a sweep pass committing a usage
  record for the same org in the window between the repair's `SUM` read and its
  commit would still leave that increment un-applied, since the repair's guard
  refuses afterwards. Closing it needs the sweeper to take a lock the repair can
  also take — a change to the sweeper, not to this script.

### DIFF
--- a/packages/module-ee/src/billing/repair-account.ts
+++ b/packages/module-ee/src/billing/repair-account.ts
@@ -36,6 +36,23 @@
  * reconstruction would be WRONG on an existing account (a Stripe renewal resets
  * `credits_used` to 0 while historical usage records remain), which is why this
  * refuses to touch an org that already has one.
+ *
+ * ALL OR NOTHING. The guard, the provision and the debt run in ONE transaction.
+ * Split across two commits, an interruption between them (Ctrl-C, a pool blip, a
+ * pod eviction) left the account provisioned and the debt unapplied — and the
+ * guard below then reported `already_provisioned` forever, writing the org's
+ * entire debt off silently while refusing to touch it again. The atomic form
+ * makes that state unreachable: a repair either lands whole or leaves no trace,
+ * so re-running it after any failure is always correct.
+ *
+ * WHY THE GUARD STAYS "DOES A ROW EXIST", rather than "does `credits_used`
+ * already cover the records". It cannot tell the two apart: a subscription
+ * cancellation downgrades the account to free with `credits_used = 0`,
+ * `period_end = NULL` and `stripe_subscription_id = NULL` while the historical
+ * usage records survive — byte for byte the signature of "provisioned, debt not
+ * applied". A guard keyed on the shortfall would re-charge that org for
+ * everything it ever spent. Existence is the only safe test, and with the
+ * transaction above it is also a sufficient one.
  */
 
 import { eq, sql } from "drizzle-orm";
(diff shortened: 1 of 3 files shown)


pierrecabriere:

PR #1417, opened 2026-09-10
fix(auth): renew the distributed lock while the critical section runs

### LINKED ISSUE #1366: auth: the distributed lock has no renewal, so a critical section past its 45s TTL still double-spends the refresh token

pierrecabriere:
Closes #1366

## Root cause

`apps/api/src/lib/distributed-lock.ts` acquired with `SET … EX ttl NX` and never renewed. A critical section slower than the 45 s TTL — the upstream exchange plus `reReadFreshness`'s DB read under pressure — lost its lock mid-flight; the waiter (polling at 100 ms) then acquired and ran `doRefresh()` concurrently. Both spend the rotating `refresh_token`, the loser writes a dead one, and a valid credential is flagged `needsReconnection`.

Renewal alone would not have closed it: the waiter's `acquireTimeoutMs` was derived as `ttl + 5 s` precisely because the holder's lease "definitively expired" at the TTL. With renewal that premise is false — the waiter would have given up at 50 s and run **unlocked**, double-spending anyway.

## Fix

`distributed-lock.ts` (+96/−22):

- `startLeaseWatchdog` renews the lease every `ttl/3` through a compare-and-PEXPIRE Lua script (same `lockId` guard as the release script, so a lapsed holder can never resurrect the new owner's key). A transient Redis error retries on the next tick — two more renewals fit inside the remaining lease; a renewal that returns 0 means the lease is already gone and the watchdog stops with a warning.
- Renewal is **capped** by a new `maxHoldSeconds` option. Past it the holder is not slow, it is wedged: renewal stops, the lease lapses within `ttlSeconds`, and a hung `fn` can never deadlock the key. Redisson-style unbounded renewal was rejected for exactly that reason.
- `acquireTimeoutMs` is **removed from the options** and derived inside the lock as `(maxHoldSeconds + ttlSeconds) * 1000 + 5 s` — the point at which the holder's lease is guaranteed gone. The invariant "a waiter gives up only once the holder is presumed dead" is now structural instead of a formula duplicated in a caller's comment.

`deduped-refresh.ts` (+21/−13, comments and constants only): `REFRESH_LOCK_MAX_HOLD_SECONDS = 90` (two full 30 s exchange timeouts past the 45 s lease) replaces `REFRESH_LOCK_ACQUIRE_TIMEOUT_MS`.

**Behaviour change worth knowing:** a caller queued behind a genuinely wedged holder now waits up to ~2 min 20 s (was ~50 s) before proceeding unlocked. That is the deliberate trade: waiting yields a slow request, proceeding early yields a broken credential the user must reconnect by hand. In the normal case (holder releases in well under a second) nothing changes.

## Tests

`apps/api/test/unit/distributed-lock.test.ts` (new, `describeRequiresRedis` — the Lua and key expiry are the behaviour under test, so it skips on tier 0 and runs in CI tier 3):

- *renews the lease so a critical section outliving the TTL keeps the lock* — `ttl: 1 s`, holder holds 2.5 s; the waiter must enter only after the holder exits.
- *stops renewing past `maxHoldSeconds`* — `ttl: 1 s`, `maxHold: 1 s`; the key is gone while `fn` is still running.

Commands run:

- Real Redis (local `redis-server --port 6399`, run from `apps/api` so the tier-3 preload's Docker Postgres is not needed): `REDIS_URL=redis://127.0.0.1:6399 bun test test/unit/distributed-lock.test.ts` → **2 pass, 0 fail**.
- Negative control: with the watchdog stubbed out, test 1 fails with `["holder:enter", "waiter:enter", "holder:exit"]` — the concurrent critical section the issue describes. Restored and re-run green.
- `set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/unit/deduped-refresh.test.ts apps/api/test/unit/distributed-lock.test.ts` → **9 pass, 3 skip, 0 fail** (the Redis suite skips on tier 0 as designed; #1365's suite unaffected).
- `bunx tsc --noEmit -p apps/api` → clean. `bunx eslint` + `bunx prettier --write` on the three touched files → clean.

## Notes for the orchestrator

- Based on `fix/issue-1365`; the only shared file is `deduped-refresh.ts`, and this PR touches just its two lock constants plus their doc comments — no overlap with #1365's flight-outcome logic.
- No migration, no env var, no OpenAPI change. Nothing to order at deploy: the lock key format and TTL are unchanged, so an old and a new instance interoperate (the old one simply does not renew).
- `withRedisLock` has exactly one caller (`dedupedRefresh`), so dropping `acquireTimeoutMs` from its options breaks nothing else.
- The new test only executes where a real Redis is configured — it runs in CI, and is a no-op in a tier-0 gate.

### DIFF
--- a/apps/api/src/lib/deduped-refresh.ts
+++ b/apps/api/src/lib/deduped-refresh.ts
@@ -32,20 +32,31 @@
 
 import { withRedisLock } from "./distributed-lock.ts";
 
-/** Distributed-lock TTL in seconds — sized as `30s network timeout` + slack. */
+/**
+ * Distributed-lock lease in seconds — sized as `30s network timeout` + slack.
+ * A watchdog renews it while the exchange runs, so this is not a cap on the
+ * critical section: it is how long a CRASHED holder's lock lingers before a
+ * peer can reclaim it.
+ */
 const REFRESH_LOCK_TTL_SECONDS = 45;
 /**
- * How long to wait for the distributed lock before proceeding unlocked.
- * Derived from the TTL so the two cannot drift apart: a waiter gives up only
- * once the holder's lock has definitively expired (holder presumed dead),
- * never at the holder's worst-case exchange duration. Trade-off: a sidecar
- * caller queued behind a wedged holder waits up to ~50 s before proceeding,
- * plus the 30 s exchange timeout in `@appstrate/connect`. A forced caller that
- * arrives while a proactive flight is already exchanging pays that once, not
- * once per hop — it adopts that exchange's token rather than queueing a second
- * one behind it (see {@link dedupedRefresh}).
+ * Hard cap on how long the lock watchdog keeps renewing. Two full exchange
+ * timeouts past the lease: a holder still running then is not slow, it is
+ * wedged, and must stop renewing so the credential's refresh is not deadlocked
+ * for every other instance.
+ *
+ * `withRedisLock` derives the waiter's give-up point from this (the waiter
+ * polls until the holder's lease is guaranteed gone), so the two cannot drift
+ * apart. Trade-off: a caller queued behind a wedged holder waits up to
+ * ~2.5 min before proceeding unlocked, plus the 30 s exchange timeout in
+ * `@appstrate/connect`. Waiting is the cheaper failure — proceeding early
+ * double-spends the rotating `refresh_token` and flags a valid credential
+ * `needsReconnection`. A forced caller that arrives while a proactive flight
+ * is already exchanging pays that once, not once per hop — it adopts that
+ * exchange's token rather than queueing a second one behind it (see
+ * {@link dedupedRefresh}).
  */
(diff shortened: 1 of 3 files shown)


pierrecabriere:

PR #1418, opened 2026-09-10
fix(oidc): make runs:read-all requestable without granting it to end-users

### LINKED ISSUE #1372: oidc: runs:read-all is unrequestable by any OAuth client, silently narrowing admins over MCP

pierrecabriere:
Closes #1372

## Root cause

`apps/api/src/modules/oidc/auth/scopes.ts:84` built `APPSTRATE_BUILTIN_SCOPES` as `OIDC_IDENTITY_SCOPES + OIDC_ALLOWED_SCOPES`, and that one array served two different jobs: the **end-user allowlist** consulted by `claims.ts:44` and the **provider's global vocabulary** handed to `oauthProvider({ scopes })` (`auth/plugins.ts:225`) and to `assertValidScopes` (`services/oauth-admin.ts:111`). Keeping `runs:read-all` out of the end-user allowlist is correct — an embedding app acting for an end-user must never see a colleague's run — but it also made the scope unregisterable by *any* client. A `dashboard_user` token was therefore permanently capped at `runs:read` (its own runs) even for an org admin, although `scopesToPermissions(…, "dashboard_user", role)` would have granted it: `runs:read-all` is inside `orgPermissions(role) ∪ knownSpaceLevelPermissions()` (verified).

Note the issue's illustrative scenario is not the live one: a Claude Code MCP client registers via DCR/CIMD → `level: "instance"` → `actor_type: "user"` → `resolveInstanceUser` sets `deferOrgResolution` with no scope ceiling (`auth/strategy.ts:153`), so it is not narrowed. The narrowing hits **org-level (admin-registered) clients**, which are the ones that mint `dashboard_user` tokens.

## Fix

- `OIDC_DASHBOARD_ONLY_SCOPES = { "runs:read-all" }` — requestable, never grantable to an end-user. It joins `APPSTRATE_BUILTIN_SCOPES` (→ discovery `scopes_supported`, `GET /api/oauth/scopes`, `assertValidScopes`); `OIDC_ALLOWED_SCOPES` and therefore `claims.ts` are untouched, so an end-user token that asks for it still has it dropped.
- `getSelfServiceScopes()` is untouched: a DCR/CIMD registrant still gets `invalid_scope` for it.
- `getEndUserScopeSet()` + level-aware `invalidScopesIn(scopes, level)` / `assertValidScopes(scopes, level)`. A **space-level** client only ever mints end-user tokens, so registering a dashboard-only scope there would be silently dropped at every mint forever — the same silent-narrowing failure this PR fixes, relocated. It is now refused loudly at registration. `updateClient` reads the stored row once for both level-dependent checks (scopes + signup policy) instead of not at all / twice.
- Consent-screen FR description + `oauthClients.scopeLabels.runs:read-all` in fr/en, without which `test/unit/oauth-scope-labels.test.ts` fails and the user would be asked to grant a raw scope string.
- OpenAPI `scopes` description now states the space-level narrowing; `apps/web/src/api/schema.d.ts` regenerated.

## Tests

- `apps/api/src/modules/oidc/test/unit/scopes-to-permissions.test.ts` — new `dashboard-only scopes` block: present in the vocabulary, absent from the end-user allowlist and the self-service ceiling, granted on a `dashboard_user`/`admin` token, still dropped on an `end_user` token.
- `apps/api/src/modules/oidc/test/unit/client-scope-vocabulary.test.ts` (new) — the level matrix for `invalidScopesIn`.

Commands, from the worktree root:

- `set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/src/modules/oidc/test/unit apps/api/test/unit/oauth-scope-labels.test.ts` → **133 pass / 0 fail**
- `… TEST_TIER=0 bun test apps/api/src/modules/oidc/test/integration` → **447 pass / 0 fail**
- `bunx tsc --noEmit -p apps/api` → clean; `-p apps/web` → clean
- `bun run verify:openapi` → ALL CHECKS PASSED; `bun run generate:api` + `bun run verify:api-types` → up to date; `bun run detect:breaking` → 0 breaking, 0 non-breaking
- `bunx eslint` + `bunx prettier --write` on every touched file → clean

Pre-existing, unrelated to this diff: `apps/web/src/locales/test/locale-keys.test.ts › declared keys › are all still referenced by the source` times out at 5 s on this machine. Verified by reverting only the two locale JSON edits and re-running — it fails identically.

## Notes for the orchestrator

- **No migration, no env var, no re-consent, no backfill.** Widening the vocabulary does not touch a single stored row: existing `oauth_clients.scopes` arrays and existing consent grants keep exactly what they had. An admin must explicitly add `runs:read-all` to a client for anything to change, and the consent screen then lists it (a `skipConsent` first-party client would take it without a prompt — but only after that explicit admin edit). This is the check the #1291 memory note asks for: nothing here re-matches or re-derives an existing grant, and script `0008` is not involved.
- **Deploy ordering**: none. Platform-only change, no runtime image coupling.
- **Sibling #1339** (MCP `run_and_wait` hard-codes `runs:read`): this PR is the half that makes `runs:read-all` *askable at all*. If #1339's fix wants an MCP caller to be able to hold the space-wide read, it needs this merged first — but the two diffs do not overlap in any file (`modules/mcp/tools.ts` is untouched here).
- Operator-visible behaviour change worth a release note: `GET /api/oauth/scopes` (and discovery `scopes_supported`) now advertise `runs:read-all`, so it appears as a tickable box in the dashboard's OAuth-client editor. The picker is not level-filtered, so ticking it on a *space*-level client returns a 400 naming the scope and the level rather than silently ignoring it.


pierrecabriere:

PR #1420, opened 2026-09-10
fix(db): bound NOTIFY payload text in bytes, not characters

### LINKED ISSUE #1340: db: the run_log_insert NOTIFY budget is counted in characters while pg_notify's limit is in bytes
## Summary

The `run_log_insert` NOTIFY payload budget is enforced in **characters** while `pg_notify`'s 8000 limit is in **bytes**. A 2000-character non-ASCII log line therefore aborts the `run_logs` INSERT instead of dropping a frame.

## Mechanism

`packages/db/src/notify.ts` — the block #1307 touched caps with `LEFT(NEW.message, 2000)` and `LEFT(…, 5000)` for the data, with a comment asserting the two "stay under that ceiling".

`LEFT` counts characters. A 4-byte code point makes 2000 characters 8000 bytes.

## Executed (PGlite)

```sql
SELECT pg_notify('run_log_insert', json_build_object('message', repeat('😀', 2000))::text);
-- ERROR: payload string too long
-- (1500 → OK)
```

The trigger raises *inside* the `run_logs` INSERT, so the write is aborted — an agent emitting an emoji-heavy or CJK log line loses the row, not just the live frame.

The cap itself predates #1307 (which loosened the total from 8000 to ~7080), but that commit is the one asserting the budget is safe.

## Proposed fix

Cap on `octet_length` (e.g. `left(NEW.message, …)` chosen so the encoded payload fits, or `convert_to(...)` truncation), and add a non-ASCII case to the notify-trigger tests.

pierrecabriere:
Closes #1340

## Root cause

`packages/db/src/notify.ts` capped the two free-form strings it interpolates into
NOTIFY payloads with `LEFT(..., 2000)` — `run_logs.message` (old line 137) and
`runs.error` (old line 93). `LEFT()` counts **characters**; `pg_notify`'s ceiling
is 8 000 **bytes**, so 2 000 emoji are 8 000 bytes and the cap bounded nothing.

The raise happens *inside* the trigger, so the consequence is not a lost frame:
an emoji- or CJK-heavy agent log line aborted the `run_logs` INSERT (the row is
gone), and the identical cap on `runs.error` rolls back the finalize UPDATE and
leaves the run orphaned in `running`.

Two columns the budget comment claimed to account for were never bounded at all:
`run_logs.type` and `run_logs.event` are free-form by design (`schema/runs.ts:432`
— "progress/event kinds modules invent freely").

## Fix

A new SQL function `notify_text_budget(_value text, _max_bytes int)`, installed
by `createNotifyTriggers()` alongside the trigger functions, trims a value until
its **JSON encoding** fits the byte budget, then rescales by the measured
bytes-per-character and re-checks. Measuring the encoding rather than
`octet_length(_value)` is deliberate: one control character costs six bytes once
escaped (`\u0001`), which the raw byte length does not see. `left()` cuts on
code points, so a multibyte value is never sliced mid-character.

The loop is strictly decreasing whenever it runs (the ratio is < 1), and exits on
`''`, so it terminates for every input including NULL and a non-positive budget —
all four exercised below.

Applied to `runs.error` (2 000 B), `run_logs.message` (2 000 B), `run_logs.event`
(400 B) and `run_logs.type` (100 B). The `data` cap was already byte-based and is
unchanged, so the existing 4 900-under / 6 000-over test still pins it.

**No migration.** These function bodies are a boot-time artifact — `CREATE OR
REPLACE` from `bootBackground()` (`apps/api/src/lib/boot.ts:248`) is the only
installer, and `0053_applications_to_spaces.sql` documents at length why the
drizzle chain must not own them. Copying the bodies into 0061 would create a
second, divergent source of truth. The reserved platform index **0061 is
therefore unused and free for another agent.**

## Tests

`apps/api/test/integration/services/notify-triggers.test.ts` — two new cases:
a 2 000-emoji `message` must keep its INSERT and arrive trimmed on whole code
points (with a 400-char CJK message alongside proving an in-budget multibyte
value is still delivered verbatim), and a 2 000-emoji `runs.error` must not roll
back the finalize UPDATE.

Commands run:

- `bunx tsc --noEmit -p packages/db` -> clean
- `bunx tsc --noEmit -p apps/api` -> clean
- `bunx eslint` + `bunx prettier --check` on both touched files -> clean
- `bun run verify:no-migration-dml` -> green (68 files, 2 trees)
- **The integration suite could not be executed here: the Docker daemon
  (OrbStack) is down on this machine and would not start, so the shared
  Postgres the suite needs is unavailable.** Instead the SQL was validated
  end-to-end on PGlite in a scratch harness that installs the helper plus the
  real `notify_run_log_insert` body verbatim:
  - negative control — the pre-fix expression
    `LEFT(repeat('emoji', 4000), 2000)` raises `payload string too long`;
  - helper termination and output on 8 inputs: 2 000 emoji -> 499 chars /
    1 996 B, 2 000 CJK -> 666 chars / 1 998 B, 5 000 ASCII -> 1 998 B,
    3 000 control chars -> 333 chars / 2 000 B *encoded*, 400 CJK -> untouched,
    `''`, `NULL`, and budget `0`;
  - worst case through a real trigger + `LISTEN`: multibyte `type`, `event` and
    `message` with `data` at the cap -> INSERT survives, payload **7 663 bytes**
    (ceiling 7 999), `message` 1 996 B and all whole emoji;
  - a 400-char CJK message round-trips byte-identical.

  The new bun:test assertions are the same predicates against the same inputs, so
  they should be green once the suite runs on the consolidated branch.

## Notes for the orchestrator

- **Reserved migration 0061 was not used** — reassign it.
- Overlap with siblings: none. The INFO-tier rolling-deploy asymmetry on this
  trigger was left alone as instructed; it is unchanged by this PR (a pod running
  the old body simply keeps the old caps until the new boot replaces them, which
  is the pre-existing behaviour).
- Scope note: the issue names only `run_log_insert`, but `notify_run_change`'s
  `LEFT(NEW.error, 2000)` is the same defect with a worse blast radius, and the
  same helper covers it in one line. Bounding `type` / `event` is what makes the
  budget comment true rather than aspirational.
- No env var, no OpenAPI change, no deploy ordering constraint: the helper is
  created before the trigger functions in the same `createNotifyTriggers()` call,
  so it always exists before a new body can reference it.

### DIFF
--- a/apps/api/test/integration/services/notify-triggers.test.ts
+++ b/apps/api/test/integration/services/notify-triggers.test.ts
@@ -250,6 +250,73 @@ describe("NOTIFY triggers (regression)", () => {
     expect(received[1]).toMatchObject({ message: "large data", data: "[payload too large]" });
   });
 
+  // `left()` counts CHARACTERS while pg_notify's ceiling is BYTES: a 2 000-
+  // character emoji log line is 8 000 bytes on its own. The trigger used to cap
+  // with `LEFT(NEW.message, 2000)`, so such a line made pg_notify raise INSIDE
+  // the trigger and the agent lost the run_logs ROW, not just the live frame.
+  it("notify_run_log_insert keeps a multibyte log line and trims it by bytes", async () => {
+    const run = await seedRun({
+      packageId: "@notifyorg/trigger-agent",
+      orgId: ctx.orgId,
+      spaceId: ctx.defaultSpaceId,
+      userId: ctx.user.id,
+      status: "running",
+    });
+
+    const received: Array<Record<string, unknown>> = [];
+    await listenClient.listen("run_log_insert", (raw) => {
+      try {
+        const payload = JSON.parse(raw) as Record<string, unknown>;
+        if (payload.run_id === run.id) received.push(payload);
+      } catch {
+        /* ignore */
+      }
+    });
+
+    // 400 × 3 bytes = 1 200 bytes — inside the 2 000-byte message budget.
+    const fits = "漢".repeat(400);
+    // 2 000 × 4 bytes = 8 000 bytes — over the whole NOTIFY ceiling by itself.
+    const overflows = "😀".repeat(2_000);
+
+    // Both `seedRunLog` calls RETURNING the row is half the assertion: a
+    // trigger that raises aborts the INSERT, so the second one used to throw
+    // `payload string too long` here.
+    await seedRunLog({ runId: run.id, orgId: ctx.orgId, event: "fits", message: fits });
+    await seedRunLog({ runId: run.id, orgId: ctx.orgId, event: "overflows", message: overflows });
+    for (let i = 0; i < 40 && received.length < 2; i++) {
+      await new Promise((r) => setTimeout(r, 25));
+    }
+
+    expect(received).toHaveLength(2);
+    // Under budget → delivered verbatim; the budget is bytes, not a blanket
+    // character cap that would mangle CJK that comfortably fits.
+    expect(received[0]).toMatchObject({ event: "fits", message: fits });
+
+    // Over budget → trimmed on whole code points, never mid-character.
+    const trimmed = received[1]!.message as string;
+    const bytes = new TextEncoder().encode(trimmed).length;
+    expect(bytes).toBeLessThanOrEqual(2_000);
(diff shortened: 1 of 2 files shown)


pierrecabriere:

PR #1422, opened 2026-09-10
fix(rbac): bound the target of a space-member removal or role change

### LINKED ISSUE #1333: security(rbac): nothing bounds the target of a space-member removal or role change

pierrecabriere:
Closes #1333

## Root cause

`assertCanGrantSpaceRole` (`apps/api/src/lib/space-role-policy.ts:24`) bounds only the role being **granted**.

- `DELETE /api/spaces/:id/members/:userId` (`apps/api/src/routes/spaces.ts:534`) passed it the **post-removal implicit** role. In a `closed` or `private` space that is `null`, `spacePermissions(null)` is empty and the check passes trivially — so `{space-members:remove}` alone ejected a preset-`admin` member of the same space. In an `open` space it was refused only by accident, because the open default happens to grant something.
- `PATCH /api/spaces/:id/members/:userId` → `saveSpaceMember(..., requireExisting: true)` had **no** target check in any visibility, so `{space-members:change-role}` alone demoted a space admin.

## Fix

One new predicate, `assertCanManageSpaceMember`, the mirror of the grant check on the other side of the write: **a member may only be removed or re-roled by someone who could have granted them the role they currently hold.** It reuses the same `missingPermission` / `spacePermissions` pair, so there is one subset rule, applied at both ends.

- DELETE reads the explicit row before dropping it and asserts, then keeps the existing implicit-grant check.
- PATCH asserts inside the write transaction. The existence probe that was already there is replaced by `loadSpaceMember(spaceId, userId, tx)` — one query instead of two, and the read now happens on the same connection as the write. `loadSpaceMember` gained an optional executor (`Pick<typeof db, "select">`, the precedent `deleteSpaceMembershipsInOrg` already uses) and `apps/api/test/unit/lib/view-as-callsites.test.ts` gained the matching allowlist entry.

Consequences, all deliberate: self-removal and self-changes stay permitted (an actor always holds their own role); owners/admins hold preset `admin` everywhere so the bound never binds them; no last-admin rule is introduced — the spec states none. The rule was unstated, so it is now written down in `docs/architecture/RBAC_PERMISSIONS_SPEC.md` §6.4 and in the two OpenAPI descriptions.

One existing test changed meaning: *"removal remains permitted when it grants no access in a closed space"* used a remove-only actor against a `viewer` target, which is exactly the hole. Its actor now also holds `viewer` — the point it pins (a closed-space removal grants nothing, so it is allowed) is unchanged.

## Tests

`apps/api/test/integration/routes/space-role-delegation.test.ts` — new `a narrow bundle cannot eject or demote a member it could not have granted` (closed space, so the grant-side check is provably vacuous and only the target bound can refuse; the changer holds `viewer` and demotes *to* `viewer`, so the grant check is cleared too), plus a self-removal assertion on the amended closed-space test.

```
  apps/api/test/integration/routes/space-role-delegation.test.ts \
  apps/api/test/integration/routes/space-membership.test.ts \
  apps/api/test/integration/routes/space-member-management.test.ts \
  apps/api/test/integration/services/space-member-concurrency.test.ts \
  apps/api/test/integration/routes/roles.test.ts \
  apps/api/test/integration/routes/view-as.test.ts
→ 100 pass, 4 skip, 0 fail
```

Negative control, run twice with each assert disabled in turn: DELETE `403 → 200`, PATCH `403 → 200`. (The first PATCH control I wrote used a `custom_role_id` and was refused by the `custom_roles` feature gate, not by the new bound — the committed test uses a preset for that reason.)

`bunx tsc --noEmit -p apps/api` clean · `bun run verify:openapi` ALL CHECKS PASSED · `bun run generate:api` → `apps/web/src/api/schema.d.ts` committed, `verify:api-types` up to date · `bun run detect:breaking` PASSED (0 breaking) · eslint + prettier clean.

## Notes for the orchestrator

- Base is `fix/issue-1371` (the space-roles chain tail). No migration, no env var, no deploy ordering.
- `apps/api/src/services/space-members.ts` is the file shared with the already-merged #1334; the overlap is limited to the existence-probe block in `saveSpaceMember` — `assignmentColumns` is untouched.
- `apps/web/src/api/schema.d.ts` is regenerated output (two description strings), not hand-edited.


pierrecabriere:

PR #1423, opened 2026-09-10
fix(client-ip): a forwarded chain shorter than TRUST_PROXY no longer names the client's own IP

### LINKED ISSUE #1316: security(env): the TRUST_PROXY production gate can force an IP-spoofable configuration
## Summary

The production boot gate added in #1287 refuses `TRUST_PROXY=false`, inferring "a proxy is in front" from the **URL scheme** — then trusts `X-Forwarded-For` with no proxy-identity check and no validation of the value. On a topology where the origin port is directly reachable, this converts "collapsed but unspoofable" into "spoofable".

Not yet deployed (prod is at `beta.57`, migration 0055). The window opens at the deploy.

## Mechanism

The gate: `packages/env/src/index.ts:961-971`. The resolver: `apps/api/src/lib/client-ip.ts:52-59`:

```ts
const idx = parts.length - hops;
return idx >= 0 ? parts[idx] : parts[0];   // ← fail-open
```

Counting from the right is correct **when a proxy is genuinely in front**: with `TRUST_PROXY=1` and `XFF = "<attacker value>, <real peer>"`, `idx = 1` selects the real peer. The two holes are elsewhere.

Probed against the real `getClientIpFromRequest`:

| TRUST_PROXY | XFF sent | resolved |
|---|---|---|
| `1` | `203.0.113.9` (attacker only) | **`203.0.113.9`** |
| `1` | `203.0.113.9, 198.51.100.7` | `198.51.100.7` ✓ |
| `2` | `203.0.113.9, 198.51.100.7` (1 real hop) | **`203.0.113.9`** |
| `3` | `203.0.113.9` | **`203.0.113.9`** (`parts[0]`) |
| `1` | `X-Real-IP: 203.0.113.99` only | **`203.0.113.99`** |

`x-real-ip` is accepted unconditionally whenever `hops > 0` (`client-ip.ts:64`).

Reachability: `docker-compose.yml:257` publishes `"${PORT:-3000}:3000"` on **all** host interfaces — no `127.0.0.1:` bind — and Docker's iptables rules bypass host ufw/firewalld.

## Blast radius

- `apps/api/src/routes/auth-bootstrap.ts:93` — `rateLimitByIp(5, 60)` is the **stated** control against `AUTH_BOOTSTRAP_TOKEN` brute force (`packages/env/src/index.ts:927-931`). Spoofing the IP removes it: unlimited attempts at claiming instance ownership.
- better-auth's production limiter and session-IP records (`lib/boot.ts:140-141`, `packages/db/src/auth.ts:882`).
- Every `ip:` limiter (`middleware/rate-limit.ts:186`, `:205`, `:270`) and audit records (`services/audit.ts:186`).

Sub-defect: the value is never validated as an IP. `X-Forwarded-For: not-an-ip` propagates to better-auth, whose `getIP` rejects it and drops **all** callers into the shared `NO_TRUSTED_IP_KEY` bucket — a one-attacker rate-limit DoS.

## Proposed fix

1. Distrust the header when `parts.length < hops` — return `undefined` and fall back to the socket address, never `parts[0]`.
2. Validate the resolved value is an IP before stamping it.
3. Make the gate a documented operator decision rather than deriving topology from a URL scheme — a TLS-terminating L4 LB (AWS NLB TLS listener, GCP TCP proxy) adds no XFF and the repo has no PROXY-protocol support, so `https://` in `APP_URL` does not imply an XFF-writing hop.

## Ops note

Independently of the fix: verify port 3000 is not publicly reachable on the Coolify host, and set `TRUST_PROXY=1` in the Coolify `.env` before deploying — the boot gate refuses `false` otherwise.

## Verification

Table above produced by probing the real module. Confirmed.

pierrecabriere:
Closes #1316

## Root cause

`apps/api/src/lib/client-ip.ts:58-59` — `const idx = parts.length - hops; return idx >= 0 ? parts[idx] : parts[0];`

Reading `X-Forwarded-For` from the RIGHT is correct and unspoofable *while every trusted hop appends its entry*: under `TRUST_PROXY=1`, `XFF: <attacker text>, <real peer>` picks the real peer. The hole is the `: parts[0]` fail-open branch. A chain **shorter** than the hop count was written by the caller alone, and clamping to its leftmost entry handed that caller a client IP of its choosing under any `TRUST_PROXY >= 1`.

Blast radius is every per-IP control keyed on the answer: the `rateLimitByIp(5, 60)` at `apps/api/src/routes/auth-bootstrap.ts:93` that `packages/env/src/index.ts:927-931` names as the defence against `AUTH_BOOTSTRAP_TOKEN` brute force, Better Auth's production limiter, and the address stamped on sessions and audit events.

Sub-defect, same function: the value was never checked to be an IP. `XFF: not-an-ip` propagated to Better Auth, whose `getIP` rejects it and drops **all** callers into one shared bucket — a one-caller rate-limit DoS.

## Fix

`resolveFromHeaders` / `pickFromXff` in `apps/api/src/lib/client-ip.ts` — the single reader of these headers in the repo, so the fix lands once.

1. **Fail closed on a short chain.** `parts.length < hops` resolves nothing, and the caller falls back to the socket peer — the one address no header can forge. Refusing the request instead was rejected: the condition is an operator misconfiguration (or a probe), not a client error, and a 400 would take the instance down for everyone the moment a hop count is off by one. The socket peer degrades to the proxy's address, which is unspoofable and exactly what `TRUST_PROXY=false` already means.
2. **A distrusted chain distrusts the whole forwarded set**, `X-Real-IP` included. Falling through to it would move the hole rather than close it: a caller could send a short `XFF` plus a chosen `X-Real-IP`. `X-Real-IP` still answers when no `XFF` is present at all — a proxy that writes the chain writes it on every request, so a caller cannot strip it to reach that branch.
3. **The resolved value must be an IP.** `normalizeForwardedIp` strips an IPv4 `:port` suffix and IPv6 brackets (Azure's front end and some proxies write those; they are spellings of a real address), then validates with `isIP` from `node:net` — already the repo's IP check in `packages/afps-shared/src/ssrf-dns.ts`. Anything else is dropped, which closes the shared-bucket DoS.

No refactor beyond these two functions; no new helper module, no env var, no migration. The trust model is unchanged — `TRUST_PROXY` stays a hop count read from the right.

## Tests

- `apps/api/test/unit/lib/client-ip.test.ts` — the existing case `"clamps to leftmost when not enough entries"` **codified the bug** and is replaced by the fail-closed assertion. Added: short chain at `TRUST_PROXY=2` and `=3`, no `X-Real-IP` fall-through, empty chain, non-IP `XFF` and non-IP `X-Real-IP` rejected, IPv4 port suffix stripped, bare and bracketed IPv6 accepted. Placeholder values (`"client"`, `"proxy1"`) became real addresses since they are validated now.
- `apps/api/test/integration/routes/auth-client-ip.test.ts` — the same asserted where it mattered, at the limiter: a chain shorter than the hop count buys no bucket of its own, and neither does a non-IP entry.

```
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/unit/lib/client-ip.test.ts
  → 17 pass / 0 fail
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/integration/routes/auth-client-ip.test.ts
  → 6 pass / 0 fail
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/unit/rate-limit.test.ts apps/api/test/integration/routes/auth-client-ip.test.ts
  → 25 pass / 0 fail
  apps/api/src/modules/oidc/test/integration/services/oauth-token-rate-limit.test.ts \
  apps/api/src/modules/oidc/test/integration/routes/cli-sessions-admin.test.ts \
  packages/module-observability/test/middleware.test.ts
  → 24 pass / 0 fail
bun run verify:env-docs         → green (106 schema vars, 128 example rows)
bunx eslint + prettier on touched files → clean
```

Every other suite that feeds forwarded headers already used real IP literals, so none needed changing — checked by grepping `x-forwarded-for|x-real-ip` across `apps/` and `packages/`.

## Notes for the orchestrator

- **No migration, no env var, no OpenAPI change.** `docs/ENV.md` prose only (the `TRUST_PROXY` row); `verify:env-docs` re-run and green.
- **Deploy ordering — unchanged, and the existing runbook step still applies.** Production is behind Coolify/Traefik with `TRUST_PROXY=false`, which #1287 already turned into a boot refusal: `TRUST_PROXY=1` must be set in the Coolify `.env` **before** this deploys. That was true before this PR and is not introduced by it. Traefik appends `X-Forwarded-For`, so `1` is the right count and the chain will be long enough — no behavioral change for prod traffic.
- **Open decision, deliberately not taken here — issue item 3.** The gate at `packages/env/src/index.ts:961-971` infers "a proxy is in front" from the `APP_URL` **scheme**, and that inference is not sound: a TLS-terminating L4 load balancer (AWS NLB TLS listener, GCP TCP proxy) terminates `https://` and appends **no** `X-Forwarded-For`, and the repo has no PROXY-protocol support. On such a topology the gate forces `TRUST_PROXY=1` where the truthful value is `0`, and this PR cannot help: with one claimed hop and zero real ones, `parts.length == hops` and the caller's single entry is taken. No code can tell whether a hop appends — only the operator knows. Making it an explicit operator acknowledgement (rather than refusing `false` outright) is a policy change that trades #1287's protection against a silent one-bucket collapse for correctness on L4 topologies. It needs a product call and was left out under "do not redesign the trust model".
- **Ops item, out of scope for code.** `docker-compose.yml:257` publishes `"${PORT:-3000}:3000"` on all host interfaces with no `127.0.0.1:` bind, and Docker's iptables rules bypass host ufw/firewalld. Verify port 3000 is not reachable around the proxy on the Coolify host. Left alone because binding it to loopback would break the topologies that reach the container over the Docker bridge.
- **Overlap: none.** `apps/api/src/lib/client-ip.ts` is the only reader of `X-Forwarded-For` / `X-Real-IP` in the tree; nothing else parses a forwarded chain.


pierrecabriere:

Issue #1337, opened 2026-09-10
security(oidc): per-space SMTP and social config are readable by a guest, for spaces they are not in


pierrecabriere:
## Summary

Per-space SMTP and social-provider configuration is gated on `spaces:read` — an **org-level** permission held by `member` *and* `guest`. A guest, whose whole definition is "no implicit reach into any space", reads the SMTP host/port/username and the social `clientId`/`scopes` for spaces they are not in, `private` ones included.

## Mechanism

`apps/api/src/modules/oidc/routes.ts:664` and `:759` are gated on `requireCorePermission("spaces", "read")`. The routes are mounted at the app root, so no space context middleware runs; the only space check is `assertSpaceBelongsToOrg` (`:659`), which proves org ownership only. `apiKeySpaceScopeGuard` is not in the chain.

`guest`'s org grants (`apps/api/src/lib/permissions.ts:117-127`) include `spaces:read` and deliberately exclude `members:read` — *"no enumerating the org directory"*.

## Impact

Secrets are correctly stripped (`services/smtp.ts:114`, `services/social.ts:86`), so this is metadata disclosure plus a **private-space existence oracle**: `"Space not found"` vs `"SMTP configuration not found"` distinguishes them. An API key bound to space A reads space B's configuration the same way.

Writes are gated on `spaces:write`, so there is no write escalation.

## Proposed fix

Resolve the space context on these routes (`applySpacePermissions`, as the spaces router does) and gate on a space-level permission, so the answer depends on membership rather than on org-level `spaces:read`. Make both 404s identical while you are there.

PR #1424, opened 2026-09-10
fix(api): bound the profile batch lookup by members:read

### LINKED ISSUE #1369: security(api): POST /api/profiles/batch has no permission guard and is an org-directory oracle for guests
## Summary

`POST /api/profiles/batch` carries no permission guard at all, so a `guest` — the org role whose whole point is that it cannot enumerate the directory — resolves member ids to display names and gets a membership oracle by omission.

## Mechanism

`apps/api/src/routes/profile.ts:166` — no guard. The route is org-scoped (org context does run; only `/api/profile` and `/api/profile/password` are in `skipOrgContext`), but that is membership, not permission.

`apps/api/src/lib/permissions.ts:117-120` states the intent for `guest`:

> Member reads minus `members:read` — no enumerating the org directory.

A guest posts `{ ids: [...] }` and gets display names back, plus "which of these ids are in this org" from what is omitted.

## Proposed fix

Add `requirePermission("members", "read")`.

## Verified correct alongside

`GET /api/orgs/:orgId` itself is fine — `buildOrgDetail` gates `members` on `members:read` and `invitations` on `members:invite`, and a guest gets empty arrays.

pierrecabriere:
Closes #1369

## Root cause

`apps/api/src/routes/profile.ts:166` (before this PR) registered `POST /api/profiles/batch` with no permission guard. Org context runs (the route is not in `skipOrgContext`), so the query is org-scoped — but membership is not permission. `apps/api/src/lib/permissions.ts:117-127` defines `guest` as the member read set **minus** `members:read`, "no enumerating the org directory". A guest could therefore POST `{ ids: [...] }`, get display names back for org members, and read membership off the ids the response omitted — an oracle for "is this user in my org?".

## Fix

One guard: `requirePermission("members", "read")` on the route. Resolving ids to display names *is* the org directory read one page at a time, so the route now carries the same permission `buildOrgDetail` gates its `members` array on (RBAC spec §3.2 / §6.5). No response narrowing: the resource is the directory itself, there is no subset a guest legitimately needs (the spec is explicit that even a guest holding preset `admin` in a space "enumerates nothing else" — `docs/architecture/RBAC_PERMISSIONS_SPEC.md:337`), and there is no in-repo consumer to narrow for (`rg 'profiles/batch'` finds only the API route, its OpenAPI path, and tests — the SPA never calls it).

Deliberate side effect: `members:read` is not in `API_KEY_ALLOWED_SCOPES`, so API keys now get 403 here. That matches the rest of this file — `/api/profile`, `PATCH /api/profile` and `/api/profile/password` already refuse `authMethod === "api_key"` outright, because the dashboard user's identity record is session-only.

OpenAPI: `403` documented on the now-guarded route and the description states the permission (`apps/api/src/openapi/paths/profile.ts`). `apps/web/src/api/schema.d.ts` regenerated via `bun run generate:api`. `detect:breaking` classifies the added 403 as non-breaking, so `baseline.json` is untouched.

## Tests

`apps/api/test/integration/routes/profile.test.ts` — two cases added to the existing `POST /api/profiles/batch` describe:
- a `guest` is refused with 403 (the regression),
- an ordinary `member` still gets 200 with one row (positive control, so the bound is not silently over-tightened).

Both use `orgOnlyHeaders` rather than `authHeaders`: a guest has no implicit space membership, and the refusal under test must be the permission one, not a space-context one.

Commands run (from the worktree root):

- `set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/integration/routes/profile.test.ts` → **22 pass, 0 fail** (53 expect calls)
- `... bun test apps/api/test/integration/multi-tenancy.test.ts` → **14 pass, 0 fail** (it also exercises this route's cross-org isolation)
- Negative control: with the guard temporarily removed, the new guest case fails `Expected: 403 / Received: 200`. Guard restored, suite re-run.
- `bunx tsc --noEmit -p apps/api` → clean (exit 0)
- `bun run verify:openapi` → ALL CHECKS PASSED; `bun run verify:api-types` → up to date; `bun run detect:breaking` → 0 breaking, 1 non-breaking (the new 403)
- `bunx eslint` + `bunx prettier --check` on the touched files → clean

## Notes for the orchestrator

- **Behaviour change for API keys**: any external integration calling `POST /api/profiles/batch` with an API key now gets 403. No in-repo caller does; worth one line in the release notes.
- Overlap with siblings: none. #1337 (space SMTP/social) and #1341 (route ordering) touch different files; `apps/api/src/routes/profile.ts` and `apps/api/src/openapi/paths/profile.ts` are mine alone.
- `apps/web/src/api/schema.d.ts` is generated — if a sibling also regenerates it, take both hunks (mine is the `batchGetProfiles` 403 + description).

### DIFF
--- a/apps/api/src/openapi/paths/profile.ts
+++ b/apps/api/src/openapi/paths/profile.ts
@@ -151,7 +151,8 @@ export const profilePaths = {
       operationId: "batchGetProfiles",
       tags: ["Profile"],
       summary: "Batch lookup profiles",
-      description: "Retrieve display names for a list of user IDs.",
+      description:
+        "Retrieve display names for a list of user IDs. Reads the organization directory, so it requires `members:read`.",
       parameters: [{ $ref: "#/components/parameters/XOrgId" }],
       requestBody: {
         required: true,
@@ -199,6 +200,7 @@ export const profilePaths = {
         },
         "400": { $ref: "#/components/responses/ValidationError" },
         "401": { $ref: "#/components/responses/Unauthorized" },
+        "403": { $ref: "#/components/responses/Forbidden" },
       },
     },
   },
--- a/apps/api/src/routes/profile.ts
+++ b/apps/api/src/routes/profile.ts
@@ -14,6 +14,7 @@ import { readJsonBody } from "@appstrate/core/request-body";
 import { listResponse } from "../lib/list-response.ts";
 import { scopedWhere } from "../lib/db-helpers.ts";
 import { getErrorMessage } from "@appstrate/core/errors";
+import { requirePermission } from "../middleware/require-permission.ts";
 import { setDisplayName } from "../services/profile.ts";
 
 export const profileUpdateSchema = z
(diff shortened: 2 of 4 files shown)


pierrecabriere:

PR #1425, opened 2026-09-10
fix(api): prove permission before resolving the agent on eleven routes

### LINKED ISSUE #1341: security(api): eleven routes run requireAgent() before the permission guard, leaking a catalog existence oracle

pierrecabriere:
Closes #1341

## Root cause

`requireAgent()` (`apps/api/src/middleware/guards.ts:12`) resolves the agent named in the route params and 404s when the caller's space cannot reach it. Eleven routes mounted it **before** their permission guard, so a caller lacking the guarded permission got **403 for an agent that exists** and **404 for one that does not** — an enumeration oracle over the space's private agent catalog. An API key scoped to any unrelated permission could walk it through, for example, `PUT /api/agents/@x/y/proxy`.

Sites: `routes/agents.ts` (input-settings, connection-readiness, PUT proxy, PUT model, GET/DELETE persistence, DELETE persistence/memories/:id, DELETE persistence/pinned/:id), `routes/runs.ts:218` (POST run) and `:829` (DELETE runs), `routes/schedules.ts:306` (POST schedules). Four sibling routes already had the correct order and `agents.ts:322` already documented why.

## Fix

**Commit 1 — the ordering.** The permission guard moves ahead of the lookup on the eleven routes. Nothing else changes: both status codes were already reachable on every one of these routes (the guard was mounted, just later), so **no OpenAPI response declaration changes** — I checked all eleven, each already declares both 403 and 404, and `bun run verify:openapi` passes including its "every guaranteed 403 is documented" check. On `DELETE .../runs` the lookup moves after *both* permission guards.

I considered a combined `requireAgentWithPermission(...)` factory to make the order impossible to express. It does not fit: the permissions differ per route, two routes use the `requireAnyPermission` composites (`requireAgentRead`, `requireRunsRead`), `DELETE .../runs` stacks two guards, and two routes carry `rateLimit`/`idempotency` ahead of everything. It would have become `requireAgentAfter(...guards)` — the same ordering spelled through an indirection. The order stays explicit; a test enforces it.

**Commit 2 — so the class cannot recur.** The defect is a property of the mount **order**: invisible at runtime, invisible in a diff that adds one more route by copying its neighbour. `test/integration/middleware/agent-lookup-permission-order.test.ts` reads the real order off Hono's route table and asserts that a permission guard precedes every agent lookup, listing offending routes by name.

This needs the two middleware families to be identifiable, so each carries a symbol marker — the idiom `idempotency()` already established. The symbol-vs-name reasoning and the unwrapping of Hono's own `COMPOSED_HANDLER` wrapping (needed because `app.route()` re-wraps sub-router handlers) were already written once for idempotency; rather than copy them twice more I lifted them into `middleware/handler-marker.ts` and pointed `idempotency()` at it. `isIdempotencyAware` keeps its signature and its callers.

## Tests

New: `apps/api/test/integration/middleware/agent-lookup-permission-order.test.ts` (3 tests) — the ordering assertion, a marker-readability check so the assertion cannot pass vacuously if a marker stops being stamped, and a negative control on the detector itself.

Validated the test genuinely catches this bug, not just its own fixture: with the three route files restored to `origin/main` it fails and names **exactly the eleven routes the issue lists**, route for route.

```
TEST_TIER=0 bun test apps/api/test/integration/middleware/            → 112 pass, 0 fail (16 files)
  ...plus test/unit/guards.test.ts, test/unit/require-permission.test.ts
  (includes idempotency-contract.test.ts — covers the handler-marker refactor)
TEST_TIER=0 bun test agents.test.ts agent-connection-readiness.test.ts
  agents-space-packages-read-permissions.test.ts
  runs-schedules-read-permissions.test.ts schedules.test.ts            → 109 pass, 0 fail
TEST_TIER=0 bun test runs.test.ts runs-read-isolation.test.ts
  runs-body-validation.test.ts openapi-response-validation.test.ts
  multi-tenancy.test.ts                                                → 138 pass, 0 fail
bunx tsc --noEmit -p apps/api                                          → clean
bun run verify:openapi                                                 → ALL CHECKS PASSED
bunx eslint + prettier on the 8 touched files                          → clean
```

## Notes for the orchestrator

- **No migration, no env var, no OpenAPI change.** Nothing to sequence at deploy.
- **Overlap with siblings #1313/#1338 in `guards.ts`:** kept deliberately minimal and far from the code they move. My edit adds an import, a marker constant + `isAgentLookup()` at the top of the file, and changes the two `return async (…) => {…};` in `requireAgent`/`requireOrgAgent` into `return markHandler(async (…) => {…}, AGENT_LOOKUP);`. I did not touch `apiKeySpaceScopeGuard` (the function they rename to `pinnedSpaceScopeGuard`), so the merge should be clean.
- **Behaviour change worth naming in release notes:** these eleven routes now answer **403 instead of 404** to a caller who lacks the permission and names a nonexistent agent. Both codes were already declared and already reachable, so no client contract breaks — but a client that branched on 404 to mean "gone" will now see 403 in that one combination.
- **Pre-existing, not fixed here (out of scope):** `GET /api/agents/{scope}/{name}/schedules` is already permission-first yet still runs `requireAgent()`, so it can return 404 — but `src/openapi/paths/schedules.ts:42` declares only 200/401/403. An under-declaration unrelated to this issue; worth its own ticket.

### DIFF
--- a/apps/api/src/middleware/guards.ts
+++ b/apps/api/src/middleware/guards.ts
@@ -6,11 +6,32 @@ import { getPackage, getPackageWithAccess } from "../services/package-catalog.ts
 import { assertPackageMutationAccess } from "../lib/package-access.ts";
 import { getRunningRunsForPackage } from "../services/state/runs.ts";
 import { ApiError, forbidden, conflict, invalidRequest } from "../lib/errors.ts";
+import { hasHandlerMarker, markHandler } from "./handler-marker.ts";
+
+/**
+ * Marker stamped on every middleware that resolves an agent from the route
+ * params and 404s when it is unreachable.
+ *
+ * Such a middleware answers "does this agent exist?" before the route has
+ * proven the caller may ask. Mounted ahead of the permission guard it turns
+ * 403-vs-404 into an enumeration oracle over the space's private catalog, so
+ * the mount ORDER is a security property — and one no reviewer reliably sees.
+ * The marker lets a conformance test read the real order off Hono's route
+ * table instead (see
+ * `test/integration/middleware/agent-lookup-permission-order.test.ts`).
+ */
+const AGENT_LOOKUP = Symbol.for("appstrate.agentLookup");
+
+/** True when `handler` is a middleware produced by {@link requireAgent} or
+ *  {@link requireOrgAgent} — i.e. it can 404 on an unreachable agent. */
+export function isAgentLookup(handler: unknown): boolean {
+  return hasHandlerMarker(handler, AGENT_LOOKUP);
+}
 
 /** Middleware: load an agent by route param and set it on context, or 404.
  *  Also checks that the current space has access to the package. */
 export function requireAgent() {
-  return async (c: Context<AppEnv>, next: Next) => {
+  return markHandler(async (c: Context<AppEnv>, next: Next) => {
     const scope = c.req.param("scope");
     const name = c.req.param("name");
     const packageId = `${scope}/${name}`;
@@ -28,14 +49,14 @@ export function requireAgent() {
     }
     c.set("package", agent);
     return next();
-  };
+  }, AGENT_LOOKUP);
 }
 
 /** Middleware: load an agent by route param and set it on context, or 404.
  *  Checks org ownership only — does NOT check space-level access.
  *  Use for org-level operations (editing manifest, skills, tools). */
 export function requireOrgAgent() {
-  return async (c: Context<AppEnv>, next: Next) => {
+  return markHandler(async (c: Context<AppEnv>, next: Next) => {
     const scope = c.req.param("scope");
     const name = c.req.param("name");
     const packageId = `${scope}/${name}`;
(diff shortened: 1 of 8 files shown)


pierrecabriere:

PR #1427, opened 2026-09-10
fix(cli): read keyring failures the way @napi-rs/keyring 2.x reports them

### LINKED ISSUE #1321: cli: the @napi-rs/keyring 2.0 bump activates dead error markers — hard refusal instead of file fallback, and logout can leave a plaintext token

pierrecabriere:
Closes #1321

## Root cause

`@napi-rs/keyring` 2.0 (bumped in `ebb3764d2`, #1299) reworded every `keyring-core` error Display **and** made `getPassword`/`deletePassword` throw where 1.x swallowed. That armed two `catch` blocks whose markers no longer exist in the shipped binary — `apps/cli/src/lib/keyring.ts:139` matched `"Platform secure storage failure"` / `"No storage"`, and `:146` matched `"No matching entry"`; the binary emits `"Platform failure: "` and `"Couldn't access platform storage: "` (verified by reading the installed `keyring.darwin-arm64.node`). So every store failure classified as `"broken"`:

- a CI runner / stripped container with no secret-service daemon stopped falling back to the 0600 `credentials.json` and died on *"the OS keyring is installed but not serving"*;
- `deleteTokens` (`:388`) threw **before** `deleteFromFile`, so `appstrate logout` on a locked keychain left the live 30-day refresh token in plaintext on disk — the exact outcome the refusal exists to prevent.

## Fix

Classification now works off the two Display prefixes 2.x actually emits, split by what they *mean* instead of lumped as the 1.x list had them:

- `PlatformFailure` → `store-unavailable`: no store on this host, so there is no keyring protection to downgrade from — the 0600 file is the documented fallback.
- `NoStorageAccess` (and any unknown wording) → `store-locked`: the machine does protect secrets and merely will not serve us, so the **write** path still refuses unless `APPSTRATE_ALLOW_PLAINTEXT_TOKENS=1`. Unknown wording lands on the refusing side by construction: a future rewording fails loud, never into a silent plaintext write.

The `entry-missing` class is gone: 2.x reports an absent credential as `getPassword() === null` / `deletePassword() === false`, never as a throw (its own `index.d.ts` states the contract; verified against the real binding). That also makes `_shouldRefuseWindowsFallback` unconditional, so it no longer takes the error.

`deleteTokens` never refuses any more — refusing a *delete* guarantees the plaintext survives. It clears the file store first and only then reports a surviving keyring entry, with an error that says what remains and how to remove it. `lib/api.ts`'s three `deleteTokens(...).catch(() => {})` scrub paths get the same guarantee for free.

Read-path behaviour on a locked store is deliberately unchanged: it still refuses with the documented `APPSTRATE_ALLOW_PLAINTEXT_TOKENS=1` escape hatch, which is this file's stated posture (only the *classification* of which errors reach that branch was wrong).

## Tests

- `apps/cli/test/keyring-error-markers.test.ts` (new): pins both prefixes against the installed native binary and asserts the three retired 1.x wordings are absent, so the next bump that rewords them is red instead of silently mis-classifying. The `.node` is read as **bytes and never loaded** — `new Entry(...)` on a daemon-less Linux runner segfaults the process (same reason as the note on `_isKeyringFactoryOverriddenForTesting`), so no test may call the real binding.
- `apps/cli/test/keyring.test.ts`: fixtures moved to the real 2.x wordings (they were all 1.x strings, which is why the suite stayed green through the bump); the delete case now asserts the credentials file **is** cleared when the keyring delete fails; the classification block covers both sides of the split plus the refuse-on-unknown-wording case.

Commands (from `apps/cli`, per the CLI test convention):

- `TEST_TIER=0 bun test test/keyring.test.ts test/keyring-error-markers.test.ts test/logout.test.ts` → **54 pass, 0 fail**
- `TEST_TIER=0 bun test` over the auth/keyring-dependent files (keyring, keyring-error-markers, logout, auth-fixture, token, api, device-flow, whoami, org-command, jwt-identity, doctor) → **192 pass, 0 fail**
- `bunx tsc --noEmit -p apps/cli` → clean; `bunx eslint` + `bunx prettier --check` on the touched files → clean; `bunx knip --workspace apps/cli` → no findings

## Notes for the orchestrator

- No migration, no env var, no OpenAPI surface. Branch is stacked on `fix/issue-1364` as instructed.
- No overlap with #1361. `apps/cli/src/commands/logout.ts` is **not** touched: the fix belongs in the keyring seam, and logout already surfaces whatever `deleteTokens` reports.
- Linux wording is unverified by hand (the issue's open point), but it no longer needs to be: the pin reads whichever platform binary is installed, so CI on linux-x64 checks the linux strings itself.
- Behaviour change worth a release note: on a locked keychain, `appstrate logout` now exits non-zero after clearing the file store, saying the keyring copy may remain — previously it threw with the plaintext still on disk.

### DIFF
--- a/apps/cli/src/lib/keyring.ts
+++ b/apps/cli/src/lib/keyring.ts
@@ -7,8 +7,9 @@
  * libsecret/DBus on Linux, Credential Manager on Windows). Fallback:
  * `$XDG_CONFIG_HOME/appstrate/credentials.json` with `0600` permissions
  * when no keyring daemon is available (CI runners, stripped containers
- * — confirmed during preflight PF-1 where `@napi-rs/keyring` threw
- * `Platform secure storage failure` on a bare Debian slim image).
+ * — confirmed during preflight PF-1 where `@napi-rs/keyring` threw a
+ * `PlatformFailure` on a bare Debian slim image). Which throws take
+ * that fallback and which are refused is {@link classifyKeyringError}.
  *
  * Tokens are scoped by profile: the keyring entry key is
  * `(appstrate, <profile>)` so profiles share the service name.
@@ -33,9 +34,9 @@ import { getErrorMessage } from "@appstrate/core/errors";
  * Opt-in escape hatch for environments where the keyring daemon is
  * present but refuses to serve (SSH-attached macOS without a logged-in
  * loginwindow, frozen gnome-keyring, stripped container with a stale
- * libsecret socket). Without this env var set, a `"broken"` keyring
- * error is fatal instead of silently writing plaintext tokens to the
- * file fallback. The rationale is symmetric with the Windows refusal:
+ * libsecret socket). Without this env var set, a `"store-locked"`
+ * keyring error is fatal instead of silently writing plaintext tokens
+ * to the file fallback. The rationale is symmetric with the Windows refusal:
  * if the user's machine is configured to protect secrets via the OS
  * keyring, a broken backend is a signal, not a reason to quietly
  * downgrade.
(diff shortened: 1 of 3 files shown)


pierrecabriere:

PR #1429, opened 2026-09-10
fix(scripts): enforce 0008's run-once window with a marker, not a comment

### LINKED ISSUE #1319: scripts: operator script 0008 is destructive if re-run, guarded only by a comment

pierrecabriere:
Closes #1319

## Root cause

`scripts/migration/0008-org-viewer-to-guest.sql:93-95` captures the OAuth signup
clients step 4 must snapshot with `signup_role = 'guest' AND
signup_space_assignments = '[]'::jsonb`. Unlike steps 1-3, that predicate does
not remove its own condition: once the new application is up, `guest` is
writable and the pair also describes a client an admin **deliberately** left
with no space assignments. A second run (`:155-159`) then hands that client
`viewer` rows in every space created since — a widening of the
auto-provisioning path. The only guard was the file's own header (`:28-33`) and
`0012`'s preamble saying "do not re-run".

## Fix

The window is now mechanical, not documentary.

- The script creates `drizzle.migration_scripts (script, ran_at)` if absent and,
  as its last statement inside the same transaction, records
  `'0008-org-viewer-to-guest'`. An abort rolls the marker back with everything
  else, so a failed run stays re-runnable in full.
- Step 4's **capture** gains `AND NOT EXISTS (… migration_scripts …)`. On every
  run past the first the captured set is empty and step 4 writes nothing. The
  guard sits on the capture rather than on the `UPDATE` so step 5's coverage
  aborts still verify exactly what was captured.
- Rather than refusing the whole script, a later run stays useful: steps 1-3 are
  genuinely idempotent and still converge, and a `RAISE NOTICE` names the
  clients whose empty snapshot the stale predicate still matches — i.e. exactly
  what it declined to widen — plus the timestamp of the run that closed the
  window. Re-running step 4 on purpose means deleting that row by hand, which
  the header spells out.

Why `drizzle` and not `public`: it is the schema the migrator already keeps its
journals in (`__drizzle_migrations`, `ee_migrations`), so the app never reads
the table, drizzle never generates from it, and `scripts/check-index-drift.ts`
(which reads `public` only) does not see it. No drizzle migration, no new env
var. Prose updated where the old claim became false: `0012`'s "0008 CANNOT BE
RE-RUN" paragraph (its other two reasons for being a separate file stand),
`scripts/migration/README.md` step 5, `docs/architecture/RBAC_PERMISSIONS_SPEC.md`
§ rows-0008, and the `[Unreleased]` RBAC entry in `CHANGELOG.md`.

## Tests

`apps/api/test/unit/migration-0008-org-viewer-to-guest.test.ts` — new case
"skips step 4 past its window instead of widening a client left empty on
purpose". A legacy client is snapshotted by the first run (positive control, so
the assertion below cannot pass on a broken fixture), then a deliberately-empty
`guest` client and a new space appear and a second run leaves it at `[]`. The
existing "never widens on replay" case only covered a client whose snapshot step
4 had already filled, which stops matching the predicate by itself.

```
  apps/api/test/unit/migration-0008-org-viewer-to-guest.test.ts \
  apps/api/test/unit/migration-0059-drop-org-viewer.test.ts
→ 17 pass, 0 fail, 71 expect()
bun run verify:no-migration-dml → 68 file(s) scanned, green
bunx eslint / prettier on the touched files → clean
```

Negative control: with the `NOT EXISTS` removed, the new case fails with the
deliberately-empty client collecting `viewer` rows in all three spaces.

## Notes for the orchestrator

- No migration index consumed, no env var, no OpenAPI change.
- Deploy ordering is unchanged. Operators who have *already* run `0008` from
  `main` have no marker row; their next run behaves exactly as today's re-run
  would for steps 1-3 and, for step 4, will still fill snapshots once more and
  then close the window. Production has not run `0008` (README log: "not
  applied", the two viewer rows were moved by hand), so this lands ahead of any
  real run.
- Overlap: `scripts/migration/README.md` is also touched by #1318 (`0010`
  section) and #1317/#1350 (`0057` section). My edit is one sentence in step 5
  of the RBAC section — should merge cleanly, but it is the same file.


pierrecabriere:

PR #1430, opened 2026-09-10
fix(oidc): gate per-space SMTP and social config on space membership

### LINKED ISSUE #1337: security(oidc): per-space SMTP and social config are readable by a guest, for spaces they are not in

pierrecabriere:
Closes #1337

## Root cause

`apps/api/src/modules/oidc/routes.ts` mounts the per-space SMTP and social-provider routes at the app root. `/api/spaces` is not in `SPACE_SCOPED_PREFIXES` (`apps/api/src/middleware/space-context.ts:37`), so no space context ever resolved for them — the routes were gated on the **org-level** `spaces:read` / `spaces:write`, strings `member` and `guest` hold for the whole org (`apps/api/src/lib/permissions.ts`), and the only space check was `assertSpaceBelongsToOrg`, which proves org ownership and nothing else. A guest read the SMTP `host`/`port`/`username` and the social `clientId`/`scopes` of every space in the org, `private` ones included, and the `"Space not found"` vs `"SMTP configuration not found"` split turned it into a private-space existence oracle. `apiKeySpaceScopeGuard` was not in the chain either, so a key bound to space A reached space B by path param.

Secrets themselves were already safe: `mapRow` in `services/smtp.ts` and `services/social.ts` project the view without `passEncrypted` / `clientSecretEncrypted` — verified, and the new tests assert it.

## Fix

The router now enters the space named by `:id` itself, via `enterSpaceContext` — the core seam documented for module routes in `apps/api/src/modules/README.md` and used by `webhooks`. It runs the canonical `validateSpaceInOrg` (400 on a malformed id, 404 outside the org) and then `applySpacePermissions`, which refuses a non-member with 403 `not_a_space_member`, or 404 for a `private` space. Each of the seven routes then declares the **space-level** `space-settings:write`.

Why that permission and not a new one: `space-settings` is exactly "per-space configuration" (`packages/core/src/permissions.ts:69`), it is what `PATCH /api/spaces/:id` already requires, and it is held by preset `admin` — which org owners and admins get implicitly in every space (`resolveSpaceRole`), so no operator loses access. Reads take the same string as writes because the resource has one action by design; adding a `read` action would mean a new core permission string, a role-grant matrix change, an API-key allowlist entry and a `@appstrate/core` release, to serve a role nobody wants (read a space's mail relay without administering it).

`apiKeySpaceScopeGuard` runs first in the chain: an API key's membership is resolved from its **creator**, so a key bound to space A whose creator administers space B would otherwise still reach B.

No refactor beyond deleting the local `assertSpaceBelongsToOrg` (`enterSpaceContext` subsumes it). No new helper — the existing seam and the existing guard were the right pieces.

Also moved: the SPA's `space/auth` route guard and its nav entry, from `spaces:write` to `space-settings:write`, so the UI matches the API; and the `space-settings` row of `docs/architecture/RBAC_PERMISSIONS_SPEC.md`. The seven OpenAPI operations now state the permission (regenerated `apps/web/src/api/schema.d.ts`; `detect:breaking` reports no schema change).

## Tests

Seven regression tests added to the two nearest existing files —
`apps/api/src/modules/oidc/test/integration/routes/space-smtp.test.ts` and `space-social.test.ts`:
a guest gets 404 on a `private` space it is not in (and the response does not contain the stored host / clientId), 403 on an open space it is not in, a `viewer` member of the space gets 403, a guest's write is still refused, and a space `admin` who is only an org **guest** gets 200 — proving the gate is membership, not org role.

```
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/src/modules/oidc/test
  → 572 pass, 0 fail (53 files)
set -a && . ./.env && set +a && TEST_TIER=0 bun test apps/api/test/integration/routes/strict-request-bodies.test.ts
  → 7 pass, 0 fail
bun run verify:api-types        → up to date (after bun run generate:api)
bun run detect:breaking         → 0 breaking, 0 non-breaking
```

Negative control: with `apps/api/src/modules/oidc/routes.ts` reverted to `origin/main` and the new tests kept, 5 of the 7 fail — including both private-space reads, which return the configuration on `main`.

## Notes for the orchestrator

- No migration, no env var.
- **Overlap with #1313**: this branch imports the existing `apiKeySpaceScopeGuard` from `apps/api/src/middleware/guards.ts`. If #1313 renames it to `pinnedSpaceScopeGuard`, the conflict is one identifier in `apps/api/src/modules/oidc/routes.ts` (import + three `router.use` lines). I did not touch `guards.ts` or the spaces router.
- Behaviour change for API consumers: a token holding only org-level `spaces:read`/`spaces:write` now gets 403 on these seven routes. Owners and admins are unaffected.
- `apps/web/src/api/schema.d.ts` is regenerated output, committed separately.


pierrecabriere:

PR #1431, opened 2026-09-10
fix(models): hold one org_models row per un-aliased binding

### LINKED ISSUE #1358: models: duplicate org_models rows are not refused, and per-model spend reporting splits across them

pierrecabriere:
Closes #1358

## Root cause

`org_models` had no uniqueness on `(org_id, credential_id, model_id)`
(`packages/db/src/schema/organizations.ts:449-499`) and `createOrgModel`
(`apps/api/src/services/org-models.ts:399`) did no existence check. Only the
LABEL was deduped, by `deriveModelLabel:380` — a read in a **different**
transaction from the insert, so the multi-add path's N sequential POSTs race it.
Because `llm_usage.model` stores the `org_models` **row id**, a second row for
the same binding splits that model's spend across the copies and under-reports
each; the credential FK is `ON DELETE RESTRICT`, so detaching it means finding
every copy. `seedOrgModelsForCredential:527` already guarded; `POST /models` did
not.

## Fix

**Migration 0064** — `uq_org_models_unaliased_binding`, a unique index on
`(org_id, credential_id, model_id)`.

**Partial on `aliased = false`, and that predicate is the design decision.** An
alias is a public identity the org minted deliberately: its own label, its own
reporting line, its backing hidden from every user-facing surface. Several
aliases over one backing model — or an alias beside the direct row — is the
alias pattern working as intended, and their spend is *meant* to report apart.
Only the un-aliased rows, the ones the picker shows under their catalog name and
that spend reporting is expected to fold together, are held to one per binding.
An existing test (`models.test.ts:253`, aliased round-trip) creates exactly that
pair; it passes unchanged.

**API.** `createOrgModel` wraps its transaction and `updateOrgModel` wraps its
UPDATE; a 23505 becomes `409 model_already_added` with `existing_model_id` in the
problem body — via the existing `isUniqueViolation` in `lib/db-helpers.ts`,
mapped by one private `asDuplicateBinding` shared by both paths (the
`asKeyConflict` idiom from `services/space-roles.ts`, minus its pre-read: the
database is the only arbiter, a read-then-insert check reopens the window the
index exists to close). The catch sits **outside** the transaction because a
failed statement aborts it, so the lookup could not run inside. Declared as 409
on `createModel` and `updateModel` in `openapi/paths/models.ts`; `generate:api`
output committed.

`POST /api/models/seed` refuses a self-duplicating `modelIds` body at the edge
(Zod `.refine`) — its insert is one atomic statement, so a body naming the same
id twice would otherwise fail the whole batch on a constraint the caller cannot
see.

**Data repair.** Per doctrine the migration is shape-only and **fails loudly**
(23505 rolls the batch back) on a database holding duplicates. The repair is
`scripts/migration/0013-org-models-dedupe-bindings.sql`: keeps the oldest
un-aliased row per binding, repoints `organizations.default_model_id`,
`space_packages.model_id`, `package_schedules.model_id_override` and
`llm_usage.model` at it, then deletes the losers — one transaction, so no pointer
outlives its row. Repointing the ledger is what re-consolidates the split spend;
the EE module settles by serial id (`ee_billed_llm_usage`, `ee_billing_cursor`)
and never reads that column, so no billed state, amount or cursor position moves.
`scripts/migration/README.md` gains the pre-flight/runbook section and the log
row.

**Web (minimal).** `ModelFormSubmitOutcome` gains `duplicateModelIds` (a subset
of `failedModelIds`), classified from the `model_already_added` code in
`lib/model-form-submit.ts`. The modal shows a distinct line for them and stops
keeping them checked for retry — the model is there; re-submitting changes
nothing. Two i18n keys, en + fr.

## Tests

- `apps/api/test/integration/routes/models.test.ts` — +3 cases: second POST for
  the same binding → 409 + `existing_model_id` + exactly one row; an aliased
  create over an already-added binding still 201; a PUT repointing onto another
  row's binding → 409. `TEST_TIER=0 bun test apps/api/test/integration/routes/models.test.ts`
  → **58 pass / 0 fail**.
- `apps/web/src/lib/test/model-form-submit.test.ts` — +2 cases (single + batch
  duplicate reporting), existing expectations updated for the new field.
  From `apps/web`: `TEST_TIER=0 bun test src/components/test/model-form-modal.test.tsx src/lib/test/model-form-submit.test.ts`
  → **30 pass / 0 fail**.
- Other org-model consumers (`chat-platform-services`, `run-input-resolution`,
  `run-aliased-model-credential`, `model-providers-oauth-pair-redeem`) →
  **26 pass / 0 fail**.
- `bun test apps/api/test/unit/migration-index-parity.test.ts` → **4 pass** (the
  journal replays with 0064 and the snapshot index is in the catalog).
- `bun run verify:openapi` → ALL CHECKS PASSED. `bun run verify:no-migration-dml`
  → green (69 files, 2 trees). `bun run generate:api` committed.
- `bunx tsc --noEmit` on `apps/api`, `apps/web`, `packages/db` → clean. eslint +
  prettier on every touched file → clean.

## Notes for the orchestrator

- **Migration index 0064** (platform), built from `0060_snapshot.json` with
  `idx: 64`; `prevId` points at 0060's id — re-chain at consolidation. Journal
  entry appended.
- **Operator script index `0013`** under `scripts/migration/` was **not**
  reserved by the orchestrator — 0012 was the last on this base. Renumber if a
  sibling also took 0013.
- **Deploy ordering**: run the README pre-flight count *before* the drizzle
  batch. Non-zero → run `0013-…sql` first, else `0064` aborts the whole release's
  migrations. No env var.
- `apps/api/test/unit/migration-schema-parity.test.ts` fails on this machine with
  a 5 s `beforeAll` timeout — **pre-existing**: it fails identically with 0064
  removed from the journal (control run). Machine load, not this diff.
- **Overlap with #1355** (the base): `apps/web/src/lib/model-form-submit.ts`,
  `model-form-payload.ts`, `model-form-modal.tsx` and
  `lib/test/model-form-submit.test.ts` are that sibling's files. The overlap is
  additive — one new outcome field plus its classification — and its retry/rebind
  contract and every existing assertion are untouched.
- Also touches `apps/web/src/components/test/model-form-modal.test.tsx` (one
  fixture line) purely to satisfy the widened outcome type.


pierrecabriere:

PR #1432, opened 2026-09-10
fix(cli): resolve self-update latest across every release page

### LINKED ISSUE #1361: cli: self-update resolves 'latest' within one release page and can silently downgrade users

pierrecabriere:
Closes #1361

## Root cause

`apps/cli/src/lib/self-update.ts` — `resolveTargetVersion` `return`ed from
inside the page loop as soon as a page held any platform `v*` Release, so
`candidates.sort(compareSemver)` only ever ranked **one** page. The comment
above the loop stated the invariant ("a hotfix for an older line published
after a newer release must not become latest") that the code did not hold.

`performCurlUpdate` only skipped the install when the target **equalled** the
running version (`compareSemver(current, target) === 0`), so any lower target
silently replaced the binary — and the next run resolved the same lower target
and kept the user there.

## Fix

- **Walk every page, then take the maximum.** `candidates` is hoisted out of
  the page loop; the sort happens once, after the walk. The loop stops on a
  *short* page (GitHub fills every page but the last) instead of spending a
  request to discover an empty one.
- **Page size 30 → 100 (the API maximum), cap 5 → 2.** The walk is no longer
  early-exiting, so it must stay cheap: two requests now cover the same 200
  releases a 30-per-page walk would spend seven on. GitHub's unauthenticated
  limit is 60 req/h per IP and this code already special-cases the 403.
- **Never move backwards.** `performCurlUpdate` returns a new
  `"refused-downgrade"` status when the target is older than the running
  version; the command reports it and names `--force` as the override.
  Defence in depth — it holds even if a release vanishes from the list.
  Exit code stays `0`: `self-update` runs unattended (cron, CI images), and a
  box that is simply ahead of the newest release must not start failing. The
  refusal is loud in the message instead.

`compareSemver` was already correct (including §11.4) and is untouched.

## Tests

`apps/cli/test/self-update.test.ts` — 3 new cases plus a `fullPage` fixture
helper (a multi-page fixture needs full pages now that a short page ends the
walk); existing per-page URL and page-cap assertions updated to the new
constants.

- `picks the highest version ACROSS pages, not the highest on the first page holding one`
- `stops at the page cap and still returns the highest candidate seen`
- `refuses to downgrade when the target is older than the running binary`
- `--force downgrades deliberately`

```
cd apps/cli && TEST_TIER=0 bun test test/self-update.test.ts   → 49 pass, 0 fail
bunx tsc --noEmit -p apps/cli                                   → clean
bunx eslint <3 touched files>                                   → clean
```

Negative control: with the early `return` and the downgrade guard reverted,
the three new cases fail (46 pass / 3 fail) — they do catch the bug.

## Notes for the orchestrator

- **Deliberately out of scope: the `scripts/bootstrap.sh` half of the issue.**
  The issue's "Related" section reports three defects in
  `resolve_latest_platform_release` (creation-order pick instead of highest
  semver; `grep -oE` matching `"prerelease": true` inside a release *body*;
  `[ -z "$fields" ] && return 1` aborting the whole 5-page walk on one
  transient curl failure). Same code is duplicated in
  `scripts/bootstrap-runner.sh:87`. They are untouched here: fixing them means
  a semver comparator and a JSON scanner in POSIX `sh`/awk with no test
  harness on the critical install path, and the **published** installer never
  takes that path (`publish-installer.yml` rewrites `__APPSTRATE_VERSION__`,
  so the resolver is only reached when running the script raw from the repo).
  `Closes #1361` will close the issue on merge — **that shell half deserves
  its own issue before this merges**, or the report is lost. I did not open
  one (not authorised to create issues).
- Base is `fix/issue-1364` (the CLI chain tail); `fix/issue-1321` was not on
  origin when this started. No overlap with the 5 commits already on that
  base — none of them touch `self-update.*`.
- Behaviour change worth a release note: `self-update` on a binary newer than
  the latest release now no-ops with a message instead of downgrading.

### DIFF
--- a/apps/cli/src/commands/self-update.ts
+++ b/apps/cli/src/commands/self-update.ts
@@ -137,6 +137,19 @@ export async function runSelfUpdate(opts: SelfUpdateOptions = {}): Promise<SelfU
     };
   }
 
+  if (result.status === "refused-downgrade") {
+    // Exit 0, not a failure code: `self-update` runs unattended (cron, CI
+    // images), and a box that is simply ahead of the newest release must not
+    // start failing those. The refusal is loud in the message instead, and
+    // names the flag that overrides it.
+    return {
+      exitCode: SELF_UPDATE_EXIT.OK,
+      message:
+        `Already on appstrate ${result.version}, which is NEWER than ${target}. ` +
+        `Refusing to downgrade — pass --force to install ${target} anyway.`,
+      detail: result,
+    };
+  }
   if (result.status === "already-up-to-date") {
     return {
       exitCode: SELF_UPDATE_EXIT.OK,
--- a/apps/cli/src/lib/self-update.ts
+++ b/apps/cli/src/lib/self-update.ts
@@ -44,13 +44,20 @@ const RELEASE_URL_BASE = "https://github.com/appstrate/appstrate/releases";
  * `release.yml` can steer an update.
  */
 const RELEASES_API_URL = "https://api.github.com/repos/appstrate/appstrate/releases";
-const RELEASES_PAGE_SIZE = 30;
 /**
- * Pages scanned before giving up: 150 releases. Every platform cycle creates
- * one `v*` plus at most three npm Releases, so a `v*` sits inside the first
- * page in practice; the cap only bounds the API calls when it does not.
+ * 100 is the GitHub API maximum. Every page is walked before the highest
+ * version is picked (see `resolveTargetVersion`), so a larger page is strictly
+ * fewer requests: two calls cover the same 200 releases a 30-per-page walk
+ * would spend seven on.
  */
-const RELEASES_MAX_PAGES = 5;
+const RELEASES_PAGE_SIZE = 100;
+/**
+ * Pages scanned before giving up: 200 releases, ~50 platform cycles back (each
+ * creates one `v*` plus at most three npm Releases). The cap bounds the API
+ * calls — GitHub's unauthenticated limit is 60 req/h per IP, and every
+ * resolution now pays for the full walk rather than stopping early.
+ */
+const RELEASES_MAX_PAGES = 2;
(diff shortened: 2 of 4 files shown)


pierrecabriere:

PR #1433, opened 2026-09-10
fix(scripts): migration 0010 tells a booted target from a copied one, and reads the source's level

### LINKED ISSUE #1318: scripts: migration 0010 refuses after a single EE boot, and its refusal message invites data loss

pierrecabriere:
Closes #1318

## Root cause

**Defect 1 — the refusal could not tell a prior copy from a single boot.**
`scripts/migration/0010-ee-tables-into-platform-db.ts:373-381` refused on *any*
non-empty `ee_*` table. But `packages/module-ee/src/index.ts:139` calls
`ensureCursorSeeded` at `init()`, which INSERTs the singleton row into
`ee_billing_cursor` — a restart, a health probe or an early `docker compose up`
is enough. The message's first branch then read "If the move already ran,
nothing is left to do", i.e. an invitation to abandon the source database
holding every Stripe customer/subscription id. Worse, that row would also have
collided with the copied cursor on the primary key.

**Defect 2 — the source's migration level was never read.** The header asserted
"the expected production source is at 0003" but the script only ever inferred
shape from column presence, which structurally cannot tell `0003` from `0002`:
`0003_normalize_free_subscription_status.sql` rewrites rows and adds nothing.
A pre-`0003` source lands free-plan rows carrying
`subscription_status = 'canceled'`, and `packages/module-ee/src/billing/quota-check.ts:126-131`
answers `402 subscription_blocked` for every one of them.

## Fix

**Defect 1** — `BOOT_SEEDED_TABLE` names `ee_billing_cursor` as the one table a
boot may legitimately have written. A row there alone is deleted inside the copy
transaction and replaced by the source's authoritative watermark (the source
cursor indexes the same platform `llm_usage` ledger, and the boot-seeded one is
a frontier from a later moment — restoring the source's is the direction that
cannot lose revenue). A row in any *other* table still refuses, and the refusal
now prints both sides' per-table counts (reusing `summary()`) as evidence
instead of an instruction: no branch of it says to empty, drop or overwrite a
database.

**Defect 2** — the level is READ from the source's own journal and compared to
`REQUIRED_SOURCE_TAG` (`0003…`, resolved to an idx against the module's journal
so a rename fails loudly). Matched on the `created_at` stamp drizzle copies from
the journal's `when`; a stamp matching no entry, or no journal at all, refuses
too. Both journal names are accepted — `drizzle.__drizzle_migrations` is what
the module wrote while it owned a database outright, `drizzle.ee_migrations`
what the in-tree one writes; that is the same before/after split the
`cloud_`/`ee_` prefixes already have.

The issue's "copy first, then migrate" is deliberately *not* what this does:
every DML the module ever shipped (`0001`, `0003`) names the pre-rename
`cloud_*` tables, so once the target is at `0005` those statements have no table
to touch and can never be replayed over copied rows. Refusing a source below the
last row-rewriting migration is the reachable form of the same guarantee, and is
stated as such in the header and next to the constant.

Small refactor, DRY: the module's journal is now parsed once into `JOURNAL`
(`{idx, when, tag}`), reused by `declaredColumns()`, the pending-migration
count (`shippedMigrations()` deleted — it re-read the same file) and the new
level check. The source-level check runs right after prefix detection, before
the per-table accounting, so nothing touches the target.

## Tests

`scripts/test/migration-0010-ee-tables-into-platform-db.test.ts`, +4 cases
(7 → 11), plus `applyMigration` now records each replayed file in
`drizzle.__drizzle_migrations` with the journal's `when` — which is what the
pre-in-tree module actually did, so the `cloud_*` fixture is more faithful than
before:

- a target the module has already booted against: the seeded watermark is
  replaced by the source's (`last_llm_usage_id` 7000 → 41, `floor_id` → the
  declared default), the copy succeeds;
- the same target holding a billing account as well: still refused, asserting
  the counts table is printed and that the words "nothing is left to do" /
  "empty the target" do not appear;
- a source at `0002`: refused, target untouched (no `ee_*` table created);
- a source whose journal is gone: refused.

```
pg-test-lock.sh sh -c 'PATH=<docker-stub>:$PATH; set -a && . ./.env && set +a &&
  bun test scripts/test/migration-0010-ee-tables-into-platform-db.test.ts'
→ 11 pass / 0 fail, 60 expect() calls
```

Negative control: same file against `origin/main`'s script → **7 fail / 4 pass**
(the four new cases plus the three they cascade into, since the old script
happily copied the `0002` source into the shared refusal target).

Also green: `bunx tsc --noEmit -p scripts/tsconfig.json`, `bunx eslint` +
`prettier --check` on the touched files, `bun run verify:module-isolation`,
`verify:license-boundary`, `verify:no-migration-dml`.

## Notes for the orchestrator

- **No migration, no env var, no route.** Script + its test + one row of
  `scripts/migration/README.md`.
- **Runbook change**: "run 0010 BEFORE any EE boot" is no longer a hard
  precondition — a target the module has merely booted against is now handled.
  The README row says so.
- **Overlap with #1317/#1350**: my README edit is confined to the `0010` row of
  the Log table (their edits are in the `0057` section and the `0011` header).
  Unavoidable churn: `0010` was already the widest row in both columns, so
  prettier re-padded the other 11 rows' trailing whitespace. Pure whitespace —
  a conflict can only appear on a row a sibling also edited, and resolves to
  "take both cell contents, let prettier re-pad".
- Not touching #1319 (`0008` destructive re-run).


pierrecabriere:

PR #1435, opened 2026-09-10
fix(lint): extend the type-aware block to the cli, runtime-pi and scripts, and fix the defects it finds

### LINKED ISSUE #1359: lint: the type-aware block excludes the sidecar, the runners and the CLI — and there are real hits there
## Summary

The type-aware lint block added by `557db0e97` scopes to `apps/api/src/**` and `packages/*/src/**`, leaving out `runtime-pi/sidecar/**`, `runtime-pi/runners/**`, `apps/cli/src/**` and `scripts/**` — while its own rationale names "an un-awaited promise in a **run orchestrator**" as the motivating failure. The sidecar is exactly that layer.

Running the two rules over the excluded trees finds real hits, one of them a byte-for-byte twin of the bug the same commit fixed.

## Mechanism

`eslint.config.mjs:793`:

```js
files: ["apps/api/src/**/*.ts", "packages/*/src/**/*.ts"],
```

## Hits

**`apps/cli/src/lib/download.ts:161` — CONFIRMED.** `sink.write(bytes)` un-awaited in the download loop, on a `Bun.file(destPath).writer()` with **no `highWaterMark`**. This is the same defect as the `packages/core/src/storage-fs.ts` one the commit fixed, in the copy that was out of scope. Per core's own comment (`storage-fs.ts:266-271`), a default `FileSink` accumulates every chunk in memory until `end()` — so the CLI buffers the whole artifact in RAM, and a mid-write failure escapes the surrounding `try` as an unhandled rejection (fatal in Bun), leaving the partial file un-unlinked.

**`runtime-pi/sidecar/integration-mitm-listener.ts:264` — PLAUSIBLE, low.** The `async` connection handler passed to `netCreateServer` is neither awaited nor `void`-ed nor `.catch`-ed. `handleInboundConnection` does wrap `resolveTlsServer` in try/catch (`:456-460`), so the practical rejection surface is thin — but the safety is by inspection, which is exactly what these rules exist to stop relying on.

## Proposed fix

Extend the block's `files` to the sidecar, the runners and the CLI, and fix the two hits. `scripts/**` is a judgement call; the other three are backend code with the same failure mode the rule was added for.

pierrecabriere:
Closes #1359

## Root cause

`eslint.config.mjs`'s type-aware backend block scoped itself to
`files: ["apps/api/src/**/*.ts", "packages/*/src/**/*.ts"]`, so `apps/cli/src`,
`runtime-pi/**` and `scripts/**` got none of `no-floating-promises` /
`no-misused-promises` / `await-thenable` — while the block's own rationale names
"an un-awaited promise in a **run orchestrator**" as the motivating failure, and
that orchestrator lives in `runtime-pi/`.

Running the three rules over the excluded trees found **8 violations**, four of
them defects.

### 1. The four defects

- **`apps/cli/src/lib/download.ts:161`** — the one the issue confirmed, and a
  byte-for-byte twin of the `packages/core/src/storage-fs.ts` defect that the
  commit which introduced this block fixed. `sink.write(bytes)` was un-awaited
  on a `Bun.file(destPath).writer()` opened with **no `highWaterMark`**; per
  core's own note (`storage-fs.ts:266-271`) such a sink accumulates every chunk
  in memory and flushes only on `end()`. So the CLI held a whole 113 MB
  artifact resident — this module's own doc comment promises "peak memory ≈ one
  chunk", which was false — and a mid-write failure escaped the surrounding
  `try` as an unhandled rejection (fatal in Bun), leaving the partial file at
  `destPath` un-unlinked. Fixed with an explicit 1 MiB watermark and an awaited
  write, matching core.
- **`runtime-pi/sidecar/integration-mitm-listener.ts:264`** — the `async`
  connection handler passed to `netCreateServer` had its promise dropped.
  `handleInboundConnection` awaits the SSRF/DNS rebind guard and the
  ClientHello reads outside any `try`, so a rejection there is an unhandled
  rejection that takes the **sidecar process** down, and with it the run it
  proxies for. Routed into the same `tls-error` + `rawSocket.destroy()` path
  the function's five other failure branches already use.
- **`runtime-pi/provision.ts:277`** — floating `writer.write(value)`; a
  mid-write failure bypassed the `catch` that routes stream errors through
  `die()`. Now awaited (the existing `flush()` still does the backpressure).
- **`runtime-pi/entrypoint.ts:752`** — `if (piSdkWarmup && …)` is a truthiness
  test on a value that is unconditionally a Promise, so the guard was dead and
  read as if the handle were optional. Removed.

The other four hits (three `server.stop()` calls in a dev-only OAuth helper,
one `Promise.all` over a null-mixed iterable in a migration) are not defects and
are fixed in place — `void` where ignoring is intended, a uniformly-async mapper
where it is not.

### 2. Heap headroom for the gate — the precondition, not scope creep

Widening the block is what discovered that `bun run lint` had **no memory
headroom left**. `scripts/lint.ts:256` spawns `node_modules/.bin/eslint`, a
NODE script, so the lint runs under node's default ~4 GB old-space, and this
block's TypeScript programs are essentially all of it. Cold, exit code
captured:

| config | cold `bun run lint` |
| ------ | ------------------- |
| `main` as-is | exit 0 |
| + the three trees | **exit 1 — `FATAL ERROR: Reached heap limit`, 4078 MB, 2/2 runs** |
| same, `--max-old-space-size=8192` | exit 0 |

The ceiling was **one type program away, for any tree, for anyone** — the next
widening of this block would have produced a V8 stack trace where a lint result
belongs, regardless of who did it or why. That is a property of the gate rather
than of this block, so the flag goes on the spawn in `scripts/lint.ts` via
`NODE_OPTIONS` (an existing value still wins, so the size stays overridable).
8 GB is sized against CI's 16 GB runner — about 3 GB above the observed peak —
rather than against today's number, so the gate does not need re-tuning the next
time a package lands.

### 3. Scope: all three trees, including `scripts/**`

With the headroom applied, the honest cost — same conditions, cold, exit codes
captured, `scripts/**`'s own four hits fixed:

| config | user CPU | peak RSS | exit |
| ------ | -------- | -------- | ---- |
| before | 81.0 s | 3.69 GB | 0 |
| + `apps/cli` + `runtime-pi` | 97.2 s | 5.20 GB | 0 |
| + `scripts/**` | 96.3 s | 4.35 GB | 0 |

About +20 % CPU on a **cold** lint for all three trees, and `scripts/**` — the
one that looked expensive, since its tsconfig deliberately pulls
`../apps/api/src/modules/*/index.ts` into a second program — costs nothing
measurable on top of the other two. It earns its place on what it prevents:
`scripts/migration/**` mutates PRODUCTION data, which is the sharpest form of
this block's own first rationale. Warm lint is unchanged (`--cache
--cache-strategy content` rebuilds a program only for files that moved).

Wall clock is deliberately not quoted anywhere above: the machine was running
many parallel jobs and its wall numbers moved ~40 % between identical runs,
while user CPU did not.

## Tests

- `cd apps/cli && TEST_TIER=0 bun test test/download.test.ts` → **10 pass, 0 fail**.
  Added one case: a payload larger than the sink watermark round-trips
  byte-exact with a matching streamed sha256. That is the discriminating test —
  the watermark is the only thing that makes `FileSink.write` ever return a
  promise, so a download smaller than it never exercises the awaited write.
- `TEST_TIER=0 bun test runtime-pi/sidecar/test/integration-mitm-listener.test.ts
  runtime-pi/sidecar/test/integration-mitm-body-cap.test.ts` → **18 pass, 0 fail**
- `TEST_TIER=0 bun test runtime-pi/test/provision.test.ts` → **22 pass, 0 fail**
- `bun test scripts/test/migration-0010-ee-tables-into-platform-db.test.ts` on
  real Postgres (via the shared lock) → **7 pass, 0 fail**
- Cold `bun run lint`, no external `NODE_OPTIONS` → **exit 0**, zero heap-limit
  lines (was 5 errors + an OOM)
- `bunx tsc --noEmit` on `apps/cli`, `runtime-pi`, `runtime-pi/sidecar`,
  `scripts` → clean
- `bunx prettier --check` on every touched file → clean

No regression test for the sidecar `.catch()`: the only reachable trigger today
is a throwing telemetry sink (`resolveAndCheckHost` swallows resolver throws by
design — `packages/afps-shared/src/ssrf-dns.ts:83-91`), and a test built around
injecting one would assert a contrived seam rather than a real failure. The
issue rated this hit "PLAUSIBLE, low" for the same reason. The guard is one line
and the failure it prevents is process-fatal, so it ships regardless.

## Notes for the orchestrator

- **No migration, no env var, no OpenAPI change.** Nothing to order at deploy.
- **This touches `scripts/lint.ts`, the gate every branch runs through.** The
  change only raises the eslint child's heap ceiling, so it is strictly more
  permissive — but it is a shared file, so it is worth knowing it is in here.
  Without it the config change in this PR cannot pass its own gate.
- **Merge overlap with #1347**: both touch `eslint.config.mjs`. #1347 works in
  the `no-restricted-imports` blocks around lines 530-640; this edit is confined
  to the type-aware block at ~793+ (its comment, and its `files` array which is
  now multi-line). No textual conflict expected.
- **Overlap with the CLI chain (#1320…#1364)**: `apps/cli/src/lib/download.ts`
  and `apps/cli/test/download.test.ts` only, and minimally — one constant, one
  `writer()` call, one `await`, one added test.
- **Watch `scripts/test/lint.test.ts` on the consolidated branch.** Its slow
  case ("fails on a warning that fires, however few") is bounded by a full
  **cold** whole-repo lint against a 300 s ceiling, and its own comment records
  CI already blowing past 120 s once. Cold lint is now ~20 % more CPU, so that
  case moves closer to its ceiling. Its comment says the ceiling "exists to
  catch a HANG, not to bound a known cost", so raising it in step is the
  sanctioned move if CI trips it; I did not pre-emptively raise it because this
  box (many parallel jobs, wall times swinging 40 %) cannot produce a
  representative CI number.

### DIFF
--- a/apps/cli/src/lib/download.ts
+++ b/apps/cli/src/lib/download.ts
@@ -66,6 +66,19 @@ const DEFAULT_TOTAL_MS = 20 * 60_000;
 /** Never fire `onProgress` more than once per this interval (plus a final tick). */
 const PROGRESS_THROTTLE_MS = 250;
 
+/**
+ * Bytes buffered in the destination `FileSink` before it flushes to disk.
+ *
+ * Load-bearing, not a tuning knob: a `FileSink` opened WITHOUT a
+ * `highWaterMark` accumulates every written chunk in memory and flushes only
+ * on `end()` — which for a 113 MB CLI binary is the whole artifact resident,
+ * exactly the buffering this streaming helper exists to avoid (see the module
+ * doc above). With an explicit watermark, peak resident bytes are bounded by
+ * ~one watermark. Same value, and the same reason, as core's streaming upload
+ * (`packages/core/src/storage-fs.ts`).
+ */
+const SINK_FLUSH_BYTES = 1024 * 1024;
+
 /** Reason tags carried on the AbortController so the catch can explain itself. */
 const STALL = Symbol("stall");
 const TOTAL = Symbol("total");
@@ -148,7 +161,7 @@ export async function streamDownload(
       throw new Error(`GET ${url} → empty response body`);
     }
 
-    sink = Bun.file(destPath).writer();
+    sink = Bun.file(destPath).writer({ highWaterMark: SINK_FLUSH_BYTES });
     const reader = res.body.getReader();
     armStall();
     try {
@@ -158,7 +171,13 @@ export async function streamDownload(
         armStall();
         const bytes = chunk.value;
         hasher.update(bytes);
-        sink.write(bytes);
+        // AWAITED: `FileSink.write` returns `number | Promise<number>`, and
+        // returns the promise exactly when the sink has to drain — which is
+        // what `SINK_FLUSH_BYTES` above makes happen. Un-awaited it dropped
+        // both the backpressure and the rollback: a mid-write failure left the
+        // `catch` below untriggered and escaped as an unhandled rejection
+        // (fatal in Bun), so the partial file at `destPath` was never unlinked.
+        await sink.write(bytes);
         received += bytes.byteLength;
         emit(false, total);
       }
--- a/apps/cli/test/download.test.ts
+++ b/apps/cli/test/download.test.ts
@@ -54,6 +54,29 @@ describe("streamDownload", () => {
     expect(new Uint8Array(await readFile(dest))).toEqual(payload);
   });
 
+  it("round-trips a payload larger than the sink watermark", async () => {
+    // The destination `FileSink` is opened with an explicit `highWaterMark`
+    // (SINK_FLUSH_BYTES, 1 MiB) so it flushes mid-stream instead of holding
+    // the whole artifact in memory until `end()`. That watermark is the only
+    // thing that makes `FileSink.write` ever return a promise — a download
+    // smaller than it never drains, so it never exercises the awaited write.
+    // Cross it: the bytes on disk and the streamed sha256 must still match.
+    const dest = await tmpDest();
+    const payload = new Uint8Array(2 * 1024 * 1024 + 7);
+    for (let i = 0; i < payload.length; i++) payload[i] = i % 251;
+    const chunkSize = 64 * 1024;
+    const chunks: Uint8Array[] = [];
+    for (let o = 0; o < payload.length; o += chunkSize) {
+      chunks.push(payload.subarray(o, Math.min(o + chunkSize, payload.length)));
+    }
+    const res = await streamDownload("https://example/big", dest, {
+      fetchImpl: async () => streamingResponse(chunks),
+    });
+    expect(res.bytesWritten).toBe(payload.length);
+    expect(res.sha256).toBe(sha(payload));
+    expect(new Uint8Array(await readFile(dest))).toEqual(payload);
+  });
+
   it("reports progress with total from Content-Length", async () => {
     const dest = await tmpDest();
     const chunks = [new Uint8Array(10), new Uint8Array(10)];
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -790,7 +790,84 @@ export default tseslint.config(
     //
     // If this ever needs to get cheaper, narrow `files` before dropping a rule:
     // the cost is the type program, not the rule count.
-    files: ["apps/api/src/**/*.ts", "packages/*/src/**/*.ts"],
+    //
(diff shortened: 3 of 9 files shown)


pierrecabriere:

PR #1436, opened 2026-09-10
fix(licence): close the three leaks past the commercial boundary

### LINKED ISSUE #1373: licence boundary: the generated SPA schema, the removal recipe and the isolation gate each leak past it

pierrecabriere:
Closes #1373

Three gaps let commercial-module content or dependencies cross the Apache-2.0 boundary with every gate green. None was a licence violation; all three made the boundary weaker than the gates claimed. The common cause is that `packages/module-ee` is opt-in at **runtime** (`MODULES`), so a build-time coupling to it costs nothing until someone actually deletes the directory.

## Root cause

1. **The generated SPA schema** (`apps/web/src/api/schema.d.ts`) is the one file `verify:license-boundary` exempts, and it byte-commits the module's `Ee*` schemas and `/api/billing/*` paths. Its banner said "built-in module contributions" — neither where they come from nor a hint that a licence question exists. `GENERATED_ALLOWLIST` additionally took the list's word for it: an entry was exempt because it was listed, not because it was generated.
2. **The removal recipe** (`packages/module-ee/README.md:36-43`) named four files. Executed verbatim in a throwaway worktree, the real removal touches ~40.
3. **`verify-module-isolation`** (`scripts/verify-module-isolation.ts:350-356`) scanned `src` trees plus `scripts/`, so a static `@appstrate/module-ee` import from `apps/api/test/**` — or from `apps/web/**`, waived from the core→module rule on purpose — passed every gate.

## Fix

**1 — the generated schema states what it carries.** Filesystem discovery is deliberate (a byte-compared artifact must not vary by `MODULES`) and is itself the removal mechanism, so the banner now says so: what it merges, that `packages/module-ee/` is source-available, and that deleting the directory and regenerating drops its contract. A `GENERATED_ALLOWLIST` entry now has to *prove* it is generated — an `AUTO-GENERATED by <script>` marker naming a script that exists — so the one list that waives the licence gate cannot waive it for a hand-written file. Only the banner changed in `schema.d.ts`.

**2 — the recipe is complete, and gated.** Rewritten as eight ordered steps covering the module and the `0010` one-off, the OpenAPI registries, 14 SPA deletions, 6 SPA edits, the 112 i18n keys (a reverse guard fails on an orphan), `verify-no-migration-dml` (**required** — it exits `ENOENT` on a directory it scans), `knip.config.ts`, the dead `@appstrate/emails` re-exports, the six gate self-tests that assert against the real repo, and the regeneration step. Purely cosmetic residue (`docs/ENV.md`, `.env.example`, `stripe-live.yml`, comment-only mentions) is one closing line, explicitly "no gate behind them".

`scripts/test/module-ee-removal-recipe.test.ts` keeps it from rotting again. It derives two sets mechanically from the tracked files — SPA sources naming an `/api/billing` path or an `Ee*` schema identifier, and any file outside the module carrying a quoted `packages/module-ee` path — and fails unless the recipe names each one. It deliberately does not check what the section *says* about a file: naming it is what forces an author to look. The i18n keys and the SPA files that reach the module through a helper stay prose, because no cheap grep finds them; the test's header says so rather than implying coverage it lacks.

**3 — isolation covers every tracked source file.** A third direction over the git index (the same population `verify-license-boundary` reads), narrowed to the one source-available tree so it needs no acceptance list. The lexer had to stop reading an import *quoted inside a string literal*: test fixtures contain those, and the gate's own test reported itself five times.

## Tests

- `scripts/test/module-ee-removal-recipe.test.ts` (new, 26 cases) + `verify-module-isolation.test.ts` + `verify-license-boundary.test.ts` + `tracked-files.test.ts` → **62 pass, 0 fail** (`TEST_TIER=0 bun test`).
- Negative control: added a throwaway `apps/web/src/lib/zz-probe.ts` containing `/api/billing/plans` → the gate fails with `packages/module-ee/README.md § "### Removing the module from a redistribution" does not name apps/web/src/lib/zz-probe.ts`. Removed after.
- `bun run verify:license-boundary` → 2518 files, 2428 Apache-2.0, 90 commercial, 1 exempt.
- `bun run verify:module-isolation` → 206 files across 10 modules, 2429 tracked files with no static import of `packages/module-ee/`.
- `bunx tsc --noEmit` on `apps/api`, `apps/web` and `scripts/tsconfig.json` → clean.
- `bunx eslint` + `bunx prettier` on touched files → clean.

## Notes for the orchestrator

- No migration, no env var, no OpenAPI contract change — `schema.d.ts` differs by its banner comment only, so `verify:api-types` is unaffected.
- The recipe is **documentation plus a gate**; `packages/module-ee` is not removed on this branch.
- The new gate derives from `git ls-files`, so a sibling branch that adds an SPA file touching `/api/billing` will fail it until the recipe names that file. That is the intended behaviour, not a conflict.
- The isolation gate's new direction is repo-wide: any branch introducing a static `@appstrate/module-ee` import outside `packages/module-ee/` now fails `bun run check`. Nothing in the tree does today.


pierrecabriere:

Issue #1450, opened 2026-09-17
0068_packages_org_home_validate.sql: VALIDATE the packages home CHECK, in the release AFTER beta.58


pierrecabriere:
`0067_packages_org_home_required.sql` adds the CHECK `packages_org_package_has_home` as **`NOT VALID`**, deliberately. This issue tracks the other half, which cannot ship in the same release.

## Why it is not in beta.58

The rollout (`scripts/migration/README.md` → "Release beta.58") applies `0056`→`0067` at **step 3**, through a one-shot migrator — *not* by starting the application — and `scripts/migration/0014-packages-home-space-backfill.sql` runs at **step 4**, with the platform still stopped.

A `VALIDATE` inside that batch would therefore scan `packages` while every `home_space_id` is still `NULL`, raise `23514`, and roll the whole deploy back on the very statement meant to confirm it.

`NOT VALID` already governs every INSERT and UPDATE from the moment it applies, which is the half that matters during the window. `0014` ends with the `VALIDATE` for the rows that predate it.

## What the follow-up release must carry

`packages/db/drizzle/0068_packages_org_home_validate.sql`, whose only statement is:

```sql
ALTER TABLE "packages" VALIDATE CONSTRAINT "packages_org_package_has_home";
```

On a fresh install and on a replay it has nothing to walk. On a deployment that skipped `0014` it fails loudly with `23514` — the intended outcome.

## Precondition, and how to read it

`scripts/migration/0014` must have **committed everywhere the release will land**.

Confirm with the runbook's own count:

```sql
select count(*) from packages
where org_id is not null and not ephemeral and home_space_id is null;
```

It must be zero.

**Do not read `pg_constraint.convalidated` as the rollout signal.** It stays `false` for ever on an installation that never had rows to fix, and `true` on one that ran `0014` — same schema, same enforcement on every write, two different values. The count above is the signal.

## References

- `packages/db/drizzle/0067_packages_org_home_required.sql` — full argument in the file header
- `scripts/migration/README.md` → "Next release — \`0068 … VALIDATE CONSTRAINT\`, and why it is not in this one"
- `docs/NO_TRANSITIONAL_CODE.md` §2 — the `ADD … NOT VALID` / `VALIDATE` split this follows
- Comparable tracked debt: #1275 (`0056`'s window)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1468, opened 2026-09-18
chore(deploy): delete the prose describing how the deployment used to work

pierrecabriere:
`docs/NO_TRANSITIONAL_CODE.md`, first line: *the codebase describes how the system works now. It never describes how it used to work.* Four files were still narrating the cutover of the last two weeks.

## `deploy/` — five blocks of history

The billing module's former repository, the project name the compose read before the resource was rebuilt, and three separate retellings of the 2026-09-18 outage with dates and per-variable measurements. Every rule those paragraphs argued for is now enforced by a gate and a test, so the prose was arguing a case nobody has to be persuaded of.

What stays is current-state and nothing else: why this file is not the self-hosting example, why `name:` is not `appstrate`, why a bare name in `environment:` is refused. Measured: **zero lines** describing how it used to work, down from five blocks; 293 → 252 lines, 119 → 78 comment lines.

The header's hand-kept list of required variables goes too. `.env.example` is the operator contract; a second copy is a second thing to drift, and nothing compared them.

**`deploy/.env.example`** also carried the one-shot procedure for moving the billing tables out of the database the out-of-tree module used to run against. That ran on 2026-09-17. An operator reading this file today is not migrating anything, and the instruction would send them to delete a variable that no longer exists.

## `scripts/migration/README.md` — a runbook two weeks past its window

344 lines still in the present imperative — *"do NOT touch it here, it is step 5"*, *"run it FIRST"*, *"Production runs `MODULES=…,@appstrate/cloud`"* — for a release that shipped. Whoever opens this file for the next release reads the previous one's instructions as live.

Its own § *Next release* already says how this ends: **"a runbook gets filed as done."** It is filed. What replaces it is the state that release left, read back off production on 2026-09-18, and what it still owes:

| Read back | Value |
| --- | --- |
| `drizzle.__drizzle_migrations` | 68 rows — `0000` … `0067`, the whole batch |
| `drizzle.ee_migrations` | 8, with 31 billing accounts and 717 claimed ledger rows |
| `org_members` by role | 31 owner / 16 admin / 1 member / 2 guest / **0 viewer** |
| `oauth_clients.self_service IS NULL` | 0 |
| `packages_org_package_has_home` | `convalidated = true` |
| personal spaces | 6, against 50 memberships |

Per script, from witnesses rather than from the plan: `0010`, `0008`, `0011`, `0012`, `0014`, `0016`, `0017` ran. `0009` and `0013` were **not needed** — their preconditions are the unique indexes `uq_org_invitations_pending` and `uq_org_models_unaliased_binding`, both of which exist, and the batch that creates them applies in one transaction, so it could not have committed had either pre-flight counted a duplicate. `0015` is deliberately outstanding, and 6 of 50 personal spaces exist because `provisionMember` makes them at a membership door.

The Log's status column read `not applied` for every one of those.

**Still owed, and unchanged: `0068`.** Production is not waiting on it — `0014` validated the constraint there. A fresh install is, which is [#1450](https://github.com/appstrate/appstrate/issues/1450).

The sixteen cross-references into the deleted step numbering are rewritten to state their own ordering constraint (*"with the platform stopped, before the drizzle batch"*), which is what they meant and does not expire with a release. 1015 → 728 lines; every per-subject *Detail* section is kept.

## Verification

`docker compose config` over the old and the new `deploy/` files resolves to **byte-identical** output — 248 lines each, empty diff. This change is comments only.

That is the right question here and it was the wrong one during the outage, where the claim under test was about Coolify and the evidence came from Docker Compose. The Coolify-specific hazard is covered separately, by the bare-name test and by class 6 in #1466.

`verify:compose-defaults`, `verify:env-docs`, `verify:release-version`, `verify:no-migration-dml` green; 52 tests across the two affected suites pass.


pierrecabriere:
Rebasé sur `main` après #1467 (1.0.0-beta.59, et `0068` écrit).

Deux résolutions, plus une correction de fond :

- **`deploy/docker-compose.yml`** — pin `beta.59` de `main`, commentaire raccourci de cette PR. `docker compose config` sur les deux côtés redonne une sortie **identique**, comme avant le rebase.
- **`scripts/migration/README.md`** — la version de `main` l'emporte sur le §`0068`, qui est la plus récente. Ses deux renvois vers la numérotation supprimée (« runbook step 3 », « step 4 ») sont réécrits en contrainte d'ordre.
- **Le passage « ce que la release doit encore »** disait `0068`. #1467 l'a écrit ; il dit maintenant qu'il a shippé en beta.59 et pourquoi il ne pouvait pas rouler dans le lot de beta.58.

Les trois portes concernées repassent au vert après rebase.

PR #1475, opened 2026-09-21
feat(rbac): separate `agents:run-inline` from `agents:run`

pierrecabriere:
>
> 👉 1. #1475 — `agents:run-inline` scope
>

## The hole

`agents:run` gated two acts of different nature:

- **published agent** — a manifest someone in the space composed, reviewed and pinned to a version
- **inline** — a manifest the *request body* carries, declaring its own `dependencies.integrations`: arbitrary code in a container holding the space's credentials

The `runner` preset — whose entire point is *launch what you may not read* — inherited the second along with the first. Nobody decided that.

## What changes

`agents:run-inline` gates the three surfaces that take a manifest in the body:

| Surface | Before | After |
|---|---|---|
| `POST /api/runs/inline` | `agents:run` | `agents:run-inline` |
| `POST /api/runs/inline/validate` | `agents:run` | `agents:run-inline` |
| `POST /api/runs/remote` `source.kind: "inline"` | `agents:run` | `agents:run-inline` |
| `POST /api/runs/remote` `source.kind: "registry"` | `agents:run` | unchanged |

`/runs/remote` carries both kinds behind one route, so its route-level guard asserts the **disjunction** (`requireAnyPermission`) and each branch asserts the exact grant it needs before doing any work. `PATCH /runs/:runId/sink/extend` takes the same disjunction — gating it on `agents:run` alone would let a caller start a remote inline run and then be refused the sink extension its own runner needs.

## Preset matrix — no edit required

`admin` derives every space-level permission from the catalog, `builder` derives admin minus the governance prefixes. Both pick the new action up automatically. `operator`, `runner` and `viewer` enumerate their grants by hand and stay out.

| preset | `agents:run` | `agents:run-inline` |
|---|---|---|
| `admin` | ✅ | ✅ |
| `builder` | ✅ | ✅ |
| `operator` | ✅ | ❌ |
| `runner` | ✅ | ❌ |
| `viewer` | ❌ | ❌ |

## Allowlists

**API key: grantable.** `appstrate run ./agent.afps` posts an inline source with a key (`apps/cli/src/commands/run.ts` → `kind: "inline"`). Excluding it would break the shipped headless flow. The creator ceiling is what bounds it: only a creator who holds the grant (`admin`, `builder`, `operator`, or a custom role naming it — never a `runner`) can mint a key that carries it. What the split buys is that a key scoped for launching published agents no longer executes whatever manifest its holder composes.

**OIDC end-user: excluded.** An end-user is an external identity impersonating through a space. A client that could acquire this scope by consent would have the platform execute a manifest of the *embedding app's* own composition — the silent escalation that allowlist exists to prevent. Launching an agent someone in the org published stays in the vocabulary.

**OIDC dashboard user: requestable.** `agents:run-inline` joins `runs:read-all` in `OIDC_DASHBOARD_ONLY_SCOPES`. A dashboard-user token is capped by the user's live org and space roles, re-read per request, so an org-level client acting for an admin keeps the inline surfaces and a `runner` never gains them. Space-level clients and end-user tokens still cannot carry it.

## No fallback

Per `docs/NO_TRANSITIONAL_CODE.md`, there is no `agents:run` fallback — that fallback *is* the hole being closed.

⚠️ **Frozen permission arrays do not follow the derivation.** A custom space role (`space_roles.permissions text[]`) or an API key (`api_keys.scopes`) holding `agents:run` loses the inline surfaces until an admin adds the new string. Deliberate: nobody explicitly asked for arbitrary-manifest execution. Pre-deploy audit:

```sql
SELECT id, org_id, key, name FROM space_roles WHERE 'agents:run' = ANY(permissions);
SELECT id, org_id, name FROM api_keys WHERE 'agents:run' = ANY(scopes);
```

## Tests

- `apps/api/test/unit/permissions.test.ts` — new `agents:run-inline` block: preset matrix, non-implication in **both** directions, API-key grantability under the creator ceiling, space-level-ness
- `apps/api/src/modules/oidc/test/unit/client-scope-vocabulary.test.ts` — admitted at `org` and `instance` levels (dashboard users), refused at `space` level, with `agents:run` admitted everywhere as the control
- `runner-preset.test.ts` — a `runner` is refused at the guard with an answer naming `agents:run-inline`; the adjacent preset, `operator`, gets past it (the stronger discriminator). `POST /api/runs/remote` is pinned per branch by each guard's exact message: runner → inline source (403, `agents:run-inline`), a `["agents:run-inline"]` key → registry source (403, `agents:run`), an operator witness past the inline guard, and an `["agents:run"]` key → `POST /runs/inline` (403). Removing the inline branch's `assertPermission` turns the suite red
- `inline-rbac-boundaries.test.ts` — actor promoted `operator` → `builder` and key scopes retargeted, so each case still tests *dependency* authorization rather than the new route gate

## Gates

- `bun run check` — 42/42 ✅
- `bun test apps/api/test/unit packages/core/test` — 3167 pass, 0 fail ✅
- `verify:openapi` ✅ · `detect:breaking` — 0 breaking, 0 non-breaking (descriptions only; the behavioural change is a required-permission change, not a wire-shape one) · `schema.d.ts` regenerated
- Integration suite needs the `integration` label on this PR.

## Docs

`docs/architecture/RBAC_PERMISSIONS_SPEC.md` §3.3 + §3.4. The `agents` actions cell also regains `share`, which it had been missing.


pierrecabriere:
## Production evidence — and one correction to the preset matrix

Read against the live database before deciding what to ship.

### `space_roles` is empty

**0 rows.** No custom space role exists, so the "a frozen `permissions` array does not follow the derivation" risk has no instance today. The pre-deploy query in the description stays useful for later; right now it returns nothing.

### The inline surfaces are live

**239 inline runs**, of which **85 from the chat**, the most recent **today**.

| org role | effective space role | users | inline runs | holds `agents:run-inline` |
|---|---|---|---|---|
| `owner` | `admin` (`resolveSpaceRole:74`) | 6 | 100 | ✅ |
| `admin` | `admin` | 7 | 138 | ✅ |
| `member` | `operator` — `default_role` fallback (`resolveSpaceRole:78`) | 1 | 1 | ❌ before this fix |

### The correction

Excluding `operator` was **not a decision** — it fell out of the derivation (`admin`/`builder` derive from the catalog; the other three enumerate by hand). An org member with no `space_members` row in an `open` space resolves to that space's `default_role`, which is `operator`. Withholding the grant there puts the chat's inline composition behind an admin role for **every ordinary member**.

`operator` now names `agents:run-inline` explicitly. The preset that must not hold it is `runner`, and only `runner`: it launches what it may not READ, and a body-supplied manifest naming its own dependencies is exactly how it would reach around that. That was the hole this PR set out to close, and it stays closed.

The integration test's discriminating control moved from `owner` to `operator` — the adjacent preset clearing the guard is the stronger statement.

### The API-key half is confirmed too

One key has launched inline runs: **`Claude Code`**, org **TANIÈRE**, **8 runs**, still active. Had `agents:run-inline` been excluded from `API_KEY_ALLOWED_SCOPES` as first drafted, that flow would have become impossible to restore at all — no key could ever carry the grant.

### Remaining blast radius at deploy: one API key

`Claude Code` (org TANIÈRE) carries an **explicit** scope list containing `agents:run`. Frozen arrays do not follow the derivation, so it loses the inline surfaces.

There is no `PATCH /api/api-keys` — create and revoke only — so the remediation is to **re-mint it** with `agents:run-inline` included. Worth a heads-up to that org before the release.

Every other live key either carries no explicit scopes (full role access, which resolves against the live preset) or has never touched an inline surface.

pierrecabriere:

PR #1476, opened 2026-09-21
feat(chat): bind the model to the conversation, and show which one answered

pierrecabriere:
>
> 👉 2. #1476 — modèle lié à la conversation
>

## The question behind this

*« Bloquer la sélection du modèle une fois le chat démarré ? Que fait le SOTA ? »*

**The SOTA does not lock.** ChatGPT, Claude.ai, Cursor, Copilot, Perplexity, LibreChat, OpenWebUI — all allow switching mid-conversation, most per message. The only real locks anywhere are *technical*, and this codebase already dismantled every one of them **on purpose**:

| Technical lock | Status here |
|---|---|
| Signed Anthropic `thinking` blocks | `structured-session.ts:200` — `reasoning` parts deliberately not projected |
| `previous_response_id` / `reasoning.encrypted_content` | `HISTORY_MODEL_SENTINEL` forces Pi's `isDifferentModel` branch, which scrubs unpaired items |
| Incompatible tool-call id formats (`fc_…` 450 chars vs 64-char Anthropic) | same sentinel — cross-model id normalization runs |

Locking would throw that work away. **The actual defect is elsewhere.**

## The defect

`model-store.ts` held **one global value** in localStorage. Open a conversation answered by model A while the store holds B → it continues on B. Nothing pinned it. Nothing showed it. That is the silent part, and it is what people reach for a lock to fix.

## What changes

**Two scopes.** localStorage is the *default* — what a new conversation starts on, what an explicit pick updates. The open conversation's model is seeded from its **own transcript** and overridden by a pick.

One slot, not a map: the conversation component is keyed by id with `gcTime: 0`, so reopening refetches and re-seeds. Caching older conversations here would cache something already authoritative elsewhere — and unboundedly.

**No migration.** `AppstrateTurnMetadata` gains `modelId` + `modelLabel`, stamped by `closePiTurn`. That metadata already rides the live `finish` chunk *and* the stored `content` (`toContent` keeps everything but the id), so one write reaches the stream, the reload and the resume alike.

Stamped on **every** exit, failures included — *which model failed* is the question asked of a turn that errored.

**Two fields, two lifetimes** — the split `runs` already makes between `llm_usage.model` and `runs.model_label`:

- `modelId` re-seeds the picker; it **may dangle** once an org model is deleted, and a miss just falls back to the default
- `modelLabel` is frozen at write time, so a transcript still names the model after that row is gone. For an aliased model it is the **alias** label, never the hidden backing id

**A badge** on each assistant message, low emphasis, in the row the action bar already reserves (`h-7`, so zero layout change). Always rendered rather than hover-gated: the point is that someone scrolling notices the switch without looking for it.

## ⚠️ The trap this must not fall into

Neither field feeds the history projection. `buildStructuredPiTurn` stamps historical assistant messages with `HISTORY_MODEL_SENTINEL` **on purpose** so `isSameModel` fails and Pi's cross-model normalization runs. Feeding a real model id back there would re-enable `isSameModel` and break replay for every conversation that ever switched models.

Both `chat-turn-metadata.ts` and `turn-model.ts` carry that warning where a future edit would look.

## Tests

- `model-store-conversation.test.ts` — the regression itself (A's model must not leak into B), fallback to the default for a transcript-less conversation, a pick surviving a late history load, a seed for a conversation the user already left being ignored
- `turn-model.test.ts` — the three **absent** cases (user message, pre-feature row, turn that died before binding) read as no model, never a default; newest turn wins; a server-authored notice between turns is skipped rather than treated as a reset
- `pi-chat-turn-closure.test.ts` — the model survives the real `readUIMessageStream` assembly on the happy exit **and** the error exit, and is absent rather than guessed when the turn bound none

## Gates

- `bun run check` — 42/42 ✅
- `bun test packages/module-chat/test packages/core/test` — 1693 pass, 0 fail ✅

**Not verified live in a browser.** Producing a real transcript needs a seeded instance, a configured model credential and a billed LLM call; the metadata round-trip is covered through the real stream-assembly path instead. The one thing that leaves unchecked is the badge's visual placement — worth a glance in a dev instance that has conversations.

## Deliberately out of scope

Greying out a model that cannot read the open thread (a text-only model on a conversation containing images — `OrgModel.input` carries the modalities). `composerSlot` is built in `ChatPage`, which has no access to the conversation's messages, so it needs either moving the slot or a second store. That is a distinct concern and deserves its own review.


pierrecabriere:

PR #1477, opened 2026-09-21
feat(chat): show the caller's role and what the assistant may do with it

pierrecabriere:
>
> 👉 3. #1477 — chip d'accès (dans le composer)
>

## Why

The assistant runs on the caller's **own** permissions — the platform-MCP bearer it authorizes with carries exactly the caller's effective set, no amplification (`packages/module-chat/src/prompt.ts`). So *"what may the assistant do"* and *"what may I do"* are the same question.

Until now the only way to learn the answer was to ask for something and read the refusal.

## What

A chip in the composer, beside the model picker, naming the caller's role **in this space**, opening a short list of the acts the chat actually performs:

| Row | Reads |
|---|---|
Every row first requires `chat:write` (`POST /api/chat`) and `mcp:read` (the MCP entry); the rows the assistant performs through `invoke_operation` / `run_and_wait` also require `mcp:invoke`. Each row then reads the grant its server path checks:

| Row | Reads (on top of the above) |
|---|---|
| Use the Appstrate API on your behalf | `mcp:invoke` |
| Run your agents | `agents:run` ∧ (`runs:read` ∨ `runs:read-all`) — `run_and_wait` must read the run back |
| Create agents | `agents:write` — editing an existing agent is authorised by its home space, so the row does not claim it |
| See your runs | `runs:read` ∨ `runs:read-all` |
| Browse your files | `files:read` (listing/detail; `list_files` needs no `mcp:invoke`) |
| Connect your accounts | `integrations:connect` |
| Activate an integration for the space | `maySetPackageActive(…)` |
| Schedule runs | `schedules:write` |

A row never shows ✓ for an act the server would refuse; each has a test that removes only its own permission.

## Three deliberate choices

**Not a permission dump.** ~40 `resource:action` strings in English, most of which the chat never touches. Printing them answers a question nobody asked while burying the one they did.

**Not in the model popover.** Picking a model is an **act**; a role is a **state** you are subject to. One control for both suggests the second is changeable there like the first.

**In the shell, not in `module-chat`.** It reaches the composer through a new `composerActions` prop — a host-owned slot beside the model picker, the same kind of seam as `headerActions`. Everything it reads is host state, so the `ChatHost` bag gains nothing; `module-chat` only gains the slot. The conversation header keeps to the conversation's own actions.

## The role preview handles itself

`/api/orgs` and `/api/spaces` both carry `X-View-As` through the typed client's middleware (`api/client.ts:79`), so `usePermissions()` and the space row already answer **as the persona**. An owner previewing `runner` sees the runner's verdicts — the point of a preview.

The chip does not START a preview — that belongs to role administration (settings → roles, space members). While one is active it shows an eye and says so, since its verdicts are then the persona's.

## The row that is not a permission check

`activateIntegrations` asks `maySetPackageActive`, not `can("integrations:install")`. In a personal space the owner holds `operator`, which carries no activation grant — and `gateSpacePackageWrite` waives it because owning the space **is** the authorization (RBAC spec §3.6).

A row keyed on the bare permission would claim a refusal that never happens. **A chip that over-claims is worse than no chip**: the user is told the assistant can do something, asks for it, and gets a refusal they now have no way to explain. Every case in `chat-access.test.ts` is a way of over-claiming.

## Tests

- `chat-access.test.ts` (10) — nothing granted to a caller holding nothing (including `spaceGrant: undefined`, i.e. still loading, which must read as *no*); the run-read disjunction; the personal-space exemption **and** its `access === "none"` guard; connecting ≠ activating; `mcp:invoke` not implied by holding agents; unique ids and label keys (the copy-paste bug this table invites)
- `chat-access-chip.test.tsx` (4) — renders **nothing** while the permission sets are in flight (`can()` answers `false` for a set still loading, which is indistinguishable from a denial — without the guard every hard reload paints "you may do nothing"); space role wins; a custom role renders its author's name, not a slug; org role as fallback

Both label keys are spelled in full rather than built from `id`, so the locale guard's reverse check sees them and no `DYNAMIC_KEY_PREFIXES` entry is needed.

## Gates

- `bun run check` — 42/42 ✅
- `bun test` in `apps/web` — 981 pass, 0 fail ✅ (includes the fr/en parity + dead-key locale guards)

**Verified in a browser** on the assembled stack: chip beside the model picker, popover opening upward, verdicts for the owner and under a `runner` preview.

## Follow-up

The "compose agents on the fly" row (`agents:run-inline`) is added by #1478, on top of this PR.


pierrecabriere:

PR #1478, opened 2026-09-21
feat(chat): inline-agents toggle in the composer, and the grant in the access chip

pierrecabriere:
>
> 👉 4. #1478 — bouton agents à la volée + ligne du chip
>

## The two layers

[#1475](https://github.com/appstrate/appstrate/pull/1475) decides whether a caller **may** compose an inline agent. This is whether they **want to, right now**.

| | |
|---|---|
| **`agents:run-inline`** | the ceiling — RBAC, server-side, someone else's decision |
| **the composer's switch** | the caller lowering it on themselves — a persisted preference, applied to every turn until changed |

The turn's authority is the **conjunction**. `inline_agents: true` from a caller who does not hold the grant changes nothing. An absent flag means **on**, so every client that predates the switch — and the CLI — behaves exactly as before.

Composing an inline agent costs a container plus its own LLM budget, and can send the assistant off *building* when it was asked a question. That is a reason to switch it off that has nothing to do with anyone's role.

## Why the switch is real and not cosmetic

It narrows the turn's own **platform-MCP bearer**.

That token is minted from the caller's resolved permission set and is the whole authority the engine has for the turn. Dropping `agents:run-inline` from it means `run_and_wait` with `kind:"inline"` — and `POST /api/runs/inline` by any other path the engine might reach — is refused by **the platform guard #1475 added**, not by a second policy written into the chat.

**One gate, narrowed. Never two gates to keep in step.**

A client-side-only switch would have been decoration: the model would still call the tool.

## The MCP surface follows the same permissions

The narrowed token is also what the platform's MCP server reads to build `run_and_wait`'s descriptor and its own instructions, per request. Without `agents:run-inline`, `kind` offers only `"agent"`, the `manifest` / `prompt` / `context_files` arguments and the inline notes are gone, and the server instructions stop naming `runInline` / `validateInlineRun`. That holds for **every** MCP client lacking the grant (a `runner`, a scoped key), not only the chat. Flipping the switch changes prompt and tool descriptor on the same turn: one prompt-cache miss, not two.

## The prompt

`SYSTEM_PROMPT` becomes `buildSystemPrompt({ canRunInline })`.

When the capability is absent, the inline-composition instructions are **absent** — not contradicted by a late override. That includes every inline-only argument (`context_files`, a top-level `prompt`): an existing agent takes a file through a declared file field of its `input`. When no existing agent matches, the trimmed persona says so and stops. A prompt that taught the capability and then revoked it would spend the tokens teaching something the platform refuses, and leave the model to pick which half of a self-contradicting brief to believe.

Every inline-specific span is wrapped in `ifInline` / `unlessInline`, so it stays visible as such to whoever edits the prose next. Six sentence-level spans plus one block; the reduced persona is ~4k characters shorter, and a test pins that.

## The toggle, and the row

An **icon button right of the attachment button** (bot / bot-off, `aria-pressed`), with a tooltip that states the current mode and what a click does. Nothing about inline agents is left in the model popover.

The access chip from #1477 gains a plain ✓/✗ row, "compose agents on the fly", from the guards the server checks (`mcp:invoke`, run-read, `agents:run-inline`): the chip states the grant, the toggle is what the caller keeps of it.

Hidden outright for a caller without the grant — a switch for a capability the platform would refuse is a lie discoverable only by trying. `ChatHost` gains one boolean for that; the module resolves no RBAC of its own and must not start.

Global and persisted (per browser), default on — the tooltip says so, and the icon itself shows the off state. ON is the **absence** of a stored value, so a user who never touched it and one who turned it back on are the same state.

## A fixture that was wrong

`chat-admission-gate.test.ts` set `permissions: []` where `enterSpaceContext` writes a `Set` and `ChatEnv` declares one. It survived because no reader ran before the gate; the first one that did found `.has` missing. Fixed in the fixture, not worked around in the handler — the two existing readers are unguarded and a guard here would imply a case that cannot happen.

## Tests

- `apps/api/.../chat-inline-agents-ceiling.test.ts` — a real operator holds `agents:run-inline` through its space role; the token minted with the switch off gets **403** on `POST /api/runs/inline` (control: 400 with the grant) and a `run_and_wait` descriptor limited to `"agent"`
- `mcp/test/unit/run-and-wait.test.ts` — descriptor and instructions with and without the grant (neutralising `offersInlineRuns` turns 4 tests red)
- `chat-stream-handler.test.ts` — end to end through the real handler, decoding the minted bearer back through `chatLoopbackStrategy`: switch on → grant kept **and** instructions present; switch off → grant **dropped from the token**, `agents:run` untouched, instructions gone; **absent flag → on**; `inline_agents: true` without the grant → nothing gained, nothing taught
- `prompt-invariants.test.ts` — the existing invariants now bind to the inline-enabled persona; a new block pins the reduced one: no `kind:"inline"` anywhere, **and** every unrelated rule still present (the discriminating half — a builder returning a stub would pass the first check and fail this one), the draft-fallback sentence swapped rather than left dangling, and a size floor so the branch cannot be emptied of anything worth omitting
- `inline-agents-store.test.ts` — default on, round-trip, notification only on a real change, and survival on a runner with no `localStorage` at all

## Gates

On the top of the stack:

- `bun run check` — 42/42 ✅
- api unit (+ mcp, oidc) — 2178 pass · inline/RBAC integration — 27 pass · module-chat — 500 pass · web — 998 pass · core — 1218 pass — 0 fail ✅
- Verified in a browser: toggle on/off, tooltip for both states, `aria-pressed`, persisted value, chip row.


pierrecabriere:

